PR #157, opened 2016-07-21
Reset-GcSqlSslConfig

marceloyuela:
The cmdlet and tests for Reset-GcSqlSslConfig


marceloyuela:
I work for google

Google.PowerShell/Sql/GcSqlSslCmdlets.cs
```diff
@@ -313,4 +313,94 @@ protected override void ProcessRecord()
             WriteObject(result);
         }
     }
+
+    /// <summary>
+    /// <para type="synopsis">
+    /// Deletes all client certificates and generates a new server SSL certificate for the instance.
```
ILMTitan: Stop the synopsis here. Synopsis should generally be shorter than description.

ILMTitan:
Your synopsis is a little long, but otherwise LGTM.

PR #163, opened 2016-07-25
Add Set-GcsObject cmdlet

chrsmith:
This PR adds a new `Set-GcsObject` cmdlet that allows you to update the metadata of an object. Currently the only operation is that it will allow you to specify a default ACL (e.g. "publicRead".)

In addition the PR changes the project for a few API calls, previously we had the default behavior where the API would return the object without the ACL information. This leads to confusion since it looks like objects have no ACLs at all, when in reality they do.

I'll log some issues (assigned to me) to continue fleshing out ACL-support in GCS. Specifically some integration tests for default object ACLs on buckets, and that they work correctly. Also, the ability to read/write object metadata as key/value pairs.


chrsmith:
PTAL @ILMTitan

ILMTitan:
LGTM

PR #169, opened 2016-07-26
Remove Copy-GcSqlInstance

marceloyuela:
Since Copying an instance is explicitly declared as not supporting second-generation instances, it is better to remove it (like we did with the database cmdlet).


ILMTitan:
LGTM

PR #170, opened 2016-07-26
Updating documentation 2

marceloyuela:
The second big update to XML documentation. With this, most sql cmdlets should have correct and updated XML documentation.


ILMTitan:
LGTM

PR #172, opened 2016-07-26
Remove Test-IsElevated

ILMTitan:
Test-IsElevated is part of a gwindow module, not standard powershell. This replaces it with its definition.


chrsmith:
LGTM

Issue #127, opened 2016-07-14
Read-GCSObject Remove-GCSObject should get bucket from the piped GCSObject


YnR:
Currently: `Find-GcsObject -Bucket ynrpizza | Read-GcsObject` asks for a bucket, where I assumed it was passed by the GcsObject.

Same behavior with Remove-GCSObject

ILMTitan:
Remove-GcsObject was fixed in PR #77.

ILMTitan:
Read-GcsObject fixed in PR #178.

PR #178, opened 2016-07-28
Make Read-GcsObject take an input object from the pipeline.

ILMTitan:
Folded into this is a fix for making Write-GcsObject write zero bytes correctly. This fixes issue #127.

### DIFF
--- a/Google.PowerShell/Storage/GcsObject.cs
+++ b/Google.PowerShell/Storage/GcsObject.cs
@@ -411,7 +411,7 @@ namespace Google.PowerShell.CloudStorage
         /// </para>
         /// </summary>
         [Parameter(Position = 0, Mandatory = true, ParameterSetName = ParameterSetNames.FromName)]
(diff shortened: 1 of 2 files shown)


Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -476,29 +476,43 @@ protected override void ProcessRecord()
...
         /// Name of the object to read.
         /// </para>
         /// </summary>
-        [Parameter(Position = 1, Mandatory = true)]
+        [Parameter(Position = 1, Mandatory = true, ParameterSetName = ParameterSetNames.ByName)]
         public string ObjectName { get; set; }
 
         /// <summary>
         /// <para type="description">
+        /// The Google Cloud Storage bucket object to read.
```
chrsmith: Cloud Storage *object* right? Also, might clarify what types it could be. (String or Object object, right?)

"Cloud Storage object to read. May be a Storage Object instance or the name of a storage object (which requires the Bucket parameter be set as well)."

chrsmith:
Nit on the doc comment, otherwise LGTM.

PR #186, opened 2016-08-01
Replica testing

marceloyuela:
Finally tested creating a read replica instance.
Fixed some small issues in documentation.
Created the XML documentation for New-GcSqlInstanceReplicaConfig

### DIFF
--- a/Google.PowerShell.IntegrationTests/SQL/SQL.GcSqlInstances.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/SQL/SQL.GcSqlInstances.Tests.ps1
@@ -73,6 +73,24 @@ Describe "Add-GcSqlInstance" {
         $myInstance.Settings.MaintenanceWindow.Hour | Should Be 2
         gcloud sql instances delete $instance --quiet 2>$null
     }
+
+    It "should be able to create a read replica instance" {
+        $r = Get-Random
+        # A random number is used to avoid collisions with the speed of creating
+        # and deleting instances.
+        $instance = "test-inst$r"
+        $setting = New-GcSqlSettingConfig "db-n1-standard-1" 
+        $replicaConfig = New-GcSqlInstanceReplicaConfig
+        $config = New-GcSqlInstanceConfig $instance -SettingConfig $setting `
+            -ReplicaConfig $replicaConfig -MasterInstanceName "test-db2"
+        Add-GcSqlInstance $config
+        $newInstances = Get-GcSqlInstance
+        ($newInstances.Name -contains $instance) | Should Be true
+        $myInstance = Get-GcSqlInstance $instance
+        $myInstance.MasterInstanceName | Should Be "gcloud-powershell-testing:test-db2"
+        $myInstance.InstanceType | Should Be "READ_REPLICA_INSTANCE"
+        gcloud sql instances delete $instance --quiet 2>$null
+    }
 }
 
 Describe "Remove-GcSqlInstance" {
--- a/Google.PowerShell.IntegrationTests/SQL/SQL.GcSqlInstances.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/SQL/SQL.GcSqlInstances.Tests.ps1
@@ -57,7 +57,7 @@ Describe "Add-GcSqlInstance" {
         $newInstances = Get-GcSqlInstance
         ($instances.Name -contains $instance) | Should Be false
         ($newInstances.Name -contains $instance) | Should Be true
-        gcloud sql instances delete $instance --quiet 2>$null
+        Remove-GcSqlInstance $instance
     }
     
     It "should be able to reflect custom settings" {
@@ -71,10 +71,10 @@ Describe "Add-GcSqlInstance" {
         $myInstance = Get-GcSqlInstance $instance
         $myInstance.Settings.MaintenanceWindow.Day | Should Be 1
         $myInstance.Settings.MaintenanceWindow.Hour | Should Be 2
-        gcloud sql instances delete $instance --quiet 2>$null
+        Remove-GcSqlInstance $instance
     }
 
-    It "should be able to create a read replica instance" {
+    It "should be able to create a read-replica instance" {
         $r = Get-Random
         # A random number is used to avoid collisions with the speed of creating
         # and deleting instances.
@@ -84,12 +84,13 @@ Describe "Add-GcSqlInstance" {
         $config = New-GcSqlInstanceConfig $instance -SettingConfig $setting `
             -ReplicaConfig $replicaConfig -MasterInstanceName "test-db2"
         Add-GcSqlInstance $config
+
         $newInstances = Get-GcSqlInstance
         ($newInstances.Name -contains $instance) | Should Be true
         $myInstance = Get-GcSqlInstance $instance
         $myInstance.MasterInstanceName | Should Be "gcloud-powershell-testing:test-db2"
         $myInstance.InstanceType | Should Be "READ_REPLICA_INSTANCE"
-        gcloud sql instances delete $instance --quiet 2>$null
+        Remove-GcSqlInstance $instance
     }
 }
 
--- a/Google.PowerShell/Sql/GcSqlInstanceConfigCmdlets.cs
+++ b/Google.PowerShell/Sql/GcSqlInstanceConfigCmdlets.cs
@@ -90,7 +90,7 @@ namespace Google.PowerShell.Sql
         /// <summary>
         /// <para type="description">
         /// The name of the instance which will act as master in the replication setup. 
-        /// Should only be used for read replica instances.
+        /// Should only be used for read-replica instances.
         /// </para>
         /// </summary>
         [Parameter]
@@ -152,7 +152,7 @@ namespace Google.PowerShell.Sql
             }
             if (MasterInstanceName != null)
             {
-                //This should only be specified for read replica instances, so we make some adjustments.
+                // This should only be specified for read-replica instances, so we make some adjustments.
                 instance.Settings.BackupConfiguration.BinaryLogEnabled = false;
                 instance.Settings.BackupConfiguration.Enabled = false;
                 instance.InstanceType = "READ_REPLICA_INSTANCE";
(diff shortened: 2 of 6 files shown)


Google.PowerShell/Sql/GcSqlInstanceReplicaConfigCmdlet.cs
```diff
@@ -28,7 +57,7 @@ public class NewGcSqlInstanceReplicaConfigCmdlet : GcSqlCmdlet
         /// </para>
         /// </summary>
         [Parameter]
-        public SwitchParameter FailoverTarget { get; set; }
+        public bool FailoverTarget { get; set; } = false;
```
ILMTitan: Why did you make this  a false defaulted bool?

marceloyuela: Woops, this was something I did when debugging. Will fix

Google.PowerShell/Sql/GcSqlInstanceReplicaConfigCmdlet.cs
```diff
@@ -114,7 +115,7 @@ public class NewGcSqlInstanceReplicaConfigCmdlet : GcSqlCmdlet
         /// </para>
         /// </summary>
         [Parameter]
-        public SwitchParameter MySqlVerifyCertificate { get; set; }
+        public bool MySqlVerifyCertificate { get; set; } = false;
```
marceloyuela: Fragment from debugging. Will go back to switchparameter.

ILMTitan:
I am a little confused about why you changed FailoverTarget from a switch parameter to a bool that defaults to false.

marceloyuela:
Fixed that fragment from debugging.

Google.PowerShell.IntegrationTests/SQL/SQL.GcSqlInstances.Tests.ps1
```diff
@@ -73,6 +73,24 @@ Describe "Add-GcSqlInstance" {
         $myInstance.Settings.MaintenanceWindow.Hour | Should Be 2
         gcloud sql instances delete $instance --quiet 2>$null
     }
+
+    It "should be able to create a read replica instance" {
```
chrsmith: Should this be written as "read-replica"?

Google.PowerShell.IntegrationTests/SQL/SQL.GcSqlInstances.Tests.ps1
```diff
@@ -73,6 +73,24 @@ Describe "Add-GcSqlInstance" {
...
+        $replicaConfig = New-GcSqlInstanceReplicaConfig
+        $config = New-GcSqlInstanceConfig $instance -SettingConfig $setting `
+            -ReplicaConfig $replicaConfig -MasterInstanceName "test-db2"
+        Add-GcSqlInstance $config
+        $newInstances = Get-GcSqlInstance
+        ($newInstances.Name -contains $instance) | Should Be true
+        $myInstance = Get-GcSqlInstance $instance
+        $myInstance.MasterInstanceName | Should Be "gcloud-powershell-testing:test-db2"
+        $myInstance.InstanceType | Should Be "READ_REPLICA_INSTANCE"
+        gcloud sql instances delete $instance --quiet 2>$null
```
chrsmith: Have we implemented the corresponding cmdlet by now? Perhaps we should just use that?

Google.PowerShell.IntegrationTests/SQL/SQL.GcSqlInstances.Tests.ps1
```diff
@@ -73,6 +73,24 @@ Describe "Add-GcSqlInstance" {
...
+        $r = Get-Random
+        # A random number is used to avoid collisions with the speed of creating
+        # and deleting instances.
+        $instance = "test-inst$r"
+        $setting = New-GcSqlSettingConfig "db-n1-standard-1" 
+        $replicaConfig = New-GcSqlInstanceReplicaConfig
+        $config = New-GcSqlInstanceConfig $instance -SettingConfig $setting `
+            -ReplicaConfig $replicaConfig -MasterInstanceName "test-db2"
+        Add-GcSqlInstance $config
+        $newInstances = Get-GcSqlInstance
```
chrsmith: Maybe add a line of vertical whitespace here for readability? Currently this is a big wall of text.

Google.PowerShell/Sql/GcSqlInstanceConfigCmdlets.cs
```diff
@@ -149,6 +150,13 @@ protected override void ProcessRecord()
                     Name = FailoverReplica
                 };
             }
+            if (MasterInstanceName != null)
+            {
+                //This should only be specified for read replica instances, so we make some adjustments.
```
chrsmith: Space after `//`.

Google.PowerShell/Sql/GcSqlInstanceReplicaConfigCmdlet.cs
```diff
@@ -13,6 +13,35 @@ namespace Google.PowerShell.Sql
     /// Creates a configuration for a replicaConfiguration. 
     /// Can be pipelined into New-GcSqlInstanceConfig.
     /// </para>
+    /// <para>
+    /// WARNING: If a parameter passed in is bad, or wrong, this cmdlet will not error.
```
chrsmith: Having the WARNING is great. But perhaps we can reword this as "a parameter passed in is bad, or wrong" could just be "any parameter is incorrect.

Can we validate these parameters somehow? Is there a command we can issue that doesn't modify the project but will get the errors? If not, no worries.

Google.PowerShell/Sql/GcSqlInstanceReplicaConfigCmdlet.cs
```diff
@@ -13,6 +13,35 @@ namespace Google.PowerShell.Sql
...
+    /// <example>
+    ///   <para>
+    ///   Creates a basic replica configuration resource.
+    ///   </para>
+    ///   <para><code>
+    ///     PS C:\> New-GcSqlInstanceReplicaConfig
+    ///   </code></para>
+    ///   <br></br>
+    ///   <para>
+    ///   (If successful, the command returns a ReplicaConfiguration resource containing the replica configuration.)
```
chrsmith: Why not put this above on line 23?

Google.PowerShell/Sql/GcSqlInstanceReplicaConfigCmdlet.cs
```diff
@@ -78,10 +107,11 @@ public class NewGcSqlInstanceReplicaConfigCmdlet : GcSqlCmdlet
         /// <summary>
         /// <para type="description">
         /// Interval in milliseconds between replication heartbeats.
+        /// Defaults to 100.
```
chrsmith: I assume we specify the default here because the parameter is required in the underlying API?

100ms seems far too small. There is a performance cost of the overhead of a heartbeat. I'd put this to at least 5s. Maybe as high as 20 or 30s.

Google.PowerShell/Sql/GcSqlInstanceReplicaConfigCmdlet.cs
```diff
@@ -133,7 +163,8 @@ protected override void ProcessRecord()
                     Password = MySqlPassword,
                     SslCipher = MySqlSslCipher,
                     Username = MySqlUser,
-                    VerifyServerCertificate = MySqlVerifyCertificate
+                    VerifyServerCertificate = MySqlVerifyCertificate,
+                    Kind = "sql#mysqlReplicaConfiguration"
```
chrsmith: Why is this necessary? Seems like an API wart. We shouldn't have to deal with this.

Google.PowerShell/Sql/GcSqlSettingConfigCmdlet.cs
```diff
@@ -76,7 +76,7 @@ public enum ActivationPolicy
         /// <para type="description">
         /// Whether binary log is enabled.
         /// If backup configuration is disabled, binary log must be disabled as well.
-        /// Defaults to true;
+        /// Defaults to true for non replica instances.
```
chrsmith: non-replica instances.

Google.PowerShell/Sql/GcSqlSslCmdlets.cs
```diff
@@ -479,6 +480,10 @@ protected override void ProcessRecord()
                 default:
                     throw UnknownParameterSetException;
             }
+            if (!ShouldProcess($"{project}/{instance}", "Reset SSL Certificate Configuration"))
+            {
+                return;
```
chrsmith: Using  `if (!X) { return; }` should only be reserved if the following code block is very long. Since it is only 4-lines, I'd just put it in an `if (ShouldProcess(...)) {` block.

chrsmith:
Some more feedback.

Google.PowerShell.IntegrationTests/SQL/SQL.GcSqlInstances.Tests.ps1
marceloyuela: Won't that couple the testing of this cmdlet even more with a different cmdlet?  Is that okay?

Google.PowerShell/Sql/GcSqlInstanceReplicaConfigCmdlet.cs
marceloyuela: Will do, and will do for the rest of the XML docs as well

Google.PowerShell.IntegrationTests/SQL/SQL.GcSqlInstances.Tests.ps1
chrsmith: It is possible that you could have a cycle where cmdlet A calls cmdlet B calls cmdlet A, and if there is a bug in either of them then the test could be considered invalid.

However, I assume that `Remove-GcSqlInstance` has a separate set of tests that work great. And so it might make sense to just use that rather than rely on `gcloud`.

marceloyuela:
Should have fixed the above issues

marceloyuela:
Any other comments?

chrsmith:
LGTM if @ILMTitan is good with it.

ILMTitan:
LGTM

PR #188, opened 2016-08-01
Fix doc bugs

chrsmith:
Add `OutputType` attributes to cmdlets and add examples.

### DIFF
--- a/Google.PowerShell/Storage/GcsObject.cs
+++ b/Google.PowerShell/Storage/GcsObject.cs
@@ -492,7 +492,6 @@ namespace Google.PowerShell.CloudStorage
     /// </summary>
     [Cmdlet(VerbsCommon.Remove, "GcsObject",
         DefaultParameterSetName = ParameterSetNames.FromName, SupportsShouldProcess = true)]
-    [OutputType(new string[] { })]
     public class RemoveGcsObjectCmdlet : GcsCmdlet
     {
         private class ParameterSetNames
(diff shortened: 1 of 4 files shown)


chrsmith:
@ILMTitan , @emdat PTAL. Is this the type of thing you had in mind for fixing up our docs?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -662,8 +701,12 @@ protected override void ProcessRecord()
     /// Replaces the contents of a Cloud Storage object with data from the local disk or a value
     /// from the pipeline.
     /// </para>
+    /// <example>
+    ///   <para>Update the contents of the Storage Object with the string "OK".</para>
+    ///   <para><code>PS C:\> "OK" | Write-GcsObject -Bucket "widget-co-logs" -ObjectName "status.txt"</code></para>
+    /// </example>
     /// </summary>
-    [Cmdlet(VerbsCommunications.Write, "GcsObject")]
+    [Cmdlet(VerbsCommunications.Write, "GcsObject"), OutputType(new string[] { })]
```
ILMTitan: I would argue for leaving off OutputType entirely when the cmdlet does not output anything, as it adds not external value.

chrsmith: I'm inclined to agree. But I'd like us to at least do something to acknowledge that we thought about it. So that @emdat 's checker script doesn't complain.  Source-level linters allow for `// NOLINT` ([e.g.](https://en.wikipedia.org/wiki/Cpplint)).

But if push comes to shove, we could probably just curate a list of "false positives" and filter those out when we run the tool in the future. That would also prevent us from needing to write a lot of lame examples. e.g. `Remove-GcsBucket $bucket`.

chrsmith: I'll remove the `OutputType` attribute where it does not make sense. Logged issue #192 to support whitelisting.

chrsmith:
PTAL.

ILMTitan:
LGTM

PR #190, opened 2016-08-02
Fix website issues

chrsmith:
- Update GCE example to use the `Get-GceImage` cmdlet.
- Render type information in the cmdlet reference (escaping issue).
- Update cmdlets.json with the latest bits.


chrsmith:
@ILMTitan FYI. Since it's very minor, just going to go ahead and submit.

PR #205, opened 2016-08-04
Fix documentation and add examples.

ILMTitan:
Fix or add examples to most of the documentation.

### DIFF
--- a/Google.PowerShell/Compute/GceDisk.cs
+++ b/Google.PowerShell/Compute/GceDisk.cs
@@ -22,8 +22,8 @@ namespace Google.PowerShell.ComputeEngine
     /// such as Zone or Name, can be provided to restrict the objects returned.
     /// </para>
     /// <example>
+    ///   <code>PS C:\> Get-GceDisk -Project "ppiper-prod" "ppiper-frontend"</code>
     ///   <para>Get the disk named "ppiper-frontend".</para>
-    ///   <para><code>Get-GceDisk -Project "ppiper-prod" "ppiper-frontend"</code></para>
     /// </example>
     /// </summary>
     [Cmdlet(VerbsCommon.Get, "GceDisk")]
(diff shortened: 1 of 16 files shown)


Google.PowerShell/Compute/GceDisk.cs
```diff
@@ -133,6 +133,21 @@ protected void WriteDiskObjects(IEnumerable<Disk> disks)
     /// <para type="description">
     /// Creates a new Google Compute Engine disk object.
     /// </para>
+    /// <example>
+    ///   <code>PS C:\> New-GceDisk "disk-name" -SizeGb 10 -DiskType pd-ssd</code>
+    ///   <para>
+    ///     Creates a new empty 10GB persistant solid state disk named "disk-name" in the default project and
```
chrsmith: We wanted to put the descriptions before the code snippets, right?

Also, this includes the ">" character. If we can get away with that instead of &gt;, I'd prefer to keep it in.

Google.PowerShell/Compute/GceFirewallCmdlets.cs
```diff
@@ -32,7 +40,7 @@ public class GetGceFirewallCmdlet : GceCmdlet
 
         /// <summary>
         /// <para type="description">
-        /// The name of the firewall rule to get.
+        /// The name of the firewall rule to get. -Name and -Firewall are aliases of this parameter.
```
chrsmith: Why add the `-Alias` attribute? Shouldn't we prefer to have just one way to do things?

Google.PowerShell/Compute/GceInstanceCmdlets.cs
```diff
@@ -1101,14 +1201,10 @@ private void ProcessDisk()
...
-                if (diskParam is AttachedDisk || (diskParam as PSObject)?.BaseObject is AttachedDisk)
-                {
-                    newDisk = diskParam as AttachedDisk ?? (diskParam as PSObject)?.BaseObject as AttachedDisk;
-                }
-                else
+                var newDisk = diskParam as AttachedDisk ?? (diskParam as PSObject)?.BaseObject as AttachedDisk;
+                if (newDisk == null)
                 {
-                    Disk disk = diskParam as Disk ?? (diskParam as PSObject)?.BaseObject as Disk;
+                    var disk = diskParam as Disk ?? (diskParam as PSObject)?.BaseObject as Disk;
```
chrsmith: Should we tease this out into a method?

`Common.CoerceParameter(diskParam, typeof(Disk))`

The `as PSObject)?.` seems pretty sketch to copy and paste around a lot.

chrsmith:
A few comments and questions, but generally LGTM.

Note the merge conflit.

PR #207, opened 2016-08-05
Various fixes

marceloyuela:
Good catch Jim, that would have been a big mistake.
Notes from the (wrong) pull request:

The primary meat of this pull request is added Verbose capability for most cmdlets. The "get" cmdlets don't have it because they are much quicker and don't really have any steps to them.

All cmdlets that take time now display some progress.

The "Fixed Import Tests" commit happened because the instances lost SUPER Privilege for some reason (update with the API?), so the files being imported had to be changed slightly.

### DIFF
--- a/Google.PowerShell/Sql/GcSqlCmdlet.cs
+++ b/Google.PowerShell/Sql/GcSqlCmdlet.cs
@@ -63,6 +63,8 @@ namespace Google.PowerShell.Sql
                         break;
                     }
             }
+            WriteProgress(new System.Management.Automation.ProgressRecord(0, string.Format("Waiting for operation '{0}' to finish.", op.OperationType), "Waiting"));
+            int dots = 1;
             while (op.Status != "DONE")
             {
                 if (op.Error != null) {
@@ -70,6 +72,8 @@ namespace Google.PowerShell.Sql
                     return op;
                 }
                 Thread.Sleep(delay);
+                WriteProgress(new System.Management.Automation.ProgressRecord(0, string.Format("Waiting for operation '{0}' to finish.", op.OperationType), "Waiting" + new String('.',dots)));
+                dots = (dots + 1) % 4;
                 OperationsResource.GetRequest request = Service.Operations.Get(op.TargetProject, op.Name);
                 op = request.Execute();
             }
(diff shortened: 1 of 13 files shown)


marceloyuela:
Sorry for that other pull request- glad you caught the mistake @ILMTitan

Google.PowerShell/Sql/GcSqlInstanceCmdlets.cs
```diff
@@ -153,6 +152,7 @@ public class AddGcSqlInstanceCmdlet : GcSqlCmdlet
         protected override void ProcessRecord()
         {
             InstancesResource.InsertRequest request = Service.Instances.Insert(InstanceConfig, Project);
+            WriteVerbose(string.Format("Adding instance '{0}' to Project '{1}'.", InstanceConfig.Name, Project));
```
ILMTitan: I think we prefer the less verbose $"Adding instance '{InstanceConfig.Name}' to Project '{Project}'."

Google.PowerShell/Sql/GcSqlCmdlet.cs
```diff
@@ -63,13 +63,17 @@ public Operation WaitForSqlOperation(Operation op)
                         break;
                     }
             }
+            WriteProgress(new System.Management.Automation.ProgressRecord(0, string.Format("Waiting for operation '{0}' to finish.", op.OperationType), "Waiting"));
```
ILMTitan: Have you considered moving the progress bar rather than changing the text?

ILMTitan:
I see a lot of string.Format that could probably simplified. Also, perhaps we should move the progress bar rather than changing the text. It would be similar, but instead of %=4 we would have a %=100.

Google.PowerShell/Sql/GcSqlCmdlet.cs
marceloyuela: Will look into this.

ILMTitan:
I am concerned the progress bar will not disappear without writing a ProgressRecord of RecordType = Complete. If you are sure it does, LGTM.

marceloyuela:
Updated that just in case. It seemed to complete regardless but I'd rather not risk sing up extra resources.

ILMTitan:
LGTM

PR #208, opened 2016-08-05
Add cmdlets to query load balancing resources.

ILMTitan:
Add cmdlets to query load balancing resources, including target pools, target proxies, frontends, backend services and url maps.


Google.PowerShell.IntegrationTests/Compute/Compute.GceForwardingRule.Tests.ps1
```diff
@@ -0,0 +1,87 @@
+﻿. $PSScriptRoot\..\GcloudCmdlets.ps1
+Install-GcloudCmdlets
+
+$project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
+$r = Get-Random
+Describe "Get-GceForwardingRule"{
```
chrsmith: Nit: Vertical whitespace here for readability, and again on line 10.

Google.PowerShell.IntegrationTests/Compute/Compute.GceForwardingRule.Tests.ps1
```diff
@@ -0,0 +1,87 @@
...
+            gcloud compute url-maps delete "url-map-$r" -q 2>$null
+            gcloud compute backend-services delete "backend-$r" -q 2>$null
+            gcloud compute http-health-checks delete "health-check-$r" -q 2>$null
+        }
+
+
+    }
+}
+
+
```
chrsmith: Nit: Only need one line of whitespace here.

Google.PowerShell.IntegrationTests/Compute/Compute.GceTargetProxy.Tests.ps1
```diff
@@ -0,0 +1,67 @@
...
+        { Get-GceTargetProxy $httpProxyName } | Should Throw "Can not find target proxy"
+    }
+
+    Context "with data" {
+        BeforeAll {
+            gcloud compute http-health-checks create "health-check-$r" 2>$null
+            gcloud compute backend-services create "backend-$r" --http-health-check "health-check-$r" 2>$null
+            gcloud compute url-maps create "url-map-$r" --default-service "backend-$r" 2>$null
+            gcloud compute target-http-proxies create $httpProxyName --url-map "url-map-$r" 2>$null
+            #TODO(jimwp) make this a target-https-proxy (create self signed certificate)
```
chrsmith: `# TODO(jimwp): `

- Space after `#`.
- Colon after `)`.

Google.PowerShell/Compute/GceBackendServiceCmdlets.cs
```diff
@@ -0,0 +1,86 @@
...
+        /// </summary>
+        [Parameter(ParameterSetName = ParameterSetNames.ByName, Mandatory = true, Position = 0)]
+        public string Name { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            switch (ParameterSetName)
+            {
+                case ParameterSetNames.ByName:
+                    WriteObject(Service.BackendServices.Get(Project, Name).Execute());
```
chrsmith: I'd put the API call on a separate line. Personally I'd use 2-3 for clarity.

Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -124,6 +123,20 @@ public static string GetZoneNameFromUri(string uri)
         {
             return GetUriPart("zones", uri);
         }
+
+        /// <summary>
+        /// Library method to pull the name of a zone from a uri.
```
chrsmith: URI is an acronym.

Google.PowerShell/Compute/GceForwardingRuleCmdlets.cs
```diff
@@ -0,0 +1,174 @@
...
+    /// <code>PS C:\> Get-GceForwardingRule "my-forwarding-rule"</code>
+    /// <para>Gets the forwarding rule named "my-forwarding-rule" in the default project and region</para>
+    /// </example>
+    /// <example>
+    /// <code>PS C:\> Get-GceForwardingRule -Project my-project -Global</code>
+    /// <para>Lists all global forwarding rules for the project named "my-project".</para>
+    /// </example>
+    /// <example>
+    /// <code>PS C:\> Get-GceForwardingRule "my-forwarding-rule" -Gobal</code>
+    /// <para>Gets the global forwarding rule named "my-forwarding-rule" in the default project</para>
```
chrsmith: End this snippet with a period, and all previous text snippets as well. Is there any reason not to use complete senences?

Google.PowerShell/Compute/GceForwardingRuleCmdlets.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// <para type="description">
+        /// The project the forwarding rules belong to. Defaults to the project in the Cloud SDK config.
+        /// </para>
+        /// </summary>
+        [Parameter(ParameterSetName = ParameterSetNames.OfProject)]
+        [Parameter(ParameterSetName = ParameterSetNames.OfRegion)]
+        [Parameter(ParameterSetName = ParameterSetNames.ByLocalName)]
+        [Parameter(ParameterSetName = ParameterSetNames.ByGlobalName)]
+        [ConfigPropertyName(CloudSdkSettings.CommonProperties.Project)]
+        public string Project { get; set; }
```
chrsmith: Since this parameter is used by all four parameter sets, can we just put `[Parameter]` on it, and not need to explicitly call out each parameter set?

Google.PowerShell/Compute/GceForwardingRuleCmdlets.cs
```diff
@@ -0,0 +1,174 @@
...
+                    break;
+                case ParameterSetNames.ByGlobalName:
+                    WriteObject(Service.GlobalForwardingRules.Get(Project, Name).Execute());
+                    break;
+                default:
+                    throw UnknownParameterSetException;
+            }
+        }
+
+        private IEnumerable<ForwardingRule> GetRegionForwardingRules()
```
chrsmith: This code is fine, but consider making it stateless. i.e. pass the project and region as parameters, so that the function does not depend on `this`. It's a good principle as it makes code easier to unit test, etc. But feel free to disregard in this situation.

Google.PowerShell/Compute/GceTargetPoolCmdlets.cs
```diff
@@ -0,0 +1,323 @@
...
+
+        /// <summary>
+        /// <para type="description">
+        /// The name of the target pool to get.
+        /// </para>
+        /// </summary>
+        [Parameter(ParameterSetName = ParameterSetNames.ByName, Mandatory = true, Position = 0)]
+        public string Name { get; set; }
+
+        // TODO(jimwp) Understand what this health status is an uncomment this parameter.
```
chrsmith: Colon after `)`. Also, can you elaborate? "what this health status is an"? What do you need to do to resolve this TODO?

Also, don't leave commented out code (//[Parameter]). Just have the TODO say, "figure out if this should be a Parameter" or something.

And finally, what is the meaning of having a `SwitchParameter` type that is not actually used? Should we just delete the public property as well?

chrsmith:
LGTM. Though the number of unit tests seems a little light for the surface area of cmdlets we are adding. Though without the "read/write" cmdlets, I'm not sure what we can do. Perhaps we need to think about end-to-end integration tests outside of Pester?

Google.PowerShell/Compute/GceTargetPoolCmdlets.cs
ILMTitan: That was supposed to read "Understand ... and uncomment this parameter." Basically, gcloud compute target-pools get-health is a commonly called command that "get(s) the health of instances in a target pool". I don't know exactly what that means, and my experiments with it never returned a value, so I don't know how to test if the cmdlet version is working. For now I will remove the switch property entirely, but keep the now orphaned method.

Issue #181, opened 2016-07-28
Need progress indicator for long running operations [userstudy]


YnR:
If operation will take more than 1s, show some kind of progress indicator, such as for:
Uploading files
Spinning up new VMs
Attaching disks
...

ILMTitan:
There is a way to write to a progress bar, but for most of those things, there is no feedback on how progress is proceeding.

chrsmith:
Yes, Jim is correct. The GCE API only returns an "operation" object that lets us know if it has failed, completed, or is still in-progress.

For operations where we do know progress -- IIRC file uploads or bulk-copying GCS objects -- we can use the built-in PowerShell progress indication.

@ILMTitan , taking a closer look it appears we only do this for `Remove-GcsObject` ([here](https://github.com/GoogleCloudPlatform/google-cloud-powershell/blob/4c7758e204dd2da32f7cade51820eed0bef519b8/Google.PowerShell/Storage/GcsBucket.cs#L288)). Does it make sense to also apply a progress bar when you pipe multiple objects to `New-GcsObject` too? Are there other cmdlets where we should wire in a progress bar?

ILMTitan:
One thing I am thinking about is having a progress bar that moves forward,
and resets to zero at some point, just to show we haven't hung.

chrsmith:
Jim, did you look into this? Are there some specific cmdlets we should wire progress into? Or just the infinite spinner?

YnR:
Seems like there is a progress bar that is often used in powershell, the yellow-on-teal bar that appears at the top. Could we use that?

ILMTitan:
Closed in #213.

PR #213, opened 2016-08-08
Issue 181 add progress spinner

ILMTitan:
Adds a progress bar that acts as a spinner.

### DIFF
--- a/Google.PowerShell/Compute/GceCmdlet.cs
+++ b/Google.PowerShell/Compute/GceCmdlet.cs
@@ -133,21 +133,7 @@ namespace Google.PowerShell.ComputeEngine
         {
 
             int activityId = op.Id.GetHashCode();
-            string activity = progressMessage ?? op.Description;
-            if (activity == null)
-            {
-                string operationType = op.OperationType.Substring(0, 1).ToUpper()
-                    + op.OperationType.Substring(1);
-
-                string target = op.TargetLink;
-                const string baseUri = "https://www.googleapis.com/compute/v1/";
-                if (target.StartsWith(baseUri))
-                {
-                    target = target.Substring(baseUri.Length);
-                }
-
-                activity = operationType + " " + target;
-            }
+            string activity = progressMessage ?? op.Description ?? BuildActivity(op);
             string statusDescription = op.StatusMessage ?? op.Status;
             int percentComplete;
             if (op.Progress == null || op.Progress == 0)
@@ -167,13 +153,35 @@ namespace Google.PowerShell.ComputeEngine
         }
 
         /// <summary>
+        /// Builds an activity. This is displayed on the top line of the progress bar.
+        /// </summary>
+        /// <param name="op">The operation to build an activity from.</param>
+        /// <returns>The type and target of the operation </returns>
+        private static string BuildActivity(Operation op)
+        {
+            // Capitalize the operation type.
+            string operationType = op.OperationType.Substring(0, 1).ToUpper()
+                                   + op.OperationType.Substring(1);
+
+            // Cut the target uri down to the interesting data.
+            string target = op.TargetLink;
+            const string baseUri = "https://www.googleapis.com/compute/v1/";
+            if (target.StartsWith(baseUri))
+            {
+                target = target.Substring(baseUri.Length);
+            }
+
+            return $"{operationType} {target}";
+        }
+
+        /// <summary>
         /// Closes the progress bar of the operation.
         /// </summary>
         /// <param name="op">The operation the progress bar was created for.</param>
         private void WriteProgressComplete(Operation op)
         {
             int activityId = op.Id.GetHashCode();
-            string activity = op.Description ?? op.OperationType;
+            string activity = op.Description ?? op.OperationType ?? BuildActivity(op);
             string statusDescription = op.StatusMessage ?? op.Status;
             ProgressRecord record = new ProgressRecord(activityId, activity, statusDescription)
             {
--- a/Google.PowerShell/Compute/GceCmdlet.cs
+++ b/Google.PowerShell/Compute/GceCmdlet.cs
@@ -20,8 +20,9 @@ namespace Google.PowerShell.ComputeEngine
         // The Servcie for the Google Compute API
         public ComputeService Service { get; }
 
-        // The value of the progress bar when being used as a spinner.
-        private int _spinnerValue;
+        // The value of the progress bar when being used as a spinner. If the operation does not give valid
+        // percent complete feedback, this spinner will move from 0 to 99 and then repeat.
+        private int _spinnerValue = 0;
 
         protected GceCmdlet() : this(null)
         {
@@ -50,7 +51,7 @@ namespace Google.PowerShell.ComputeEngine
                 WriteWarnings(op);
                 WriteProgress(op, progressMessage);
             }
-            EndProgress(op);
+            WriteProgressComplete(op);
 
             if (op.Error != null)
             {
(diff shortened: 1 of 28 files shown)


Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -19,6 +20,9 @@ public abstract class GceCmdlet : GCloudCmdlet
         // The Servcie for the Google Compute API
         public ComputeService Service { get; }
 
+        // The value of the progress bar when being used as a spinner.
+        private int _spinnerValue;
```
chrsmith: When being used as a spinner? Can it be used as anything else?

Why not, "Value of the current progress bar for cmdlets which have an infinite spinner to fake progress."

Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -110,6 +122,51 @@ private void WriteWarnings(Operation op)
             }
         }
 
+        private void WriteProgress(Operation op, string progressMessage)
```
chrsmith: Definitely needs a comment explaining what it does and what it is used for.

Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -110,6 +122,51 @@ private void WriteWarnings(Operation op)
...
+        {
+            int activityId = (int)(op.Id % int.MaxValue ?? 0);
```
chrsmith: I assume `op.Id` is an int64? What is the `?? 0` needed for? What does that even mean when dealing with integers?

Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -110,6 +122,51 @@ private void WriteWarnings(Operation op)
...
+            string activity = progressMessage ?? op.Description;
+            if (activity == null)
+            {
+                string operationType = op.OperationType;
+                operationType = operationType.First().ToString().ToUpper() + operationType.Substring(1);
+                int startIndex = op.TargetLink.IndexOf("projects", StringComparison.Ordinal);
```
chrsmith: There is far too much magic going on here. We cannot assume that "projects" is part of the internal naming mechanism of the operations. To do so would be taking a dependency on an internal implementation detail.

Instead, we should just provide a lame, generic message like "processing". If it is lame, then the cmdlet author should wire in a more meaningful name.

Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -110,6 +122,51 @@ private void WriteWarnings(Operation op)
...
+                {
+                    startIndex = 0;
+                }
+                activity = operationType + " " + op.TargetLink.Substring(startIndex);
+            }
+            string statusDescription = op.StatusMessage ?? op.Status;
+            int percentComplete;
+            if (op.Progress == null || op.Progress == 0)
+            {
+                percentComplete = _spinnerValue = (_spinnerValue + 1) % 100;
```
chrsmith: This seems kinda lame that the spinner is dependent upon how often `WriteProgress` is called. So different cmdlets will "progress" at different rates depending on arbitrarily chosen sleep intervals.

Of course, I can't think of anything better short of requiring that you explicitly start and stop the faux progress bar.

Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -110,6 +122,51 @@ private void WriteWarnings(Operation op)
...
+            {
+                RecordType = ProgressRecordType.Processing,
+                PercentComplete = percentComplete,
+            };
+            WriteProgress(record);
+        }
+
+        private void EndProgress(Operation op)
+        {
+            int activityId = (int)(op.Id % int.MaxValue ?? 0);
```
chrsmith: 1. We need to explicitly initialize `_spinnerValue` to 0. Yes, the CLR spec says it will default to 0, but we are better than that.
2. When a progress bar ends, we need to reset the value of `_spinnerValue` to 0.

chrsmith:
The design doesn't quite seem right.

- We have an `EndProgress` method but no `BeginProgress` method. But it seems like we should to initialize `_spinnerValue`.
- What if the Global/Zone operations are known to be very fast? In which case we would just spam a bunch of very short-lived progress bars. It seems like we should be able to enable/disable the progress bar via parameter to `WaitForOperation`. Might I suggest: if you do not specify an operation name, you do not get a progress bar. If you do, then you get a progress bar.

As for the feature itself, I am excited to see this. I'd love to see a demo.

Google.PowerShell/Compute/GceCmdlet.cs
ILMTitan: `op.Id` is indeed an `int64?`, meaning it could be null.

chrsmith: Why not:

    int activityId = opt.Id.GetValueOrDefault() % int.MaxValue;

ILMTitan:
I don't think we need to worry about very fast operations. If they were very fast, then they wouldn't be operations; the api would just return them.

Google.PowerShell/Compute/GceCmdlet.cs
ILMTitan: Because this is in GceCmdlet, all of our sleep intervals is 150ms.

ILMTitan:
PTAL

Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -111,6 +124,66 @@ private void WriteWarnings(Operation op)
...
+        /// Writes a progress record to give feedback to the user.
+        /// </summary>
+        /// <param name="op">The operation we are waiting on.</param>
+        /// <param name="progressMessage">The custom message to write to the progress bar. If null, this
+        ///  method will generate a message from the operation.</param>
+        private void WriteProgress(Operation op, string progressMessage)
+        {
+
+            int activityId = op.Id.GetHashCode();
+            string activity = progressMessage ?? op.Description;
```
chrsmith: What if `op.Description` is null too? (Is that ever the case?) Would it make sense to append `?? "Processing";` as a catch-all?

Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -111,6 +124,66 @@ private void WriteWarnings(Operation op)
...
+            if (activity == null)
+            {
+                string operationType = op.OperationType.Substring(0, 1).ToUpper()
```
chrsmith: This code needs a comment, since it takes (a slow person like myself) > 0.1s to understand why.

Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -111,6 +124,66 @@ private void WriteWarnings(Operation op)
...
+                    + op.OperationType.Substring(1);
+
+                string target = op.TargetLink;
+                const string baseUri = "https://www.googleapis.com/compute/v1/";
+                if (target.StartsWith(baseUri))
+                {
+                    target = target.Substring(baseUri.Length);
+                }
+
+                activity = operationType + " " + target;
```
chrsmith: Consider using the string interpolation syntax, which would clean this up IMHO.

`$"{operationType} {target}"`

Google.PowerShell/Compute/GceCmdlet.cs
chrsmith: That is assuming that Evil Jim, who will join the team later, reuses the same constant value. But Evil Jim won't do that. Instead, Evil Jim will have his cmdlets inherit from `EvilGceCmdlet` which have a sleep interval of 149ms. Why? Because Evil Jim is evil like that.

Since we won't be adding many more GCE cmdlets in the future, this is fine. But if there was a lot of future expansion, we'd need a stronger guarantee or a better solution for this.

chrsmith:
LGTM

ILMTitan:
Close #181.

Issue #164, opened 2016-07-25
Add ability to read/write GCS object metadata


chrsmith:
GCS objects can have metadata stored as key/value pairs. Provide a way to read/write this.

chrsmith:
Should be fixed now.

Issue #201, opened 2016-08-04
Gcs - Invalid tags for parameter descriptions


emdat:
In some of the Gcs cmdlets, the parameter descriptions are introduced with a tag saying `<para typedef="description">`, which prevents the description from being parsed properly. It should work if you just make it `<para type="description"> </para>`.

chrsmith:
https://github.com/GoogleCloudPlatform/google-cloud-powershell/search?utf8=%E2%9C%93&q=typedef

chrsmith:
Should be fixed now.

PR #214, opened 2016-08-09
Enable reading/writing GCS object metadata

chrsmith:
This PR adds the ability to write/update GCS object metadata in the `New-GcsObject` and `Write-GcsObject` cmdlets. (You were always able to read it.)

This PR was a huge PITA because the GCS API has two places you can specify the `ContentType` for an object. So extra work is done to make sure that we always prefer the content type you specify via the parameter, and ignore the value from metadata if there is a conflict.

Also, PowerShell dictionaries are of the `Hashtable` which are subtltey different than the type we use internally, `Dictionary<string, string>`, and what the API expects, `IDictionary<string, string>`. However, a few convenience methods were able to blur those lines sufficiently.

This PR also updates the NuGet package versions. Pester was not able to run the tests until the references were updated. (I cannot explain why that is the case.)

Fixes issue #164, and #201.

### DIFF
--- a/Google.PowerShell/Storage/GcsCmdlet.cs
+++ b/Google.PowerShell/Storage/GcsCmdlet.cs
@@ -115,13 +115,12 @@ namespace Google.PowerShell.CloudStorage
         /// 1. New content type, e.g. a ContentType parameter.
         /// 2. New metadata, e.g. Metadata value specified via parameter.
         /// 3. Default content type to apply (potentially null), e.g. sniffing file content.
-        /// 4. Default content type to apply. e.g. a catch-all like octet-stream.
+        /// If no match is found, will return OctetStreamMimeType.
         /// </summary>
         public string GetContentType(
             string newContentType,
             Dictionary<string, string> newMetadata,
-            string defaultContentType1 = null,
-            string defaultContentType2 = null)
+            string defaultContentType = null)
         {
             if (!String.IsNullOrEmpty(newContentType))
             {
@@ -133,12 +132,12 @@ namespace Google.PowerShell.CloudStorage
                 return newMetadata["Content-Type"];
             }
 
-            if (!String.IsNullOrEmpty(defaultContentType1))
+            if (!String.IsNullOrEmpty(defaultContentType))
             {
-                return defaultContentType1;
+                return defaultContentType;
             }
 
-            return defaultContentType2;
+            return OctetStreamMimeType;
         }
     }
 }
--- a/Google.PowerShell/Storage/GcsCmdlet.cs
+++ b/Google.PowerShell/Storage/GcsCmdlet.cs
@@ -4,6 +4,7 @@
 using Google.Apis.Storage.v1;
 using Google.PowerShell.Common;
 using System;
+using System.Collections;
 using System.Collections.Generic;
 
 namespace Google.PowerShell.CloudStorage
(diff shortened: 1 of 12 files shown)


chrsmith:
@ILMTitan PTAL

Google.PowerShell/Storage/GcsCmdlet.cs
```diff
@@ -39,5 +40,111 @@ protected string GetBaseUri(string bucket, string objectName)
...
+        /// 1. New content type, e.g. a ContentType parameter.
+        /// 2. New metadata, e.g. Metadata value specified via parameter.
+        /// 3. Existing object, to keep its existing content-type
+        /// 4. Default content type #1 to apply. (e.g. sniffing file content.)
+        /// 5. Default content type #2 to apply. (e.g. a catch-all like octet-stream.)
+        /// </summary>
+        public string GetContentType(
+            string newContentType,
+            Dictionary<string, string> newMetadata,
+            Google.Apis.Storage.v1.Data.Object existingObject,
```
ILMTitan: It feels like these last three parameters should just be one.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -783,17 +820,28 @@ protected override void ProcessRecord()
                 contentStream = new MemoryStream(contentBuffer);
             }
 
+            // Get the existing storage object so we can use its metadata. (If it does not exist, we will fall back to
+            // default values.)
+            Object existingGcsObject = null;
+            Dictionary<string, string> existingObjectMetadata = null;
```
ILMTitan: I dislike these dead null initializations, but I can agree to disagree.

ILMTitan:
I feel like get content type should have just three parameters, the new type, the metadata, and a single default. As it is now, every call to it has at least one null being passed in, which is ugly.

Google.PowerShell/Storage/GcsObject.cs
chrsmith: It's a fair point, the "= null;" is not only optional, it is 100% redundant. But so is the semicolon (generally) in JavaScript.

But I've written enough C++ to never trusted an uninitialized variable. You might say I am broken, and need to be more trusting of the C# runtime. But I've seen things, man.

chrsmith:
@ILMTitan PTAL. Cleaned up the `GetContentType` method. It is still pretty lame. But at least now it is less lame.

Google.PowerShell/Storage/GcsObject.cs
ILMTitan: The "= null;" is not redundant. It actually means something different. A declared variable may not be read from until it has been initialized. The compiler will throw an error if you fail to initialize existingGcsObject on all code paths.

ILMTitan:
Interestingly, when `GetContentType` takes 4 non-null parameters, the fourth is always `OctetStreamMimeType`. Would it be reasonable to hard code that as the fourth parameter, bringing the function down to three?
Also, you don't need to trust that the runtime will initialize declared variables to null. Instead trust the compiler to yell at you if you try to use an uninitialized variable. (Try commenting out the throw in the default case of a switch statement).

chrsmith:
Ack on the initializing variables before read error. I'll keep that in mind.

As for the 4th parameter (!!) of `GetContentType`, unfortunately it isn't always OctetStream. When you pass text in, we default to `text/plain`. However, in the one place where we use text/plain, we are not using the other three parameters. So I made the change, good suggestion.

ILMTitan:
LGTM

PR #215, opened 2016-08-09
Add cmdlet to query the metadata server from a VM instance.

ILMTitan:
Adds Get-GceMetadata, which retrieves metadata for the VM instance it is being called on.

### DIFF
--- a/Google.PowerShell/Compute/GceMetadataCmdlet.cs
+++ b/Google.PowerShell/Compute/GceMetadataCmdlet.cs
@@ -22,7 +22,7 @@ namespace Google.PowerShell.ComputeEngine
     /// [Metadata Documentation]
     /// </para>
     /// <example>
-    /// <code>PS C:\> $allMetadata = Get-GceMetadata -Recursive | ConvertFrom-Json</code>
+    /// <code>PS C:\> $allMetadata = Get-GceMetadata -Recurse | ConvertFrom-Json</code>
     /// <para>Gets all of the metadata and converts it from the given JSON to a PSCustomObject.</para>
     /// </example>
     /// <example>
@@ -34,7 +34,7 @@ namespace Google.PowerShell.ComputeEngine
     /// <para>Gets the value of the custom metadata with key "customKey" placed in the project .</para>
     /// </example>
     /// <example>
-    /// <code>PS C:\> $metadata, $etag = Get-GceMetadata -AppendETag -Recursive</code>
+    /// <code>PS C:\> $metadata, $etag = Get-GceMetadata -AppendETag -Recurse</code>
     /// <para>Gets the entire metadata tree, and the ETag of the version retrieved.</para>
     /// </example>
     /// <example>
@@ -65,14 +65,14 @@ namespace Google.PowerShell.ComputeEngine
         /// </para>
         /// </summary>
         [Parameter]
-        public SwitchParameter Recursive { get; set; }
+        public SwitchParameter Recurse { get; set; }
 
         /// <summary>
         /// <para type="description">
         ///   When set, the value of the respone ETag will be appended to the output pipeline after the content.
         /// </para>
         /// <para type="description">
-        /// <code>PS C:\> $metadata, $etag = Get-GceMetadata -AppendETag -Recursive</code>
+        /// <code>PS C:\> $metadata, $etag = Get-GceMetadata -AppendETag -Recurse</code>
         /// </para>
         /// <para type="description">
         ///   Gets the entire metadata tree, and the ETag of the version retrieved.
@@ -92,7 +92,7 @@ namespace Google.PowerShell.ComputeEngine
         /// <summary>
         /// <para type="description">
         /// The last ETag known. Used in conjunction with -WaitUpdate. If the last ETag does not match the
-        /// current ETag of the metadata server, it will return immediatly.
+        /// current ETag of the metadata server, it will return the updated value immediatly.
         /// </para>
         /// </summary>
         [Parameter]
@@ -103,17 +103,23 @@ namespace Google.PowerShell.ComputeEngine
         /// Used in conjunction with -WaitUpdate. The amout of time to wait. If the timeout expires, the 
         /// current metadata will be returned.
         /// </para>
+        /// <para type="description">
+        /// Check the ETag using AppendEtag to see if the data was updated within the timeout period.
+        /// </para>
         /// </summary>
         [Parameter]
         public TimeSpan? Timeout { get; set; }
 
+        /// <summary>
+        /// Make this a field, so it can be aborted if cmdlet stops e.g. the user hits Ctrl-C.
+        /// </summary>
         private HttpWebRequest _request;
 
         protected override void ProcessRecord()
         {
             const string basePath = "http://metadata.google.internal/computeMetadata/v1/";
             var queryParameters = new List<string>();
-            if (Recursive)
+            if (Recurse)
             {
                 queryParameters.Add("recursive=true");
             }
@@ -166,7 +172,7 @@ namespace Google.PowerShell.ComputeEngine
 
         protected override void StopProcessing()
         {
-            _request.Abort();
+            _request?.Abort();
             base.StopProcessing();
         }
     }
--- a/Google.PowerShell/Compute/GceMetadataCmdlet.cs
+++ b/Google.PowerShell/Compute/GceMetadataCmdlet.cs
@@ -1,0 +1,65 @@
+﻿// Copyright 2016 Google Inc. All Rights Reserved.
+// Licensed under the Apache License Version 2.0.
+
+using Google.PowerShell.Common;
+using Newtonsoft.Json;
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Management.Automation;
+using System.Net;
+
+namespace Google.PowerShell.ComputeEngine
+{
+    [Cmdlet(VerbsCommon.Get, "GceMetadata")]
+    public class GetGceMetadataCmdlet : GCloudCmdlet
+    {
+        [Parameter]
+        public string Path { get; set; } = "";
+
+        [Parameter]
+        public SwitchParameter WaitUpdate { get; set; }
+
+        [Parameter]
+        public string ETag { get; set; }
+
+        [Parameter]
+        public TimeSpan? Timeout { get; set; }
+
+        [Parameter]
+        public SwitchParameter NotRecursive { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            const string basePath = "http://metadata.google.internal/computeMetadata/v1/";
+            var queryParameters = new List<string>();
+            if (!NotRecursive)
+            {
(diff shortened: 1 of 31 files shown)


Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// <code>PS C:\> $hostName = Get-GceMetadata -Path "instance/hostname" -NotRecursive</code>
+    /// <para>Gets the hostname of the instance.</para>
+    /// </example>
+    /// <example>
+    /// <code>
+    ///     PS C:\> $newTags, $newEtag = Get-GceMetadata -Path "instance/tags" -AppendEtag -WaitUpdate `
+    ///         -LastETag $oldETag
+    /// </code>
+    /// </example>
+    /// </summary>
```
chrsmith: Could you use the XML Doc "link" or whatever so that we can point users to more context? e.g. the documentation for the metadata server and how it works on GCE.

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// <para type="synopsis">
+    /// Gets the current metadata from the metadata server.
+    /// </para>
+    /// <para type="description">
+    /// Gets the current metadata from the metadata server. Get-GceMetadata can only be called from a Google
+    /// Compute Engine VM instance. Calls from any other machine will fail.
+    /// </para>
+    /// <example>
+    /// <code>PS C:\> $allMetadata = Get-GceMetadata | ConvertFrom-Json</code>
+    /// <para>Gets all of the metadata and converts it from the given json to a PSCustomObject.</para>
```
chrsmith: JSON is an acronym, it deserves some capital letters. Please, honor JSON with your shift key.

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+    /// </example>
+    /// <example>
+    /// <code>PS C:\> $hostName = Get-GceMetadata -Path "instance/hostname" -NotRecursive</code>
+    /// <para>Gets the hostname of the instance.</para>
```
chrsmith: Could you provide another example for getting a single value?

e.g. if I put a password into the metadata server, how would I get it?  `-Path "authtoken"`

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+    public class GetGceMetadataCmdlet : GCloudCmdlet
+    {
+        /// <summary>
+        /// <para type="description">
+        /// The path to the specific metadata you wish to get e.g. "instance/tags", "instance/attributes",
+        /// "project/attributes/sshKeys".
+        /// </para>
+        /// </summary>
+        [Parameter]
+        public string Path { get; set; } = "";
```
chrsmith: Why the default value? Is that so the "Get them all" case works without special casing? Perhaps we should add a comment for future code archeologists?

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        /// <summary>
+        /// <para type="description">
+        /// If set, will only get the direct children of the current path.
```
chrsmith: Can you elaborate on this? The casual user looking into this might not know what "direct children" referrs to. Perhaps a quick example "e.g. XXX".

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// </para>
+        /// </summary>
+        [Parameter]
+        public SwitchParameter NotRecursive { get; set; }
```
chrsmith: A best practice is to make flags always be positive. (This avoids the insanity that is `-NotRecursive = false`.) Can we just name this `Recurse` (which would be consistent with other PowerShell cmdlets) and do the work to invert it when we use it internally?

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+
+        /// <summary>
+        /// <para type="description">
+        /// When set, the value of the respone etag will be appended to the output pipeline after the content.
```
chrsmith: ETag.

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// </para>
+        /// </summary>
+        [Parameter]
+        public SwitchParameter AppendETag { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// If true, the query will wait for the metadata to update.
```
chrsmith: Can you elaborate on this?

"Will wait until -Timeout for the metadata value to be updated. After which time a XXX exception will be thrown."

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// </para>
+        /// </summary>
+        [Parameter]
+        public SwitchParameter WaitUpdate { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The last etag known. Used in conjunction with -WaitUpdate. If the last etag does not match the
```
chrsmith: ETag.

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+        /// current etag of the metadata server, it will return immediatly.
+        /// </para>
+        /// </summary>
+        [Parameter]
+        public string LastETag { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// Used in conjunction with -WaitUpdate. The amout of time to wait before returning.
```
chrsmith: before returning, or before failing? It isn't clear.

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+        [Parameter]
+        public SwitchParameter NotRecursive { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// When set, the value of the respone etag will be appended to the output pipeline after the content.
+        /// </para>
+        /// </summary>
+        [Parameter]
+        public SwitchParameter AppendETag { get; set; }
```
chrsmith: Can you provide an example of how you can actually use this? Because it doesn't seem possible.

    $metadataValue = Get-GceMetadata ... -AppendETag

Will `$metadataValue` have the value of the metadata or the ETag value? Is it possible to get both, a la:

    $metadata, $etag = Get-GceMetdata ... -AppendETag

If not, then perhaps we should have it return a dictionary in that case.

If `-AppendETag` is not set, just return a string. Otherwise return a dictionary: `@{ "Value"=...; "ETag"=...}`.

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,131 @@
...
+
+            string query = string.Join("&", queryParameters);
+            HttpWebRequest request = WebRequest.CreateHttp($"{basePath}{Path}?{query}");
+            request.Headers.Add("Metadata-Flavor:Google");
+            if (WaitUpdate)
+            {
+                request.Timeout = -1;
+            }
+
+            using (WebResponse response = request.GetResponse())
```
chrsmith: What is the error case if the metadata server is unavailable? We should intercept the 404 and 403 cases and provide a better error message.

Especially the "GCE Metadata server was not found. This cmdlet can only be used from within a Compute Engine VM."

Google.PowerShell/Compute/GceMetadataCmdlet.cs
ILMTitan: Turns out, it was unnecessary.

ILMTitan:
PTAL

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,173 @@
...
+        public string Path { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// If set, will get the metadata subtree as a JSON string. If -Path is not set, will get the entire
+        /// metadata tree as a JSON string.
+        /// </para>
+        /// </summary>
+        [Parameter]
+        public SwitchParameter Recursive { get; set; }
```
chrsmith: I'm not saying I like this, but `-Recurse` is what `Get-ChildItem` uses.
https://technet.microsoft.com/en-us/library/hh849800.aspx

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,173 @@
...
+        /// If true, the query will wait for the metadata to update.
+        /// </para>
+        /// </summary>
+        [Parameter]
+        public SwitchParameter WaitUpdate { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The last ETag known. Used in conjunction with -WaitUpdate. If the last ETag does not match the
+        /// current ETag of the metadata server, it will return immediatly.
```
chrsmith: it will return immediately... with what? An error? The stale values? The current values?

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,173 @@
...
+        /// </para>
+        /// </summary>
+        [Parameter]
+        public string LastETag { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// Used in conjunction with -WaitUpdate. The amout of time to wait. If the timeout expires, the 
+        /// current metadata will be returned.
```
chrsmith: Maybe add another "description" block on how to know if the timeout was hit:

You can determine if an update was received or not by checking the ETag value. See the `-AppendETag` parameter.

Google.PowerShell/Compute/GceMetadataCmdlet.cs
```diff
@@ -0,0 +1,173 @@
...
+        /// </para>
+        /// </summary>
+        [Parameter]
+        public TimeSpan? Timeout { get; set; }
+
+        private HttpWebRequest _request;
```
chrsmith: Oh hello there, why are you required? What do you do gentle HTTPWebRequest object?

chrsmith:
A few minor nits, e.g. renaming the parameter `-Recurse` to be consistent. But I think this looks good. I look forward to using this cmdlet one day.

chrsmith:
LGTM

PR #216, opened 2016-08-09
Remove attached disk config

ILMTitan:
Several of our usability testers, when asked to attach a disk to an instance, tried to use New-GceAttachedDiskConfig. Only one quickly realize it did not do what they wanted. The cmdlet does not have a strong use case (it surfaces a few more options than simply adding and removing disks) and can be replaced with the more verbose New-Object Google.Apis.Compute.v1.AttachedDisk.


chrsmith:
I don't think removing this cmdlet is the right call. We should put a stake in the ground and say if users are ever expected to call `New-Object` on some Apiary type, then we have failed at providing an idiomatic PowerShell experience.

 > ... when asked to attach a disk to an instance, tried to use `New-GceAttachedDiskConfig`

How exactly did users in the study fail to complete the scenario?

Were they trying to attach the result of a `New-GceAttachedDiskConfig` to the object? Perhaps we need a better error message? Perhaps we should allow that cmdlet to accept the output of `GceAttachedDiskConfig`?

Let's try to better understand the problem before we shave off the feature.

ILMTitan:
Ok, I will pull this back.

YnR:
I would start by clarifying the scenarios AttachedDiskConfig provides over Set-GCEInstance -AddDisk (note that we have Add and Attach as two different verbs here). As Chris experienced earlier this week, having two ways of doing the same thing is bad usability. It's better to have a way that is easy to use for 80% of use cases, and which can be harder to use for the remaining 20%. 

Another mitigation options short of removing the cmdlet is to be explicit in the reference: mention the Set-GCEInstance -AddDisk command as an alternative in AttachedDiskConfig.

ILMTitan:
New-AttachedDiskConfig is not an alternative to Set-GceInstance -AddDisk, it is a supplement. Set-GceInstance -AddDisk can take the name of a disk, a disk object, or an AttachedDisk object. You use an AttachedDisk object when you want to specify additional parameters, such as whether it auto deletes when the instance is deleted, the name of the disk as seen by the instance it is attaching to, and the type of interface it uses to attach.

PR #251, opened 2016-08-17
Json website

marceloyuela:
I admit this is a (massive) pull request. My apologies.
This pull request encompasses JSON functionality in the website. I ended up not splitting it into smaller portions because I adjusted the website alongside the generation of the cmdlet JSON. 

Some things to note about what isn't in this pull request:
Correct style for font sizes, font family, etc.
Good naming of files.
The standard footer for a google page.

Beyond these issues, the website works as intended for cmdlet JSON generation and display.


Fix-Website/Stylesheets/toc.css
```diff
@@ -1,17 +1,25 @@
...
+    overflow-y: auto;
+    position: fixed;
+    width: 400px;
 }
 .Info {
+    margin-left: 420px;
     position: static;
-    margin-left: 20px;
-    float:left;
+    width: 1200px;
```
chrsmith: We should generally never specify width for items in our CSS. As that is much better handled by the CSS framework we are using.

The problem with specifying width here (e.g. `margin-left`) is that other things are dependent upon the exact width. Moreover, the website doesn't react well when the browser window is resized.

We should keep this as-is for now, since we know we will be beating the website with a CSS stick later. But for reference, read this page: http://getbootstrap.com/css/. Bootstrapp is a proto-typical CSS framework that abstracts out the need for you to specify exact widths of things.

Fix-Website/app/app.js
```diff
@@ -4,28 +4,26 @@
     app.directive("contentTable", function(){
         return {
             restrict: 'E',
-            templateUrl: "Templates/content-table.html",
+            templateUrl: "Templates/content-table.txt",
             controller: function($scope) {
```
chrsmith: We should definitely define the controllers elsewhere, and not included them here. The Google style guide for JavaScript breaks lines at 80-chars, and already 20+ is wasted on just indentation for the app.directive's parameter, which is a lambda, which returns a JavaScript object literall, which has a field named controller, which has value of a lambda expression, which inside of the anonymous function's body has a method named clickProduct, which has an if statement, etc.

Fix-Website/app/app.js
```diff
@@ -34,25 +32,44 @@
...
-            templateUrl: "Templates/info.html",
-            controller: function () {
-                this.form = false;
-                this.clickProduct = function () {
-
-                    this.form = !this.form;
-                };
-            },
-            controllerAs: 'info'
+            templateUrl: "Templates/info.txt"
```
chrsmith: It is fine for now, but one of the things we'll want to do once the basics of the website are in place are normalize file names.

We should standardize on the files and folders related to this website being all lower-case. e.g. `templates/info.txt`. When you deploy this website to a Linux webserver, random things will randomly stop working because of a case-sensitive file system.

Fix-Website/app/app.js
```diff
@@ -34,25 +32,44 @@
...
+        $scope.active = "basic";
+        $scope.activeProduct = "";
         $scope.checkFrame = function (check) {
-            return this.frame === check;
+            return $scope.frame === check;
         };
-        $scope.setFrame = function (newFrame) {
-            this.frame = newFrame;
+        $scope.setFrame = function (newFrame, newActive) {
+            if (newFrame === 3 & $scope.frame != 3) {
```
chrsmith: This should probably be `&&` instead of `&`, right?

Fix-Website/app/app.js
```diff
@@ -34,25 +32,44 @@
...
         $scope.frame = 1;
+        $scope.active = "basic";
+        $scope.activeProduct = "";
         $scope.checkFrame = function (check) {
-            return this.frame === check;
+            return $scope.frame === check;
         };
-        $scope.setFrame = function (newFrame) {
-            this.frame = newFrame;
+        $scope.setFrame = function (newFrame, newActive) {
```
chrsmith: What does this function do? Smells like it needs some comments.

Fix-Website/app/app.js
```diff
@@ -34,25 +32,44 @@
...
+            }
+            $scope.frame = newFrame;
+            $scope.active = newActive;
         };
+        $http.get('cmdletsfull.json')
+            .then(function (res) {
+                $scope.products = res.data;
+            });
     }]);
+    /*Angular doesn't sort objects, only arrays, so I had to make my own filter.*/
```
chrsmith: This comment isn't useful. You explained why you are creating a filter, but don't explain what it actually _does_. How does it work? What does it actually do? What is `cmdlets` supposed to be? What is `cmdletField`?

Tools/GenerateFullWebsiteData.ps1
```diff
@@ -0,0 +1,82 @@
+# Generates a JSON file containing the documentation information for all
```
chrsmith: This file is just a copy of `GenerateWebsiteData.ps1` right? Could we just have that script generate two sets of files? One for the old website and another set of JSON files for your new one? Otherwise we'd need to make updates in two places to this script.

Fix-Website/app/app.js
```diff
@@ -34,25 +32,44 @@
...
         };
     });
 
-    app.controller('ReferenceController', ['$scope', function ($scope) {
+    /**
+     * Controller for setting up the scope.
+     */
+    app.controller('ReferenceController', ['$scope','$http', function ($scope, $http) {
         $scope.frame = 1;
+        $scope.active = "basic";
```
chrsmith: What does this mean? Pretty much every single exported property should have a comment explaining what it is used for.

Fix-Website/app/app.js
```diff
@@ -4,28 +4,26 @@
...
-                        }
+                        $scope.setFrame(2,name);
+                        this.activeProduct = name;
                     }
                 };
-                this.clickCmdlet = function () {
-                    this.cmdlet = true;
-                    $scope.setFrame(3);
+                this.clickCmdlet = function (name) {
+                    $scope.setFrame(3,name);
```
chrsmith: Space after `,`.

chrsmith:
Looks like it is on the right track.

Please do another pass and comment the code liberally. And teasing out the controllers into separate declarations will go a long way towards aiding readability.

Also, see if you can run the code through `gjslint` ([link](https://developers.google.com/closure/utilities/docs/linter_howto)). If you can't get the JS linter working on Windows, let me know, and I can help you out with that.

Fix-Website/Stylesheets/toc.css
```diff
@@ -1,17 +1,25 @@
...
+    word-wrap: break-word;
 }
 .Link {
-    color: #4285F6;
+    color: #4285f6;
```
iantalarico: Bonus point if you comment a general color description.

Fix-Website/app/app.js
```diff
@@ -4,28 +4,26 @@
     app.directive("contentTable", function(){
         return {
             restrict: 'E',
-            templateUrl: "Templates/content-table.html",
+            templateUrl: "Templates/content-table.txt",
```
iantalarico: angular templates should end with .ng

Fix-Website/app/app.js
iantalarico: +1 to this.  You are defining functions here which is hard to follow and read.

Fix-Website/app/app.js
```diff
@@ -4,28 +4,26 @@
...
                 this.expanded = false;
-                this.cmdlet = false;
-                this.clickProduct = function () {
-                    if (this.cmdlet) {
-                        $scope.setFrame(2);
-                        this.cmdlet = false;
+                this.activeProduct = "";
+                this.clickProduct = function (name) {
+                    if (this.activeProduct === name) {
+                        $scope.setFrame(1,"basic");
```
iantalarico: nit: space after comma

iantalarico: What is basic? Is this a constant value?

Fix-Website/app/app.js
```diff
@@ -34,25 +32,44 @@
...
-            controllerAs: 'info'
+            templateUrl: "Templates/info.txt"
         };
     });
 
-    app.controller('ReferenceController', ['$scope', function ($scope) {
+    /**
+     * Controller for setting up the scope.
+     */
+    app.controller('ReferenceController', ['$scope','$http', function ($scope, $http) {
```
iantalarico: You should be able to drop "['$scope','$http',"

Fix-Website/app/app.js
```diff
@@ -34,25 +32,44 @@
...
         $scope.frame = 1;
```
iantalarico: You should avoid defining things on the scope.  This clutters the scope object.  You should either make services or define things on controllers.

Fix-Website/app/app.js
```diff
@@ -34,25 +32,44 @@
...
-        $scope.setFrame = function (newFrame) {
-            this.frame = newFrame;
+        $scope.setFrame = function (newFrame, newActive) {
+            if (newFrame === 3 & $scope.frame != 3) {
+                $scope.activeProduct = $scope.active;
+            }
+            $scope.frame = newFrame;
+            $scope.active = newActive;
         };
+        $http.get('cmdletsfull.json')
```
iantalarico: What happens if this http request fails?

Should you show a loading indicator?

Fix-Website/app/app.js
marceloyuela: From what I can tell, doing so breaks the site.

chrsmith: IIRC, you need to declare the values up front, since otherwise Angular won't know what to inject, right? Since this isn't going through the Closure compiler, there are some things we don't get for free. The dependency injection bit is part of that. Maybe?

iantalarico: I think you should be able to use implicit annotations if you aren't using a minifier.

Fix-Website/app/app.js
marceloyuela: Wrote down to add this in a future PR.

marceloyuela:
Aside for .ng (which I'm waiting for help on from @iantalarico (?)/looking at now), and some other shenanigans, most of the above comments have been resolved.

Also running gjslint now.

Tools/GenerateWebsiteData.ps1
```diff
@@ -16,8 +16,12 @@ $apiMappings = @{
 
 # Generate a single JSON file containing all the documentation for all the
 # cmdlets. Unfortunately we can't split these into multiple files because of
-# the way we generating web pages in Jekyll.
+# the way we generating web pages in Jekyll/angular..
```
chrsmith: With this update, you can delete the other WebsiteData script.

Tools/GenerateWebsiteData.ps1
```diff
@@ -16,8 +16,12 @@ $apiMappings = @{
...
 $cmdletDocObjects = @()
+
+#Allows us to separate out cmdlets by product.
+$cmdletDocObjectsGcs = @{}
```
chrsmith: Can't we just use a higher-level hash? e.g.

    @{ "Google Cloud Storage" = .... ; "Google Compute Engine" = .... }

Otherwise this will get really ugly when we have cmdlets for a dozen or more products.

Tools/GenerateWebsiteData.ps1
```diff
@@ -62,11 +67,43 @@ foreach ($cmdlet in $cmdlets) {
         "documentation" = $docText
     }
 
+    # Contains more information than above.
+    $cmdletDocObjFull = @{
+        "name" = $name
+        "synopsis" = ($docObj.Synopsis | Out-String).Trim()
```
chrsmith: Since the `  | Out-String).Trim()` pattern is used so often, could you just create another method? Such as `convertToString()`?

Tools/GenerateWebsiteData.ps1
marceloyuela: The higher level hash is used later- see line 93. These are just used to store the cmdlets in their relevant area while it loops through all cmdlets.

Tools/GenerateWebsiteData.ps1
marceloyuela: Done

marceloyuela:
Fixed most of the above issues. Note: .ng still not working out. Bootstrapping later, along with anchor links/ other polish problems.

chrsmith:
There are a bunch of minor stylistic issues we can address in a future PR. This looks good enough to keep on getting the bones/structure in place.

PR #257, opened 2016-08-23
Ran the code formatter.

marceloyuela:
Self explanatory commit.
It changed some cmdlets beyond my own, so look through it anyway.


chrsmith:
LGTM. I don't like every change the tool made, but c'est la vie.

PR #276, opened 2016-08-25
Drop code-font on "Cloud Tools for PowerShell"

dupuy:
Especially with three full-fixed-width spaces in the name, putting it in `<code>` looks totally awkward. In any case, `<code>` should be reserved for things you will actually type at the command line, which you never do with "Cloud Tools for PowerShell", unlike `gcloud` or even `Add-GcdManagedZone`.

Also rewords text with an incorrect use of "latter" in the last paragraph. It seems "later" was meant, but even then it was in need of  more extensive restructuring for English grammar and clarity.

### DIFF
--- a/website/content/home.html
+++ b/website/content/home.html
@@ -59,13 +59,13 @@
     <div class="quote-box">
       <h3 class="block-title">What is it?</h3>
 
-      <p><code>Cloud Tools for PowerShell</code> is a set of PowerShell
+      <p><strong>Cloud Tools for PowerShell</strong> is a set of PowerShell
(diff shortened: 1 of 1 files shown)


dupuy:
I signed it. I am a Google employee (alexdupuy@google.com) but have an existing GitHub account linked to my Gmail and other personal e-mail accounts.

chrsmith:
LGTM, thank you for the PR! We are currently working with a UX designer on another rev of our cmdlet reference website, which can certainly use some more improvements like the one you just made.

Issue #284, opened 2016-08-26
Operations sometimes throw invalid argument exception.


ILMTitan:
Apparently a ProgressRecord can not take a negative activityId. Simply using the hashcode of an operation is therefore not viable.

PR #285, opened 2016-08-26
Issue 242: Check that -Project was not specified, not that is null.

ILMTitan:
Check that a parameter with a config default was not specified. Previously, we had check that the parameter was not null.

Also fix a problem with writing the progress of random operations.


ILMTitan:
Fixes issues #242  and #284.

Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -132,7 +132,7 @@ private void WriteWarnings(Operation op)
         private void WriteProgress(Operation op, string progressMessage)
         {
 
-            int activityId = op.Id.GetHashCode();
+            int activityId = Math.Abs(op.Id.GetHashCode());
```
iantalarico: This seems odd to me.  Is this common practice?  If not a comment would be nice (and below)

iantalarico:
LGTM

Google.PowerShell/Compute/GceCmdlet.cs
```diff
@@ -131,8 +131,8 @@ private void WriteWarnings(Operation op)
         ///  method will generate a message from the operation.</param>
         private void WriteProgress(Operation op, string progressMessage)
         {
-
-            int activityId = op.Id.GetHashCode();
+            // ProgressRecord was throwing an error on negative activityIds.
```
chrsmith: "was throwing" => "will throw"

Issue #220, opened 2016-08-10
Order the syntax summary putting the most likely scenario first.  [userstudy]


YnR:
Like `-AddDisk` for `set-GCEInstance` or `New-GCSObject -File` vs `-Content`

ILMTitan:
I will have to examine how that gets ordered in the first place.

chrsmith:
It's possible to reorder the cmdlet parameters, which would then influence the way the "syntax" view gets displayed. However that doesn't meet the bar for a reasonable code change. (Lots of churn, negligible benefit.) Also, while we can gather data about which parametersets are the most common, it doesn't seem like a good use of time (adjusting the parameters in code so that everything lines up).

With the new website we'll try to make the syntax view easier to understand (perhaps with colors? or some other differentiation). And also, place a larger emphasis on examples would have a more dramatic impact on usability.

PR #295, opened 2016-08-29
Issue #220: Change the order of syntax elements.

ILMTitan:
Change the order of Get-GceInstance, Set-GceInstance and New-GcsObject to put the more common ones on top, and more interesting parameters earlier.


chrsmith:
Given that this is a breaking change, we need to look carefully about whether or not this is the right thing to do. What is the impetus to "put the more common ones on top, and more interesting parameters earlier"?

In some cases, I disagree with the newer order. As to me it seems that `Project` should always go first, as it is the most relevant value, opposed to `Name` which may be redundant across projects.

I could be convinced of refactoring of the ordering of our cmdlets parameters, but I'd like to better understand (1) why we are doing it now and (2) what is the standard we will apply now and in the future?

ILMTitan:
This doesn't actually break anything on the powershell side. Positional parameter's order is determined by the position argument to the Parameter attribute. All other parameters are named an can be in any order.

What is the difference between these parameter sets:

    SYNTAX
    Set-GceInstance [-Project <string>] [-Zone <string>] [-Name] <string> -NetworkInterface <string> [-AddAccessConfig
    <AccessConfig[]>] [-RemoveAccessConfig <string[]>] [<CommonParameters>]

    Set-GceInstance [-Project <string>] [-Zone <string>] [-Name] <string> [-AddDisk <object[]>] [-RemoveDisk
    <string[]>] [<CommonParameters>]

    Set-GceInstance [-Project <string>] [-Zone <string>] [-Name] <string> [-AddMetadata <Hashtable>] [-RemoveMetadata
    <string[]>] [<CommonParameters>]

    Set-GceInstance [-Project <string>] [-Zone <string>] [-Name] <string> [-AddTag <string[]>] [-RemoveTag <string[]>]
    [<CommonParameters>]

    Set-GceInstance [-Object] <Instance> -NetworkInterface <string> [-AddAccessConfig <AccessConfig[]>]
    [-RemoveAccessConfig <string[]>] [<CommonParameters>]

    Set-GceInstance [-Object] <Instance> [-AddDisk <object[]>] [-RemoveDisk <string[]>] [<CommonParameters>]

    Set-GceInstance [-Object] <Instance> [-AddMetadata <Hashtable>] [-RemoveMetadata <string[]>] [<CommonParameters>]

    Set-GceInstance [-Object] <Instance> [-AddTag <string[]>] [-RemoveTag <string[]>] [<CommonParameters>]


How about these?

    SYNTAX
    Set-GceInstance [-AddDisk <object[]>] [-RemoveDisk <string[]>] [-Project <string>] [-Zone <string>] [-Name]
    <string> [<CommonParameters>]

    Set-GceInstance [-AddDisk <object[]>] [-RemoveDisk <string[]>] [-Object] <Instance> [<CommonParameters>]

    Set-GceInstance [-AddMetadata <Hashtable>] [-RemoveMetadata <string[]>] [-Project <string>] [-Zone <string>]
    [-Name] <string> [<CommonParameters>]

    Set-GceInstance [-AddMetadata <Hashtable>] [-RemoveMetadata <string[]>] [-Object] <Instance> [<CommonParameters>]

    Set-GceInstance [-AddTag <string[]>] [-RemoveTag <string[]>] [-Project <string>] [-Zone <string>] [-Name] <string>
    [<CommonParameters>]

    Set-GceInstance [-AddTag <string[]>] [-RemoveTag <string[]>] [-Object] <Instance> [<CommonParameters>]

    Set-GceInstance [-AddAccessConfig <AccessConfig[]>] [-RemoveAccessConfig <string[]>] [-Project <string>] [-Zone
    <string>] [-Name] <string> -NetworkInterface <string> [<CommonParameters>]

    Set-GceInstance [-AddAccessConfig <AccessConfig[]>] [-RemoveAccessConfig <string[]>] [-Object] <Instance>
    -NetworkInterface <string> [<CommonParameters>]


I think the second is much easier to answer.

chrsmith:
Discussed in person. We will punt on this for now.

Issue #218, opened 2016-08-10
Mention related cmdlets in the reference  [userstudy]


YnR:
For instance, New-GCSObject in the Write-GCSObject reference.

PR #296, opened 2016-08-29
Issue #218: Add comments to New-GcsObject and Write-GcsObject.

ILMTitan:
Within the description of New-GcsObject and Write-GcsObject, make a note about the other and when it should be used.


chrsmith:
TYVM for the change.

LGTM

Issue #299, opened 2016-08-29
Add windows-sql-cloud to list of default image projects.


ILMTitan:
When we added sql server images to GCP, they were added as part of a new project, windows-sql-cloud. That project should be included in the list of default projects we search to get an image.

PR #300, opened 2016-08-29
Issue #299: Add windows-sql-cloud as a default image project.

ILMTitan:
When sql server images were added to GCP, they were added as part of a new project, windows-sql-cloud. That project should be included in the list of default projects we search to get an image.


chrsmith:
LGTM.

The string change is obviously OK. As for renaming it to `DefaultProjects`, I personally like the change. But the CodeFormatter tool will rename it back to `s_defaultProjects`. I believe dropping static and making it const will make DefaultProjects the name name wrt .NET conventions.

ILMTitan:
Close Issue #299

Issue #91, opened 2016-07-12
Need Copy-GcsObject.


SurferJeffAtGoogle:
Without Copy-GcsObject, I have to download and upload a file to copy it from one place to another, which is a waste of everyone's bandwidth.

Issue #126, opened 2016-07-14
Support Copy-GCSObject from one bucket to another


YnR:
Currently trying:
Find-GcsObject -Bucket ynrpizza | Read-GcsObject -Bucket ynrpizza | New-GcsObject -bucket usertestbucket

Might be good to support copy so that the data does not need to transit through my machine.

ILMTitan:
Duplicate of #91.

PR #301, opened 2016-08-30
Issue #126 Copy-GcsObject

ILMTitan:
Added Copy-GcsObject.
#126


Google.PowerShell/Common/GcloudCmdlet.cs
```diff
@@ -43,7 +43,7 @@ public GCloudCmdlet()
         /// <summary>
         /// Returns an instance of the Google Client API initializer, using the machine's default credentials.
         /// </summary>
-        protected BaseClientService.Initializer GetBaseClientServiceInitializer()
+        protected static BaseClientService.Initializer GetBaseClientServiceInitializer()
```
chrsmith: Can you back this change out? It seems unrelated to creating a new `Copy-GcsObject` cmdlet, and we should talk about the design implications here.

Google.PowerShell/Storage/GcsCmdlet.cs
```diff
@@ -24,12 +24,10 @@ public abstract class GcsCmdlet : GCloudCmdlet
...
         protected const string UTF8TextMimeType = "text/plain; charset=utf-8";
 
-        // TODO(chrsmith): Cache the storage service? Create it in OnProcessRecord every time? (So it does so once?)
-
-        protected StorageService GetStorageService()
-        {
-            return new StorageService(GetBaseClientServiceInitializer());
-        }
+        /// <summary>
+        /// The storage service.
```
chrsmith: I'm not a huge fan of this, but don't have a good reason to oppose the change.

The initial design was so that we would explicitly control the lifetime of the service object. That would make it easier to mock, test, instrument, etc.

However, we could still probably do all of those same things by going through a property accessor. And this approach also cleans up a lot of the code.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -922,4 +917,75 @@ protected override void ProcessRecord()
             }
         }
     }
+
+    [Cmdlet(VerbsCommon.Copy, "GcsObject", DefaultParameterSetName = ParameterSetNames.ByObject)]
+    [OutputType(typeof(Object))]
+    public class CopyGcsObject : GcsCmdlet
```
chrsmith: Know what I could go for? Some XML doc comments. A delicious "summary" and "description" tag would be great right about now.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -922,4 +917,75 @@ protected override void ProcessRecord()
...
+    {
+
+        private class ParameterSetNames
+        {
+            public const string ByName = "ByName";
+            public const string ByObject = "ByObject";
+        }
+
+        [Parameter(ParameterSetName = ParameterSetNames.ByObject, Mandatory = true, ValueFromPipeline = true)]
+        public Object InputObject { get; set; }
```
chrsmith: What does this parameter do?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -922,4 +917,75 @@ protected override void ProcessRecord()
...
+        /// Name of the bucket containing the object. Will also accept a Bucket object.
+        /// </para>
+        /// </summary>
+        [Parameter(ParameterSetName = ParameterSetNames.ByName, Mandatory = true)]
+        [PropertyByTypeTransformation(Property = nameof(Bucket.Name), TypeToTransform = typeof(Bucket))]
+        public string SourceBucket { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// Name of the object to write to.
```
chrsmith: Name of the object to write to? On `SourceObjectName`? Do you mean to write "Name of the object to read from?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -922,4 +917,75 @@ protected override void ProcessRecord()
...
+        /// Name of the bucket in which the copy will reside.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, Position = 0)]
+        [PropertyByTypeTransformation(Property = nameof(Bucket.Name), TypeToTransform = typeof(Bucket))]
+        public string DestinationBucket { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The name of the copy.
```
chrsmith: Perhaps "Name of the newly created Object" would be clearer?

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
```diff
@@ -600,4 +600,41 @@ Describe "Test-GcsObject" {
...
+            Copy-GcsObject -SourceBucket $bucket -SourceObject "test-source" $bucket "test-dest"
+            Read-GcsObject $bucket "test-dest" | Should Be "test1-$r"
+        }
+
+        It "Should work by object" {
+            $sourceObj = New-GcsObject $bucket "test-source2" -Contents "test2-$r"
+            $sourceObj | Copy-GcsObject $bucket "test-dest2"
+            Read-GcsObject $bucket "test-dest2" | Should Be "test2-$r"
+        }
+
```
chrsmith: What if you don't have permissions to read from the source bucket? Or write to the destination bucket? I imagine there is a test we need here.

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
```diff
@@ -600,4 +600,41 @@ Describe "Test-GcsObject" {
...
+        AfterAll {
```
chrsmith: What if a GCS object already exists in the destination? I assume we need an `-Overwrite` or `-Force` flag?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -922,4 +917,75 @@ protected override void ProcessRecord()
...
+                    break;
+                case ParameterSetNames.ByObject:
+                    gcsObject = InputObject;
+                    break;
+                default:
+                    throw UnknownParameterSetException;
+            }
+            ObjectsResource.CopyRequest request = Service.Objects.Copy(gcsObject,
+                gcsObject.Bucket, gcsObject.Name,
+                DestinationBucket ?? gcsObject.Bucket, DestinationObjectName ?? gcsObject.Name);
```
chrsmith: Having defaults for these values seems reasonable. But you should mention that in the doc comments above.

chrsmith:
The code generally looks good. A few minor issues.

ILMTitan:
PTAL

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -939,14 +939,15 @@ private class ParameterSetNames
         }
 
         /// <summary>
-        /// A Google Cloud Storage object description 
+        /// A Google Cloud Storage object description of the object to read from. They can be retrieved using
```
chrsmith: Needs `<para type="description">`.

chrsmith: "object description of the object to read from" sounds confusing. How about just

"Cloud Storage object to read from. Can be obtained using Get-GcsObject or Find-GcsObject."

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -993,9 +1000,30 @@ protected override void ProcessRecord()
...
+
+            if (!Force)
+            {
+                try
+                {
+                    ObjectsResource.GetRequest objGetReq =
+                        Service.Objects.Get(destinationBucket, destinationObject);
+                    objGetReq.Execute();
+                    // If destination does not exist, jump to catch statment.
+                    if (!ShouldContinue("Object exists. Overwrite?",
```
chrsmith: If this is too long for a single line, line break at the first parameter of `ShouldContinue`, otherwise this looks kinda odd IMHO.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -993,9 +1000,30 @@ protected override void ProcessRecord()
...
+            string destinationObject = DestinationObjectName ?? gcsObject.Name;
+
+            if (!Force)
+            {
+                try
+                {
+                    ObjectsResource.GetRequest objGetReq =
+                        Service.Objects.Get(destinationBucket, destinationObject);
+                    objGetReq.Execute();
+                    // If destination does not exist, jump to catch statment.
```
chrsmith: "jump to catch statement" sounds like you are using a goto. Perhaps:

"If destination does not exist, will throw a 404 an be ignored"

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
```diff
@@ -618,6 +618,17 @@ Describe "Copy-GcsObject" {
...
+        It "Should fail to read from unaccessable source bucket" {
+            { Copy-GcsObject -SourceBucket "asdf" -SourceObject "test-source" $bucket "test-dest" } |
+                Should Throw 403
+        }
+        
+        It "Should fail to write to unaccessable source bucket" {
+            New-GcsObject $bucket "test-source0" -Contents "test0-$r"
+            { Copy-GcsObject -SourceBucket $bucket -SourceObject "test-source0" "asdf" "test-dest" } |
+                Should Throw 403
+        }
```
chrsmith: Need a test to validate `-Force works as expected`. just copy your "should work by name" test, and add another scenario where it fails (already exists) and then works via `-Force`.

chrsmith:
LGTM, but please add the missing test cases and XML doc comment. Feel free to pass on suggested wordings for doc strings.

Google.PowerShell/Storage/GcsObject.cs
ILMTitan: It sounds like I am using a goto because I basically am using a goto.

ILMTitan:
Close #91

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
```diff
@@ -642,6 +642,12 @@ Describe "Copy-GcsObject" {
             Read-GcsObject $bucket "test-dest2" | Should Be "test2-$r"
         }
 
+        It "Should overwrite existing object" {
+            $sourceObj = Get-GcsObject $bucket "test-source"
+            $sourceObj | Copy-GcsObject $bucket "test-dest2" -Force
```
chrsmith: Sorry to be a pain here, but could you confirm an error if you don't specify `-Force`? Otherwise it could be that we always overwrite, regardless of if you set the flag.

````
{ $sourceObj | Copy-GcsObject $bucket "should-fail" } |
    Should Throw 409  # Conflit
````

chrsmith:
LGTM

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
ILMTitan: I can not. I don't throw an error. Instead I request confirmation from the user. However, the pester story for testing "ShouldConfirm" does not yet exist.

chrsmith: Gotcha. All is well then. :+1:

PR #303, opened 2016-08-30
Fixed a bug in the related links.

marceloyuela:
For some reason all links in the cmdlets start with ( and end with ), throwing off the linking of the website.
This PR fixes that.

### DIFF
--- a/Tools/GenerateWebsiteData.ps1
+++ b/Tools/GenerateWebsiteData.ps1
@@ -56,7 +56,7 @@ function getLinks ($docObj) {
     $relatedLinks = $docObj.relatedLinks
     $links = @{}
     ForEach ($link in $relatedLinks.navigationLink) {
-        $uri = $link.uri.replace("(","").replace(")","")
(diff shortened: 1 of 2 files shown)


Tools/GenerateWebsiteData.ps1
```diff
@@ -56,7 +56,8 @@ function getLinks ($docObj) {
     $relatedLinks = $docObj.relatedLinks
     $links = @{}
     ForEach ($link in $relatedLinks.navigationLink) {
-        $links.Add($link.linkText,$link.uri)
+        $uri = $link.uri.replace("(","").replace(")","")
```
ILMTitan: Should this be something like trimStart and trimEnd?

ILMTitan:
I'm sorry I forgot to mention that. We do that because it is suggested by the xmldoc2cmdletdoc tool we use. I am a little concerned that this could replace an intentional parenthesis. Does javascript have a TrimStart and TrimEnd?

marceloyuela:
(it was powershell so yes)

ILMTitan:
LGTM

Issue #260, opened 2016-08-23
10GB for the NewGceDisk -SizeGB parameter


chrsmith:
"10GB" gets converted to 10E9, but the parameter expects the value in GB, not bytes. So we could detect that and downgrade it from bytes to GB. (e.g. divide by 1,000,000,000.)

PR #307, opened 2016-09-01
Add a net for people who accidently put GB behind the size of their disks.

ILMTitan:
This is a fix for issue #260. This is a fix for behavior we don't want to encourage, so I am intentionally leaving it undocumented to external users.


chrsmith:
LGTM.

If you want to put the clamping into a method that would probably be better, but it is fine either way.

"This is a fix for behavior we don't want to encourage, so I am intentionally leaving it undocumented to external users." 👍

PR #310, opened 2016-09-08
Stackdriver product rename

tmatsuo:
Renaming some APIs to Stackdriver brand


chrsmith:
Could you send me an email to talk more about this? I'd like to better understand the context for this (and when is the right time to do the rename, etc.)

I'm happy to take care of the renaming, as changing the .json file isn't sufficent for updating our docs.

PR #313, opened 2016-09-12
Fix a compile error that snuck into master.

ILMTitan:
Apparently I didn't merge the provider with master properly, as a compile error occured after the merge. Here is a fix for that.


chrsmith:
LGTM

Issue #314, opened 2016-09-13
PSModulePath is not set up properly


KirkMunro:
Today the GoogleCloudPowerShell module gets installed into "${env:ProgramFilesx86}\Google\Cloud SDK\google-cloud-sdk\platform", and the installer adds the path to the GoogleCloudPowerShell folder to $env:PSModulePath.

This setup prevents the module from being loaded using Import-Module GoogleCloudPowerShell. You can load it using the full path to the module folder or using the full path to the psd1 file it contains, but you can't load it with Import-Module. This setup also prevents Get-Command -Module GoogleCloudPowerShell from working. Generally, anything that is dependent on lookup by module name (dependencies taken by other modules, for example) will not work.

To fix this, $env:PSModulePath should contain the folder _above_ the GoogleCloudPowerShell folder rather than the GoogleCloudPowerShell folder itself. Personally I recommend creating a PowerShell folder at the same level as GoogleCloudPowerShell, moving the GoogleCloudPowerShell folder into the PowerShell folder, and setting up the PowerShell folder you create as the item that you add to $env:PSModulePath.  That is the best solution to this issue, keeping PSModulePath folders clean and PowerShell focused so that you don't impact PowerShell performance, and setting your module up to be discoverable by name properly.

ILMTitan:
Thank you for catching this.

chrsmith:
See #315 when we get to fixing this.

chrsmith:
@KirkMunro thanks for reporting this and other PowerShell issues. This has been fixed in the repo. It will be live in the next release of the Cloud SDK, in the next week or two. (The next time you `gcloud components update`.)

Issue #315, opened 2016-09-13
The "PowerShell" suffix in your "GoogleCloudPowerShell" module name is 100% redundant


KirkMunro:
You don't see people tacking "Python" and "Ruby" on the end of all of their package/gem names, do you?

Since this is beta (I think it's just beta), please consider dropping the "PowerShell" suffix from the folder name and the psd1 file, and updating references accordingly. You'll save scripters who use your platform a lot of unnecessary typing in the long run.

chrsmith:
Good point. There also was a name change half way through our development, from "Google Cloud PowerShell" to "Google Cloud Tools for PowerShell". So there are a few places here and there that still reflect the old name.

chrsmith:
@KirkMunro thanks for reporting this and other PowerShell issues. This has been fixed in the repo. It will be live in the next release of the Cloud SDK, in the next week or two.

PR #319, opened 2016-09-14
Changes to fix Issues #314 and #315.

ILMTitan:
Internal release process files will also have to be changed.


chrsmith:
I'll hold off on reviewing these until next week. I'd like to submit the changes to the Cloud SDK installer the same day, so we don't wind up accidentally shipping these changes without the installer being updated to match.

chrsmith:
I'll merge the change locally and resubmit a PR. I'd like to have the PR and the internal Cloud SDK changes made at the same time.

Issue #318, opened 2016-09-14
Remove unapproved verbs from module


KirkMunro:
Today you have the following commands that use unapproved verbs:

Failover-GcSqlInstance
Promote-GcSqlReplica

Use of unapproved verbs in exported module command names is strongly discouraged (to the point where many teams are simply told not to do it under any circumstances).

Instead of Failover-GcSqlInstance, consider:

Invoke-GcSqlInstanceFailover

Instead of Promote-GcSqlReplica, consider:

Convert-GcSqlReplicaToInstance

There are other possible verbs you can choose from of course and what I provided are just some that I feel work fine for what you need, but regardless of which you choose, the core command should have an approved verb.  If you really, _really_ feel the need for these names, define them as aliases for commands that use approved verbs.  In the long run I don't feel it's worth it (it's a best practice to avoid using aliases in scripts), and aliases like these are usually only kept around to maintain backwards compatibility with names that had to change for some reason.

ILMTitan:
Thank you. Those look like good alternative names.

chrsmith:
Notes on investigation so far:

Replacing `Failover-` with `Invoke-GcSqlInstanceFailover` sounds good. As for `Promote-`, since the conversion isn't bidirectional, we should use `ConvertTo-GcSqlInstance` ([docs](https://msdn.microsoft.com/en-us/library/ms714428(v=vs.85).aspx)).

We'll need to make this change before exiting Beta, since at that point we won't be able to make breaking changes like this. However, even at this point we still need to go through and update documentation (e.g. on cloud.google.com) to mention the changes.

rrxtns:
@chrsmith `Convert-GcSqlInstance` would be more in line with the naming convention (though admittedly, it may be non specific enough)

KirkMunro:
@rrxtns Actually if you just use the Convert verb, it sounds like you are converting a GcSqlInstance, as opposed to converting to a GcSqlInstance (which is what I believe is actually happening).

@chrsmith I like the ConvertTo-GcSqlInstance name for that command, and good catch on that being the guidance as per the docs.

quoctruong:
This is fixed with https://github.com/GoogleCloudPlatform/google-cloud-powershell/pull/359. We changed the cmdlets to ConvertTo-GcSqlInstance and Invoke-GcSqlInstanceFailover.

Issue #316, opened 2016-09-13
Replace the *'s in your psd1 file with an array of actual export names


KirkMunro:
It's a known issue in PowerShell that using '*' in Export* fields in a manifest negatively impacts performance.

Best practices dictate actually listing all exports in a list, and if you don't like doing it manually, keep it current with a script.

ILMTitan:
Kirk, thank you for your comments.
Do you mean "using '\*' in Export\* fields"? I think the asterisks got eaten by the markdown.

vors:
Kirk is right.
For the details about associated cost and benchmarking see this PR in azure-powershell
https://github.com/Azure/azure-powershell/pull/1763#issuecomment-185009796

KirkMunro:
@ILMTitan Yeah, thanks for catching that, I was trying to log a bunch of issues before supper and didn't take the time to make sure I escaped the asterisks.

quoctruong:
This is now fixed with https://github.com/GoogleCloudPlatform/google-cloud-powershell/pull/369

PR #321, opened 2016-09-16
Replace * with an actual list of cmdlets.

ILMTitan:
See issue #316.
Also, fix 2 GcSql cmdlets to use proper verbs.
See issue #318.

### DIFF
--- a/Google.PowerShell/ReleaseFiles/GoogleCloudPowerShell.psd1
+++ b/Google.PowerShell/ReleaseFiles/GoogleCloudPowerShell.psd1
@@ -69,13 +69,36 @@ NestedModules = @('.\Google.PowerShell.dll')
 FunctionsToExport = '*'
 
 # Cmdlets to export from this module
-CmdletsToExport = '*'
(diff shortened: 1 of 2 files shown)


chrsmith:
Let us not submit this PR. Replacing * with the actual list of cmdlets sounds good, but unless we have a way to automate updating that list, what we have will trivially become out of date.

The other change, change GcSql cmdlets to use valid verbs, looks great. But we'll have to tease that out into a separate PR.

chrsmith:
`(Get-Module GoogleCloud).ExportedCmdlets.Keys` is maybe 70% of the way there. Maybe we do some jank text substitution as part of the `BuildAndPackage.ps1` script? e.g. have {{PLACEHOLDER_X}} in the file, and then find/replace that in the script?

ILMTitan:
Or we could do:

    $manifest = Import-PowerShellDataFile GoogleCloud.psd1
    $manifest.CmdletsToExport = $generatedCmdletsList
    New-ModuleManifest @manifest -Path GoogleCloud.psd1

PR #324, opened 2016-09-21
Rename the module and assumed file path.

### LINKED ISSUE #314: PSModulePath is not set up properly

chrsmith:
Changes the module name to "GoogleCloud" from "GoogleCloudPowerShell". The
trailing "PowerShell" is redundant. (Although AWS' cmdlets are called
"AWSPowerShell".)

Changes the assumed location of the cmdlets.

Changes references to the product as a whole to "Cloud Tools for PowerShell", since "Google Cloud PowerShell" was the old name while we were still in alpha.

Previously the Cloud SDK installer
would the files in "${Cloud SDK}\platform\GoogleCloudPowerShell\" and add
"${Cloud SDK}\platform\GoogleCloudPowerShell\" to PSModulePath.

While this allowed PowerShell to quickly find GoogleCloudPowerShell.psd1,
it broke the ability to type Import-Module GoogleCloudPowerShell and do other
sorts of lookups.

Internal cl/133881461 will update the Cloud SDK to use a new folder for
PSModulePath: "${Cloud SDK}\platform\PowerShell". And the PowerShell
bits will be installed into "${Cloud SDK}\platform\PowerShell\GoogleCloud".

Testing:
- Manual testing that PowerShell behaves the way Microsoft tells us it does.
- Manually built the new installer, ran it, and manually replaced the "powershell"
    component bits ($INSTALLDIR/platform/) with the new bits.
- I did not test the PatchFromGcs.ps1 or AppendModulePath.ps1 scripts.

Fixes #314 and #315.


Google.PowerShell/Google.PowerShell.csproj
```diff
@@ -195,9 +195,9 @@
       <SubType>Designer</SubType>
     </None>
     <None Include="README.md" />
-    <None Include="ReleaseFiles\BootstrapGoogleCloudPowerShell.ps1" />
+    <None Include="ReleaseFiles\BootstrapCloudToolsForPowerShell.ps1" />
     <None Include="ReleaseFiles\GoogleCloudPlatform.Format.ps1xml" />
-    <None Include="ReleaseFiles\GoogleCloudPowerShell.psd1" />
+    <None Include="ReleaseFiles\CloudToolsForPowerShell.psd1" />
```
ILMTitan: This should be GoogleCloud.psd1

chrsmith: Good catch. Evidently you need to "Save All" to save changes to the project file. (Which is something I'm sure I've been burned by in the past.)

chrsmith:
PTAL

ILMTitan:
LGTM

PR #329, opened 2016-09-29
Add ability to upload folder to New-GCSObject

quoctruong:
Add -Folder parameter to New-GCSObject cmdlet to allow uploading of folder;
Use Cmdlet.GetResolvedProviderPathFromPSPath to resolve relative paths;
Instead of using -Path to combine File and Folder, I decided to add -Folder since there are certain parameters that can be applied to -File but not -Folder (like -ObjectName and -ContentType) and so it makes more sense to separate them into 2 different parameter sets.

### DIFF
--- a/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
@@ -36,7 +36,7 @@ Describe "New-GcsObject" {
 
     It "should fail if the folder does not exist" {
         { New-GcsObject $bucket -Folder "C:\I should not exist" } `
-            | Should Throw "Directory C:\I should not exist cannot be found"
+            | Should Throw "Directory 'C:\I should not exist' cannot be found"
     }
 
     It "accepts pipeline input as object contents" {
@@ -214,62 +214,64 @@ Describe "New-GcsObject" {
     }
 
     It "should upload an empty folder" {
-        $TestFolder = Join-Path $TestDrive "TestFolder"
+        $folderName = [System.IO.Path]::GetRandomFileName()
+        $testFolder = "$TestDrive/$folderName"
 
         try {
-            if (-not (Test-Path $TestFolder))
-            {
-                New-Item -ItemType Directory -Path $TestFolder | Out-Null
-            }
+            New-Item -ItemType Directory -Path $testFolder | Out-Null
 
-            $folder = New-GcsObject -Bucket $bucket -Folder $TestFolder -Force
-            $folderName = "TestFolder/"
+            $folder = New-GcsObject -Bucket $bucket -Folder $testFolder -Force
 
-            $folder.Name | should be $folderName
+            $folder.Name | should be "$folderName/"
             $folder.Size | should be 0
 
-            $folderOnline = Get-GcsObject -Bucket $bucket -ObjectName $folderName
-            $folderOnline.Name | should be $folderName
+            $folderOnline = Get-GcsObject -Bucket $bucket -ObjectName "$folderName/"
+            $folderOnline.Name | should be "$folderName/"
             $folderOnline.Size | should be 0
         }
         finally
         {
-            Remove-Item $TestFolder -Recurse -Force -ErrorAction Ignore
+            Remove-Item $testFolder -Recurse -Force -ErrorAction Ignore
         }
     }
 
     It "should upload a folder with files and subfolders" {
-        $TestFolder = Join-Path $TestDrive "TestFolder"
+        $folderName = [System.IO.Path]::GetRandomFileName()
+        $testFolder = "$TestDrive/$folderName"
 
         try {
-            if (-not (Test-Path $TestFolder))
-            {
-                New-Item -ItemType Directory -Path $TestFolder | Out-Null
-            }
+            New-Item -ItemType Directory -Path $testFolder | Out-Null
 
-            "Hello, world" | Out-File (Join-Path $TestFolder "world.txt")
-            "Hello, mars" | Out-File (Join-Path $TestFolder "mars.txt")
-            "Hello, jupiter" | Out-File (Join-Path $TestFolder "jupiter.txt")
+            "Hello, world" | Out-File "$testFolder/world.txt"
+            "Hello, mars" | Out-File "$testFolder/mars.txt"
+            "Hello, jupiter" | Out-File "$testFolder/jupiter.txt"
 
-            $TestSubfolder = Join-Path $TestFolder "TestSubfolder"
-            New-Item -ItemType Directory -Path $TestSubfolder | Out-Null
+            $testSubfolder = "$testFolder/TestSubfolder"
+            New-Item -ItemType Directory -Path $testSubfolder | Out-Null
 
-            "Hello, saturn" | Out-File (Join-Path $TestSubfolder "saturn.txt")
-            "Hello, pluto" | Out-File (Join-Path $TestSubfolder "pluto.txt")
+            "Hello, saturn" | Out-File "$testSubfolder/saturn.txt"
+            "Hello, pluto" | Out-File "$testSubfolder/pluto.txt"
 
-            $result = New-GcsObject -Bucket $bucket -Folder $TestFolder -Force
+            $result = New-GcsObject -Bucket $bucket -Folder $testFolder -Force
 
             $result.Count | should be 7
-            $result.Name -contains "TestFolder/jupiter.txt" | should be $true
-            $result.Name -contains "TestFolder/mars.txt" | should be $true
-            $result.Name -contains "TestFolder/TestSubfolder/pluto.txt" | should be $true
+            $result.Name -contains "$folderName/jupiter.txt" | should be $true
+            $result.Name -contains "$folderName/mars.txt" | should be $true
+            $result.Name -contains "$folderName/TestSubfolder/pluto.txt" | should be $true
 
-            $saturn = Get-GcsObject -Bucket $bucket -ObjectName "TestFolder/TestSubfolder/saturn.txt"
+            $saturn = Get-GcsObject -Bucket $bucket -ObjectName "$folderName/TestSubfolder/saturn.txt"
             $saturn.ContentType | should be "text/plain"
+
+            $objs = Find-GcsObject -Delimiter "/" -Bucket $bucket -Prefix "$folderName/TestSubfolder"
+            $objs.Count | should be 0
+
+            $objs = Find-GcsObject -Bucket $bucket -Prefix "$folderName/TestSubfolder"
+            $objs.Count | should be 3
+            $objs.Name -contains "$folderName/TestSubfolder/pluto.txt" | should be $true
         }
         finally
         {
-            Remove-Item $TestFolder -Recurse -Force -ErrorAction Ignore
+            Remove-Item $testFolder -Recurse -Force -ErrorAction Ignore
         }
     }
 }
--- a/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
@@ -34,6 +34,11 @@ Describe "New-GcsObject" {
             | Should Throw "File not found"
     }
 
+    It "should fail if the folder does not exist" {
+        { New-GcsObject $bucket -Folder "C:\I should not exist" } `
+            | Should Throw "Directory C:\I should not exist cannot be found"
+    }
+
     It "accepts pipeline input as object contents" {
         "test string" | New-GcsObject $bucket "pipeline-test"
         Read-GcsObject $bucket "pipeline-test" | Should Be "test string"
@@ -206,6 +211,66 @@ Describe "New-GcsObject" {
         # It will also apply to the Metadata too.
         $obj.Metadata["Content-Type"] | Should Be "image/jpeg"
         Remove-GcsObject $obj
+    }
+
+    It "should upload an empty folder" {
+        $TestFolder = Join-Path $TestDrive "TestFolder"
+
+        try {
+            if (-not (Test-Path $TestFolder))
+            {
+                New-Item -ItemType Directory -Path $TestFolder | Out-Null
+            }
+
+            $folder = New-GcsObject -Bucket $bucket -Folder $TestFolder -Force
+            $folderName = "TestFolder/"
+
+            $folder.Name | should be $folderName
+            $folder.Size | should be 0
+
+            $folderOnline = Get-GcsObject -Bucket $bucket -ObjectName $folderName
+            $folderOnline.Name | should be $folderName
+            $folderOnline.Size | should be 0
+        }
+        finally
+        {
+            Remove-Item $TestFolder -Recurse -Force -ErrorAction Ignore
+        }
+    }
+
+    It "should upload a folder with files and subfolders" {
+        $TestFolder = Join-Path $TestDrive "TestFolder"
+
+        try {
+            if (-not (Test-Path $TestFolder))
+            {
+                New-Item -ItemType Directory -Path $TestFolder | Out-Null
+            }
+
+            "Hello, world" | Out-File (Join-Path $TestFolder "world.txt")
+            "Hello, mars" | Out-File (Join-Path $TestFolder "mars.txt")
+            "Hello, jupiter" | Out-File (Join-Path $TestFolder "jupiter.txt")
+
+            $TestSubfolder = Join-Path $TestFolder "TestSubfolder"
+            New-Item -ItemType Directory -Path $TestSubfolder | Out-Null
+
+            "Hello, saturn" | Out-File (Join-Path $TestSubfolder "saturn.txt")
+            "Hello, pluto" | Out-File (Join-Path $TestSubfolder "pluto.txt")
+
+            $result = New-GcsObject -Bucket $bucket -Folder $TestFolder -Force
+
+            $result.Count | should be 7
+            $result.Name -contains "TestFolder/jupiter.txt" | should be $true
+            $result.Name -contains "TestFolder/mars.txt" | should be $true
+            $result.Name -contains "TestFolder/TestSubfolder/pluto.txt" | should be $true
+
+            $saturn = Get-GcsObject -Bucket $bucket -ObjectName "TestFolder/TestSubfolder/saturn.txt"
+            $saturn.ContentType | should be "text/plain"
+        }
+        finally
+        {
+            Remove-Item $TestFolder -Recurse -Force -ErrorAction Ignore
+        }
     }
 }
 
--- a/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
@@ -246,7 +246,7 @@ Describe "New-GcsObject" {
         }
     }
 
-    It "should upload a folder with slashes at the end of the path" {
+    It "should work for both forwards slashes and backslashes in the path" {
         $folderName = [System.IO.Path]::GetRandomFileName()
         $testFolder = "$env:TEMP/$folderName"
 
@@ -254,7 +254,7 @@ Describe "New-GcsObject" {
             New-Item -ItemType Directory -Path $testFolder | Out-Null
             "Hello, world" | Out-File "$testFolder/world.txt"
 
-            # Add a backwards slash to the end and make sure it is uploaded.
+            # Add a backslash to the end and make sure it is uploaded.
             $result = New-GcsObject -Bucket $bucket -Folder "$testFolder\" -Force
 
             $result.Count | Should Be 2
--- a/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
@@ -6,7 +6,18 @@ $project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
 Describe "New-GcsObject" {
 
     $bucket = "gcps-object-testing"
-    Create-TestBucket $project $bucket
+
+    BeforeEach {
+        if (-not (Test-GcsBucket $bucket)) {
+            Create-TestBucket $project $bucket
+        }
+    }
+
+    AfterEach {
+        if (Test-GcsBucket $bucket) {
+            Remove-GcsBucket -Name $bucket -Force
+        }
+    }
 
     It "should work" {
         $filename = [System.IO.Path]::GetTempFileName()
@@ -215,7 +226,7 @@ Describe "New-GcsObject" {
 
     It "should upload an empty folder" {
         $folderName = [System.IO.Path]::GetRandomFileName()
-        $testFolder = "$TestDrive/$folderName"
+        $testFolder = "$env:TEMP/$folderName"
 
         try {
             New-Item -ItemType Directory -Path $testFolder | Out-Null
@@ -235,9 +246,37 @@ Describe "New-GcsObject" {
         }
     }
 
+    It "should upload a folder with slashes at the end of the path" {
+        $folderName = [System.IO.Path]::GetRandomFileName()
+        $testFolder = "$env:TEMP/$folderName"
+
+        try {
+            New-Item -ItemType Directory -Path $testFolder | Out-Null
+            "Hello, world" | Out-File "$testFolder/world.txt"
+
+            # Add a backwards slash to the end and make sure it is uploaded.
+            $result = New-GcsObject -Bucket $bucket -Folder "$testFolder\" -Force
+
+            $result.Count | Should Be 2
+            $result.Name -contains "$folderName/" | should be $true
+            $result.Name -contains "$folderName/world.txt" | should be $true
+
+            # Add a forwards slash to the end and make sure it is uploaded.
+            $result = New-GcsObject -Bucket $bucket -Folder "$testFolder/" -Force
+
+            $result.Count | Should Be 2
+            $result.Name -contains "$folderName/" | should be $true
+            $result.Name -contains "$folderName/world.txt" | should be $true
+        }
+        finally
+        {
+            Remove-Item $testFolder -Recurse -Force -ErrorAction Ignore
+        }
+    }
+
     It "should upload a folder with files and subfolders" {
         $folderName = [System.IO.Path]::GetRandomFileName()
-        $testFolder = "$TestDrive/$folderName"
+        $testFolder = "$env:TEMP/$folderName"
 
         try {
             New-Item -ItemType Directory -Path $testFolder | Out-Null
@@ -284,6 +323,26 @@ Describe "New-GcsObject" {
             $objs.Name -contains "$folderName/TestSubfolder/pluto.txt" | Should Be $true
             $objs.Name -contains "$folderName/TestSubfolder/saturn.txt" | Should Be $true
             $objs.Name -contains "$folderName/TestSubfolder/" | Should Be $true
+        }
+        finally
+        {
+            Remove-Item $testFolder -Recurse -Force -ErrorAction Ignore
+        }
+    }
+
+    It "should upload a folder with the correct prefix" {
+        $folderName = [System.IO.Path]::GetRandomFileName()
+        $testFolder = "$env:TEMP/$folderName"
+
+        try {
+            New-Item -ItemType Directory -Path $testFolder | Out-Null
+            "Hello, world" | Out-File "$testFolder/world.txt"
+            $prefix = "Planet"
+            $result = New-GcsObject -Bucket $bucket -Folder "$testFolder/" -Force -ObjectNamePrefix $prefix
+
+            $result.Count | Should Be 2
+            $result.Name -contains "$prefix/$folderName/" | should be $true
+            $result.Name -contains "$prefix/$folderName/world.txt" | should be $true
         }
         finally
         {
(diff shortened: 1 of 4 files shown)


Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -222,9 +241,24 @@ protected override void ProcessRecord()
...
             Stream contentStream = null;
-            if (!string.IsNullOrEmpty(File))
+
+            if (ParameterSetName == ParameterSetNames.ContentsFromFolder)
+            {
+                // user gives us the path to a folder, we will resolve the path and upload the content of that folder
+                string resolvedFolderPath = GetFullPath(Folder);
+                if (string.IsNullOrWhiteSpace(resolvedFolderPath) || !Directory.Exists(resolvedFolderPath))
+                {
+                    throw new DirectoryNotFoundException(string.Format("Directory {0} cannot be found", resolvedFolderPath));
```
chrsmith: Please use the string interpolation operator. Also, putting quotation marks around the directory name makes it easier to diagnose errors.

````
$"Directory '{resolvedFolderPath}' be found."
````

File: Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1 [hunk 2]
```diff
@@ -207,6 +212,66 @@ Describe "New-GcsObject" {
         $obj.Metadata["Content-Type"] | Should Be "image/jpeg"
         Remove-GcsObject $obj
     }
+
+    It "should upload an empty folder" {
+        $TestFolder = Join-Path $TestDrive "TestFolder"
```
chrsmith: The naming convention we have been using for PowerShell is to have variable names camelCased (and not PascalCased). Is that the standard for PowerShell? If so, let's talk about this in-person as we should be consistent for now and do the major refactoring in a separate PR.

chrsmith: Also, where is `$TestDrive` defined? Why don't we just use `GetRandomFileName` [link]((https://msdn.microsoft.com/en-us/library/system.io.path.getrandomfilename(v=vs.110).aspx) instead? That would obviate the need for the if-statement later.

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
```diff
@@ -207,6 +212,66 @@ Describe "New-GcsObject" {
...
+    It "should upload a folder with files and subfolders" {
+        $TestFolder = Join-Path $TestDrive "TestFolder"
+
+        try {
+            if (-not (Test-Path $TestFolder))
+            {
+                New-Item -ItemType Directory -Path $TestFolder | Out-Null
+            }
+
+            "Hello, world" | Out-File (Join-Path $TestFolder "world.txt")
```
chrsmith: IMHO, it would be cleaner to just write `"$testFolder/world.txt"`. I understand `Join-Path` makes sense, but in this case we can be more fast and loose since it would aid readability.

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
```diff
@@ -207,6 +212,66 @@ Describe "New-GcsObject" {
...
+
+            $result = New-GcsObject -Bucket $bucket -Folder $TestFolder -Force
+
+            $result.Count | should be 7
+            $result.Name -contains "TestFolder/jupiter.txt" | should be $true
+            $result.Name -contains "TestFolder/mars.txt" | should be $true
+            $result.Name -contains "TestFolder/TestSubfolder/pluto.txt" | should be $true
+
+            $saturn = Get-GcsObject -Bucket $bucket -ObjectName "TestFolder/TestSubfolder/saturn.txt"
+            $saturn.ContentType | should be "text/plain"
```
chrsmith: Could you also add a couple of test scenarios that use the `-Prefix` and `-Delimiter` parameters to `Get-GcsObject`?

`Get-GcsObject -Delimiter "/"` should not include pluto and saturn.
`Get-GcsObject -Prefix "Test" -Recurse` should include pluto and saturn.

Google.PowerShell/Common/GcloudCmdlet.cs
```diff
@@ -87,9 +88,33 @@ protected string GetFullPath(string filePath)
                 return filePath;
             }
 
-            // For relative files, qualify them based on where the PowerShell session is.
-            string currentDirectory = SessionState.Path.CurrentLocation.Path;
-            return Path.Combine(currentDirectory, filePath);
+            ProviderInfo provider = null;
+            string[] result;
+
+            // resolve filepath and get the first one
```
chrsmith: Comments should always be complete sentences. So begin with a capital letter and end with a period.

chrsmith: ... again on line 101, and elsewhere.

Google.PowerShell/Common/GcloudCmdlet.cs
```diff
@@ -87,9 +88,33 @@ protected string GetFullPath(string filePath)
...
+            catch (ItemNotFoundException itemEx)
+            {
+                // in case the file path is not created but we should still return the resolve path
+                result = new string[] { itemEx.ItemName };
+            }
+
+            if (result.Length != 1)
+            {
+                // cannot really resolve this
+                return null;
```
chrsmith: I don't think this method should ever return null, right? If you ask to resolve a path (e.g. "foo/../bar") the worst thing you can do is just return "foo/../bar". Potentially returning null means every caller needs to check if the result is null, which greatly complicates things.

So an error mode could just be "do nothing and hope it works out".

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -117,13 +119,17 @@ protected bool TestObjectExists(StorageService service, string bucket, string ob
     /// <example>
     ///   <para>Upload a local log file to GCS.</para>
     ///   <para><code>PS C:\> New-GcsObject -Bucket "widget-co-logs" -ObjectName "log-000.txt" `</code></para>
-    ///   <para><code>    -File "C:\logs\log-000.txt"</code></para>
+    ///   <para><code>    -Path "C:\logs\log-000.txt"</code></para>
```
chrsmith: Shouldn't this still be -File? Reading the comment above it sounds like I can either use -Contents, -File, or -Path. If -File was removed, we should update the comment on line 108.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -117,13 +119,17 @@ protected bool TestObjectExists(StorageService service, string bucket, string ob
...
     /// </example>
     /// <example>
     ///   <para>Pipe a string to a a file on GCS. Sets a custom metadata value.</para>
     ///   <para><code>PS C:\> "Hello, World!" | New-GcsObject -Bucket "widget-co-logs" -ObjectName "log-000.txt" `</code></para>
     ///   <para><code>    -Metadata @{ "logsource" = $env:computername }</code></para>
     /// </example>
+    /// <example>
+    ///   <para>Upload a folder to GCS</para>
```
chrsmith: End the sentence with a period. But I'd suggest you elaborate on it. "Upload a folder and its contents to GCS. The names of the created objects will be relative to the folder."

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -117,13 +119,17 @@ protected bool TestObjectExists(StorageService service, string bucket, string ob
...
+    ///   <para><code>PS C:\> New-GcsObject -Bucket "widget-co-logs" -Folder "C:\Folder"</code></para>
```
chrsmith: `C:\Folder` doesn't quite fit with the example. How about a PowerShell equiv of: `%SystemDrive%\inetpub\logs\LogFiles`. (Where IIS logs go.)

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -147,15 +154,16 @@ private class ParameterSetNames
 
         /// <summary>
         /// <para type="description">
-        /// The name of the created Cloud Storage object.
+        /// The name of the created Cloud Storage object. Ignored if Folder is specified
```
chrsmith: End with a period.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -168,10 +176,20 @@ private class ParameterSetNames
         /// </para>
         /// </summary>
         [Parameter(Position = 2, Mandatory = true, ParameterSetName = ParameterSetNames.ContentsFromFile)]
+        [ValidateNotNullOrEmpty()]
```
chrsmith: I assume we should be applying this attribute liberally where appropriate?

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
quoctruong: It should be camelCased. My bad. I'll fix it. $TestDrive is a variable created automatically in a single Describe or Context block of Pester (https://github.com/pester/Pester/wiki/TestDrive).

Google.PowerShell/Storage/GcsObject.cs
quoctruong: Yes, so PowerShell will automatically check for us that the value given is not null or empty.

Google.PowerShell/Storage/GcsObject.cs
quoctruong: Fixed.

Google.PowerShell/Storage/GcsObject.cs
quoctruong: Fixed.

Google.PowerShell/Storage/GcsObject.cs
quoctruong: Fixed.

Google.PowerShell/Storage/GcsObject.cs
quoctruong: Fixed.

Google.PowerShell/Storage/GcsObject.cs
quoctruong: Fixed.

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
quoctruong: Added the tests with Find-GcsObject. I cannot find the -Recurse parameter for that cmdlet though?

Google.PowerShell/Common/GcloudCmdlet.cs
quoctruong: Changed the function to return the filePath that it is given if there an error

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
chrsmith: Thanks for the info, I was unaware of `$TestDrive`. FWIW all of our other Pester tests just read/write files directly from the Temp directory. So I'd actually be in favor of just continuing to do that. However, I don't feel strongly about it. So fee free to keep it in if you like.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -245,15 +245,15 @@ protected override void ProcessRecord()
...
                 string resolvedFolderPath = GetFullPath(Folder);
                 if (string.IsNullOrWhiteSpace(resolvedFolderPath) || !Directory.Exists(resolvedFolderPath))
                 {
-                    throw new DirectoryNotFoundException(string.Format("Directory {0} cannot be found", resolvedFolderPath));
+                    throw new DirectoryNotFoundException($"Directory '{resolvedFolderPath}' cannot be found.");
                 }
 
-                // call helper method to upload the content of the directory, the relative directory name of this directory
-                // will be its name
+                // Call helper method to upload the contents of the directory.
```
chrsmith: "Call helper method to upload the contents of the directory." just describes what the code does, not _why_. The second comment is useful, since it helps the next programmer. The first isn't, as the code is pretty self-descriptive. So I'd remove the one on line 255.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -289,39 +289,39 @@ private void UploadDirectoryToGCSBucket(string directory, Dictionary<string, str
...
             }
 
-            // have to append slash to create directory on the cloud
+            // We have to append / at the end to create a directory on the cloud.
             string relativeDirectoryNameWithSlash = relativeDirectoryName + "/";
 
             if (TestObjectExists(Service, Bucket, relativeDirectoryNameWithSlash) && !Force.IsPresent)
             {
-                // if the directory exists and force is not present, throw an error
+                // If the directory exists and force is not present, throw an error.
```
chrsmith: This comment isn't useful, just describes the code block.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -289,39 +289,39 @@ private void UploadDirectoryToGCSBucket(string directory, Dictionary<string, str
...
-                    // get a relative path compared to the root path, also have to replace \ with / so it will be created with correct folder structure
+                    // Get a relative path compared to the root path, also we have to replace \ with / so it will be created with correct folder structure.
                     Path.Combine(relativeDirectoryName, Path.GetFileName(file)).Replace('\\', '/'));
             }
 
             foreach (string subDirectory in Directory.EnumerateDirectories(directory))
             {
-                // recursively upload subfolder, we have to replace \ with /
+                // Recursively upload subfolder and replace \ with /.
                 UploadDirectoryToGCSBucket(subDirectory, metadataDict, Path.Combine(relativeDirectoryName, Path.GetFileName(subDirectory)).Replace('\\', '/'));
```
chrsmith: You should definitely break down these subexpressions into other lines.

Generally having more concise code is a good thing. However, if it hurts readability, like in this case, you should err on the side of being more verbose.

Introduce new local variables to make things easier to read. In this case, maybe introduce 2 or 3.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -289,39 +289,39 @@ private void UploadDirectoryToGCSBucket(string directory, Dictionary<string, str
...
             foreach (string file in Directory.EnumerateFiles(directory))
             {
-                // upload each individual file
+                // Upload each individual file.
                 UploadStreamToGCSObject(new FileStream(file, FileMode.Open),
                     GetContentType(ContentType, metadataDict, InferContentType(file)),
                     metadataDict,
-                    // get a relative path compared to the root path, also have to replace \ with / so it will be created with correct folder structure
+                    // Get a relative path compared to the root path, also we have to replace \ with / so it will be created with correct folder structure.
                     Path.Combine(relativeDirectoryName, Path.GetFileName(file)).Replace('\\', '/'));
```
chrsmith: Given that we call `.Replace('\\', '/')` in a couple of places, perhaps we can make a local function to make it clearer what is going on?

`private static string ConvertLocalToGcsFolderPath(string localFilePath) {`

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -289,39 +289,39 @@ private void UploadDirectoryToGCSBucket(string directory, Dictionary<string, str
         {
```
chrsmith: The .NET style guidelines dictate that it should be `Gcs` not `GCS`. Given that the context is pretty clear, why not just call this `UploadDirectory` and leave it at that? (Rename other methods if they do not obey the all-mighty style guide.)

chrsmith: Line 284 should use GCS since it is an acronym.

Also, `relativeDirectoryName` sounds somewhat misleading. Would `gcsObjectNamePrefix` be more accurate?

Google.PowerShell/Common/GcloudCmdlet.cs
```diff
@@ -87,9 +88,34 @@ protected string GetFullPath(string filePath)
                 return filePath;
             }
 
-            // For relative files, qualify them based on where the PowerShell session is.
-            string currentDirectory = SessionState.Path.CurrentLocation.Path;
-            return Path.Combine(currentDirectory, filePath);
+            ProviderInfo provider = null;
```
chrsmith: Is it possible to rewrite this method so that the scope of both `provider` and `result` are inside the `try` block? As it stands, the function is kind of confusing.

Perhaps:

````
// If the path is already fully-qualified, go with that.
...
// Try to resolve the path using PowerShell (only applicable for FileSystemProvider)
...
// Just default with the input path and hope for the best.
return filePath;
````

Comment (chrsmith) [inline Google.PowerShell/Common/GcloudCmdlet.cs]: For future programmers, it might be useful to explain how a cmdlet can return multiple resolved file paths. Is it in the case of ambiguities, such as when using case-sensitive file paths when PowerShell is assumed to be case-insensitive?

File: Google.PowerShell/Storage/GcsObject.cs (hunk 11 above)
Comment (ILMTitan) [inline Google.PowerShell/Storage/GcsObject.cs]: Doesn't making it mandatory provide this check anyway?

**Comment by ILMTitan** [comment]
I know you aren't the one who wrote it, but now would be a good time to change the pester test to clean up after itself by deleting the bucket it creates.

File: Google.PowerShell/Storage/GcsObject.cs (hunk 11 above)
Comment (quoctruong) [inline Google.PowerShell/Storage/GcsObject.cs]: You can still supply something like an empty string to the parameter and that doesn't get checked by mandatory.

File: Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1 [hunk 18]
```diff
@@ -214,62 +214,64 @@ Describe "New-GcsObject" {
...
+            "Hello, pluto" | Out-File "$testSubfolder/pluto.txt"
 
-            $result = New-GcsObject -Bucket $bucket -Folder $TestFolder -Force
+            $result = New-GcsObject -Bucket $bucket -Folder $testFolder -Force
 
             $result.Count | should be 7
-            $result.Name -contains "TestFolder/jupiter.txt" | should be $true
-            $result.Name -contains "TestFolder/mars.txt" | should be $true
-            $result.Name -contains "TestFolder/TestSubfolder/pluto.txt" | should be $true
+            $result.Name -contains "$folderName/jupiter.txt" | should be $true
```
quoctruong: Added the tests

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
```diff
@@ -214,62 +214,64 @@ Describe "New-GcsObject" {
...
+            $result.Name -contains "$folderName/mars.txt" | should be $true
+            $result.Name -contains "$folderName/TestSubfolder/pluto.txt" | should be $true
```
quoctruong: Capitalized them.

Google.PowerShell/Common/GcloudCmdlet.cs
```diff
@@ -91,27 +91,28 @@ protected string GetFullPath(string filePath)
...
+            // Resolve filepath.
             try
             {
                 result = GetResolvedProviderPathFromPSPath(filePath, out provider).ToArray();
             }
             catch (ItemNotFoundException itemEx)
             {
-                // in case the file path is not created but we should still return the resolve path
+                // In case the file path is not created, an error will be thrown.
+                // But we should still return the resolved path.
```
quoctruong: It should return fully qualified path. I added a check if the item name is rooted then I return it, otherwise I return ```filePath```.

Google.PowerShell/Common/GcloudCmdlet.cs
quoctruong: Added the comment. An example is if wildcard is used.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -222,9 +241,23 @@ protected override void ProcessRecord()
...
+            {
+                // User gives us the path to a folder, we will resolve the path and upload the contents of that folder.
+                string resolvedFolderPath = GetFullPath(Folder);
+                if (string.IsNullOrWhiteSpace(resolvedFolderPath) || !Directory.Exists(resolvedFolderPath))
+                {
+                    throw new DirectoryNotFoundException($"Directory '{resolvedFolderPath}' cannot be found.");
+                }
+
+                // The relative directory name of this directory will be the name of the directory itself.
+                UploadDirectory(resolvedFolderPath, metadataDict, Path.GetFileName(resolvedFolderPath));
```
ILMTitan: Do we want the prefix to have to be the folder name, or do we want the user to be able to specify the prefix (logical folder) we want to put the files in?

Google.PowerShell/Common/GcloudCmdlet.cs
```diff
@@ -87,9 +88,31 @@ protected string GetFullPath(string filePath)
...
-            return Path.Combine(currentDirectory, filePath);
+            // Try to resolve the path using PowerShell (only applicable for FileSystemProvider).
+            try
+            {
+                ProviderInfo provider = null;
+                string[] result = GetResolvedProviderPathFromPSPath(filePath, out provider).ToArray();
+
+                // Only return the resolved path if there are no ambiguities.
+                // If path contains wildcards, then it may resolved to more than 1 path.
+                if (result.Length == 1 && provider.ImplementingType == typeof(FileSystemProvider))
```
chrsmith: Let's say results is null, then this crashes. If `result?.Length ==1 ` compiles, why don't we go with that.

Google.PowerShell/Common/GcloudCmdlet.cs
```diff
@@ -87,9 +88,31 @@ protected string GetFullPath(string filePath)
                 return filePath;
             }
 
-            // For relative files, qualify them based on where the PowerShell session is.
-            string currentDirectory = SessionState.Path.CurrentLocation.Path;
-            return Path.Combine(currentDirectory, filePath);
+            // Try to resolve the path using PowerShell (only applicable for FileSystemProvider).
```
chrsmith: This update looks much, much cleaner. Thank you.

Google.PowerShell/Storage/GcsObject.cs
chrsmith: I don't like the complexity of adding yet another parameter, but it does seem wrong to always upload a folder into the root of the GCS bucket.

Why don't we add a new parameter here:

````
/// When uploading the contents of a directory into Google Cloud Storage, this is the prefix applied
/// to every object which is uploaded.
public string ObjectNamePrefix { get; set; } = "";
````

And then in this call to `UploadDirectory` we just prefix `Path.GetFileName` with the parameter's value.

File: Google.PowerShell/Storage/GcsObject.cs [hunk 24]
```diff
@@ -243,27 +276,99 @@ protected override void ProcessRecord()
...
+        /// </summary>
+        private void UploadDirectory(string directory, Dictionary<string, string> metadataDict, string gcsObjectNamePrefix)
+        {
+            if (string.IsNullOrWhiteSpace(directory) || !Directory.Exists(directory))
+            {
+                return;
+            }
+
+            // We have to append / at the end to create a directory on the cloud.
+            string gcsObjectNamePrefixWithSlash = gcsObjectNamePrefix + "/";
```
chrsmith: Would it be simpler to just overwrite the value of `gcsObjectNamePrefix` to ensure it has a trailing slash, and not introduce another local variable? Also, what happens if the `gcsObjectNamePrefix` already ends with a slash?

Perhaps:
````
// Confirm gcsObgcsObjectNamePrefix is a GCS folder.
if (!gcsObgcsObjectNamePrefix.EndsWith("/")) {
    gcsObgcsObjectNamePrefix += "/";
}
````

File: Google.PowerShell/Storage/GcsObject.cs [hunk 25]
```diff
@@ -243,27 +276,99 @@ protected override void ProcessRecord()
...
+
+            if (TestObjectExists(Service, Bucket, gcsObjectNamePrefixWithSlash) && !Force.IsPresent)
+            {
+                throw new PSArgumentException(
+                    $"Storage object '{gcsObjectNamePrefixWithSlash}' already exists. Use -Force to overwrite.");
+            }
+
+            // Create a directory on the cloud.
+            string objContentType = GetContentType(null, metadataDict, UTF8TextMimeType);
+            byte[] contentBuffer = Encoding.UTF8.GetBytes(Contents);
```
chrsmith: Looks like a copy+paste error. `Contents` shouldn't have a value for the directory case. And even if it did, we don't want to give the directory objects a value.

`var contentBuffer = new byte[] {}` should work, right?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -243,27 +276,99 @@ protected override void ProcessRecord()
...
+            Stream contentStream = new MemoryStream(contentBuffer);
+            UploadStreamToGcsObject(contentStream, objContentType, metadataDict, gcsObjectNamePrefixWithSlash);
+
+            foreach (string file in Directory.EnumerateFiles(directory))
+            {
+                // Upload each individual file.
```
chrsmith: This comment seems unnecesary.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -243,27 +276,99 @@ protected override void ProcessRecord()
...
+                string fileName = Path.GetFileName(file);
+                string fileWithGcsObjectNamePrefix = Path.Combine(gcsObjectNamePrefix, fileName);
+                // We have to replace \ with / so it will be created with correct folder structure.
+                fileWithGcsObjectNamePrefix = ConvertLocalToGcsFolderPath(fileWithGcsObjectNamePrefix);
+                UploadStreamToGcsObject(new FileStream(file, FileMode.Open),
```
chrsmith: Line breaking rules are hard. But I believe you should put `new FileStream` on a new line, rather than putting only one of the parameters to `UploadStreamtoGcsObject` on the same line.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -243,27 +276,99 @@ protected override void ProcessRecord()
...
+                    GetContentType(ContentType, metadataDict, InferContentType(file)),
+                    metadataDict,
+                    ConvertLocalToGcsFolderPath(fileWithGcsObjectNamePrefix));
+            }
+
+            foreach (string subDirectory in Directory.EnumerateDirectories(directory))
+            {
+                // Recursively upload subfolder.
```
chrsmith: We can either do without this comment, or if we keep it put it above the `foreach` loop.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -243,27 +276,99 @@ protected override void ProcessRecord()
...
+                string subDirectoryName = Path.GetFileName(subDirectory);
+                string subDirectoryWithGcsObjectNamePrefix = Path.Combine(gcsObjectNamePrefix, subDirectoryName);
+                UploadDirectory(subDirectory,
```
chrsmith: Same as before, when spilling function parameters on multiple lines, do it for all of them. Not just the second onwards.

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -243,27 +276,99 @@ protected override void ProcessRecord()
...
+        }
+
+        /// <summary>
+        /// Upload a GCS object using a stream.
+        /// </summary>
+        private void UploadStreamToGcsObject(Stream contentStream, string objContentType, Dictionary<string,string> metadataDict, string objectName)
+        {
+            if (contentStream == null)
+            {
+                return;
```
chrsmith: While we may be safe in that all callpaths to this function give us a valid stream, maybe we should allow this case. e.g. just create a zero-length MemoryStream so that it will create a 0-byte file on GCS? It seems strange that the function would silently do nothing, as opposed to throwing an InvalidArgumentException.

Google.PowerShell/Storage/GcsObject.cs
quoctruong: I think it should be the folder name. This would be consistent with the behavior of gsutil cp.

quoctruong:
PTAL

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
```diff
@@ -207,6 +223,132 @@ Describe "New-GcsObject" {
...
+            $folderOnline.Name | Should Be "$folderName/"
+            $folderOnline.Size | Should Be 0
+        }
+        finally
+        {
+            Remove-Item $testFolder -Recurse -Force -ErrorAction Ignore
+        }
+    }
+
+    It "should upload a folder with slashes at the end of the path" {
```
chrsmith: To be clearer, perhaps "should work for both forward and backslashes in the path"?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -222,9 +251,29 @@ protected override void ProcessRecord()
...
             string objContentType = null;
-
             Stream contentStream = null;
-            if (!string.IsNullOrEmpty(File))
+
+            if (ParameterSetName == ParameterSetNames.ContentsFromFolder)
+            {
+                // User gives us the path to a folder, we will resolve the path and upload the contents of that folder.
+                // Have to take care of / and \ in the end of the directory path because Path.GetFileName will return
+                // empty string if that is not trimmed off.
```
chrsmith: Perhaps "will return *an* empty string"?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -222,9 +251,29 @@ protected override void ProcessRecord()
...
+                {
+                    throw new DirectoryNotFoundException($"Directory '{resolvedFolderPath}' cannot be found.");
+                }
+
+                string gcsObjectNamePrefix = Path.GetFileName(resolvedFolderPath);
+                if (!string.IsNullOrWhiteSpace(ObjectNamePrefix))
+                {
+                    gcsObjectNamePrefix = Path.Combine(ObjectNamePrefix, gcsObjectNamePrefix);
+                }
+                UploadDirectory(resolvedFolderPath, metadataDict, ConvertLocalToGcsFolderPath(gcsObjectNamePrefix));
```
chrsmith: Seems like we might want to add a progress indicator, but that can wait for a future PR. A TODO or NOTE comment seems appropriate.

````
// NOTE: There man be many (many) files being uploaded. Consider doing a scan ahead of time
// and putting up a progress indicator as the objects are created.
````

or 

````
// TODO(quoct): Add a progress indicator if there are many files.
````

quoctruong:
Thanks!

PR #330, opened 2016-09-30
Upgrade our Google Analytics wrapper

chrsmith:
This PR upgrades the codepath we use to report data to Google Analytics. The functionality is pretty much the same, but we are now using the same format as the Google Cloud Tools for Visual Studio extension. The end result is that by reporting our data slightly different, it will be picked up by internal dashboards, etc.


chrsmith:
@ivannaranjo PTAL for all the Google Analytics bits.
@ILMTitan FYI

Google.PowerShell/Common/ReportCmdletResults.cs
```diff
@@ -0,0 +1,189 @@
...
+        /// Category for all analytic events reported through this instance.
+        /// </summary>
+        private AnalyticsEventCategory _category = AnalyticsEventCategory.CmdletInvocation;
+
+        public GoogleAnalyticsCmdletReporter(string clientID, AnalyticsEventCategory analyticsCategory)
+        {
+            // We report results to Google Analytics if the user has opted-into Cloud SDK metric reporting. However, we
+            // don't want to report data during development. (e.g. when running unit tests.)
+            string envVar = Environment.GetEnvironmentVariable("DISABLE_POWERSHELL_ANALYTICS");
+            bool debugFlag = (envVar == "TRUE");
```
ivannaranjo: Suggestion, maybe less stringent check is better. Just check that is not empty. Checking for only TRUE can make it a bit error prone.

Google.PowerShell/Common/ReportCmdletResults.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        public void ReportFailure(string cmdletName, string parameterSet, int errorCode)
+        {
+            Report(_category, cmdletName, parameterSet, errorCode);
+        }
+
+        private void Report(AnalyticsEventCategory category, string cmdletName, string parameterSet, int? errorCode)
+        {
+            _reporter.Value?.ReportEvent(
```
ivannaranjo: You are supposed to use the `ReportPageView()` method instead, analytics are reported as virtual page views with more information than events can handle.

Google.PowerShell/Common/GoogleAnalyticsReporter.cs
```diff
@@ -0,0 +1,224 @@
...
+            return result;
+        }
+
+        /// <summary>
+        /// Sends the hit data to the server.
+        /// </summary>
+        /// <param name="hitData">The hit data to be sent.</param>
+        private async void SendHitData(Dictionary<string, string> hitData)
+        {
+            using (var client = new HttpClient())
```
quoctruong: I think HttpClient is supposed to be reused (https://www.asp.net/web-api/overview/advanced/calling-a-web-api-from-a-net-client) so maybe we shouldn't create a new instance of HttpClient everytime.

chrsmith:
Abandoning the PR. Need to branch from a different commit of the VS plug-in code.

PR #335, opened 2016-10-10
Add .NET Core compilation with project.json. Also updated the documentation

quoctruong:


Google.PowerShell.DotnetCore/project.json
```diff
@@ -0,0 +1,58 @@
...
+
+  "frameworks": {
+    "netstandard1.6": {
+      "imports": "dnxcore50",
+      "buildOptions": {
+        "define": [ "CORECLR" ]
+      },
+      "dependencies": {
+        "System.Management.Automation": "1.0.0-*",
+        "Portable.BouncyCastle": "1.8.1.1",
```
chrsmith: I don't believe we depend on this directly, but rather bring it in through the Google client library infrastructure. So it might be OK to remove it.

Google.PowerShell/Common/CloudSdkSettings.cs
```diff
@@ -46,7 +49,11 @@ public class CommonProperties
         /// </summary>
         public static string GetCurrentConfigurationName()
         {
+#if !UNIX
```
chrsmith: Since we have the same #ifdef in two places, why don't we just create a private helper function, and just ifdef it out once in a more general: `GetCloudSdkFolder`. We can use the result of that to call `Path.Combine` with whatever Cloud SDK files we need.

Google.PowerShell/Common/MeasurementProtocolService.cs
```diff
@@ -8,7 +8,9 @@
 using System.Linq;
 using System.Net;
 using System.Text;
+#if !CORECLR
```
chrsmith: Please add a comment to call this out. What is the difference between the CORECLR build and the non-Core one? Is the feature using a different set of libraries, or removed entirely?

Google.PowerShell/Common/MeasurementProtocolService.cs
```diff
@@ -61,6 +63,9 @@ public static void SetClientId(string clientId)
             ClientId = clientId.Trim();
         }
 
+        // Removed this part of the code for CoreCLR for now, Chris is working on a PR
+        // that will remove this file.
+#if !CORECLR
```
chrsmith: Rather than ifdeffing out the method entirely AND its callsite, why don't we just stub it out? The result would be much easier to read IMHO. e.g.

````
public static HttpWebRequest GenerateRequest(string category, string action, string label, int? value = null) {
    #if CORECLR
    // TODO(quoct): Wire telemetry reporting into PowerShell for non-Windows platforms.
    // This requires finding an alternative to current use of System.Web.
    return;
    #else
    ...
    #endif
}
````

We might need a `    category = action = label = null;  // Prevent warnings.` in case the compiler whines.

File: Google.PowerShell/Compute/GceMetadataCmdlet.cs [hunk 5]
```diff
@@ -115,8 +117,10 @@ public class GetGceMetadataCmdlet : GCloudCmdlet
 
         /// <summary>
         /// Make this a field, so it can be aborted if cmdlet stops e.g. the user hits Ctrl-C.
+        /// Many properties of HttpWebRequest are not available on CoreCLR so it is better
```
chrsmith: Mentioning HttpWebRequest doesn't make sense, since the future version of the code will only have HttpClient. Right? So I'd just drop these comments, but keep your refactorings.

Google.PowerShell/Provider/GcsContentWriter.cs
```diff
@@ -40,7 +41,13 @@ public void Seek(long offset, SeekOrigin origin)
 
         public void Close()
         {
+#if !CORECLR
```
chrsmith: Would it make sense to just always call `Dispose`? I assume that `Dispose` calls `Close` under the hood?

chrsmith: Same below in GcsStringReader.

PR #341, opened 2016-10-10
Upgrade our Analytics wrapper

chrsmith:
This PR does the same as #330 , except uses the right code path. (The `q3_master` branch of the VS plug-in.)

Replaces our existing MeasurementProtocolService class with a AnalyticsReporter class. Also differentiates between cmdlet invocations and PowerShell provider commands.


chrsmith:
@ivannaranjo PTAL

Google.PowerShell/Common/ReportCmdletResults.cs
```diff
@@ -0,0 +1,189 @@
...
+        }
+
+        public void ReportFailure(string cmdletName, string parameterSet, int errorCode)
+        {
+            Report(_category, cmdletName, parameterSet, errorCode);
+        }
+
+        // The virtual page view reported to analytics.
+        private static string GetPageViewUri(string cmdletName, string parameterSet) =>
+            $"/virtual/{cmdletName}/{parameterSet}";
```
ivannaranjo: Your eventType, what you are calling cmdletName, should be "powershell", in parallel with what we're doing in visual studio. The eventName should be "cmdletInvokation". Any other parameters can be passed in the parameters in the event metadata, such as the cmdlet name, parameter group, status, etc...

chrsmith:
PTAL. Ported over the `EventsReporter` and `AnalycisReporter` classes. thanks for calling that out, as I was unaware of the required parameters.

chrsmith:
Pint @ivannaranjo

Google.PowerShell/Common/ReportCmdletResults.cs
```diff
@@ -0,0 +1,206 @@
...
+                "cmdletName", cmdletName,
+                "parameterSet", (parameterSet ?? "null"),
+                "errorCode", (errorCode.HasValue ? errorCode.ToString() : "null"));
+
+            s_reporter.Value.ReportEvent(
+                source: "virtual.powershell",
+                eventType: "powershell",
+                eventName: eventData.Name,
+                userLoggedIn: true,
+                projectNumber: null,
```
ivannaranjo: Nit: Think about providing the project number if possible, it will help us know how many projects are actually using the cmdlets.

chrsmith: Ack. I'll put a TODO in and wire that in in a future CL. It should be possible for us to get that information from the current Cloud SDK config.

However, some commands cross projects. e.g. you can access GCS bucket foo, which is "owned" by a project different than the one in the current Cloud SDK config. Though, I assume gsutil has the same problem.

PR #343, opened 2016-10-13
Issue #337 whatif support

### LINKED ISSUE #336: dir in gs:\ fails with no explanation given
Can we intercept an `AggregateException`, and provide a more reasonable error (e.g. call `.Flatten`)?

This is when running the currently released version of the cmdlets. When building/debugging locally, it works fine.

````
PS C:\Users\chrsmith> Import-Module GoogleCloud
WARNING: The names of some imported commands from the module 'GoogleCloud' include unapproved verbs that might make
them less discoverable. To find the commands with unapproved verbs, run the Import-Module command again with the
Verbose parameter. For a list of approved verbs, type Get-Verb.
PS C:\Users\chrsmith> cd\
PS C:\> cd gs:\
PS gs:\> dir | select Name
Get-ChildItem : One or more errors occurred.At line:1 char:1
+ dir | select Name
+ ~~~
    + CategoryInfo          : NotSpecified: (:) [Get-ChildItem], AggregateException
    + FullyQualifiedErrorId : System.AggregateException,Microsoft.PowerShell.Commands.GetChildItemCommand
````

### LINKED ISSUE #337: `rmdir -WhatIf` in the GCS provider actually deletes the data
```
PS gs:\> rmdir gootoso-test-bucket -WhatIf

Confirm
The item at gs:\gootoso-test-bucket has children and the Recurse parameter was not specified. If you continue, all children will be removed
with the item. Are you sure you want to continue?
[Y] Yes  [A] Yes to All  [N] No  [L] No to All  [S] Suspend  [?] Help (default is "Y"): a
PS gs:\> dir gootoso-test-bucket
dir : Google.Apis.Requests.RequestError
Not Found [404]
Errors [
    Message[Not Found] Location[ - ] Reason[notFound] Domain[global]
]
At line:1 char:1
+ dir gootoso-test-bucket
+ ~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Get-ChildItem], GoogleApiException
    + FullyQualifiedErrorId : Google.GoogleApiException,Microsoft.PowerShell.Commands.GetChildItemCommand
```

### PR DESCRIPTION
**Pull request by ILMTitan**
Add ShouldProcess calls to relevant methods.
Fix cache out of date after Set-Content.
Attempt to fix aggrigate error message in Get-ChildItem.

Fixes #337 and #338. Tries to fix #336.

### CODE REVIEW

File: Google.PowerShell.IntegrationTests/Provider/Storage.Provider.Tests.ps1 [hunk 1]
```diff
@@ -71,6 +71,15 @@ Describe "Storage Provider"{
         Compare-Object $gcsContents $localContents | Should BeNullOrEmpty
     }
 
+    It "Should create file using set-content" {
```
chrsmith: using Set-Content. (Capitalize the cmdlet noun and verb).

Google.PowerShell.IntegrationTests/Provider/Storage.Provider.Tests.ps1
```diff
@@ -71,6 +71,15 @@ Describe "Storage Provider"{
...
+        cd gs:\$bucketName\$folderName
+        rm $fileName
+        Set-Content $fileName -Value $content1
+        Sleep 1
```
chrsmith: Is this necessary? My understanding is that GCS writes are atomic, thought I wouldn't be surprised if the delay were needed.

chrsmith: Same below. Did you add the `Sleep` on line 144 because the test failed without it? Or was it "just in case"? If it was the latter, let's remove it until it is a problem.

Google.PowerShell/Provider/CacheItem.cs
```diff
@@ -37,5 +37,13 @@ public CacheItem(Func<T> update, TimeSpan cacheLifetime)
             _update = update;
             _cacheLifetime = cacheLifetime;
         }
+
+        /// <summary>
+        /// Forces the CacheItem to do a new update the next time the value is requested.
+        /// </summary>
+        public void Obsolete()
```
chrsmith: `Obsolete` sounds a bit more like an adjective/property to me. Perhaps `Invalidate` or `ForceRefresh` would be more appropriate?

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
```diff
@@ -480,14 +481,19 @@ protected override void GetChildItems(string path, bool recurse)
                             {
                                 GetChildItems(bucket.Name, true);
                             }
-                            catch (Exception e)
+                            catch (GoogleApiException e) when (e.HttpStatusCode == HttpStatusCode.Forbidden) { }
```
chrsmith: Given that `catch (Exception) { }` is such a bad code smell, please leave a comment for why it is OK to silently swallow 403 errors, since it isn't immediately obvious. (i.e. it seems like a bug, right?)

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
```diff
@@ -480,14 +481,19 @@ protected override void GetChildItems(string path, bool recurse)
...
+                            catch (AggregateException e)
                             {
-                                if ((e as GoogleApiException)?.HttpStatusCode != HttpStatusCode.Forbidden)
+                                foreach (Exception innerException in e.InnerExceptions)
                                 {
-                                    WriteError(
-                                        new ErrorRecord(e, null, ErrorCategory.NotSpecified, bucket.Name));
+                                    WriteError(new ErrorRecord(
```
chrsmith: How long is the next line? I cannot even see it right in GitHub's side-by-side view.

Perhaps declare `var error = new ErrorRecord(` as a separate local variable, and spread its parameters across a couple of lines?

chrsmith: Same below on line 495.

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
```diff
@@ -522,6 +528,10 @@ protected override void GetChildItems(string path, bool recurse)
         /// <param name="newItemValue">The value of the item to create. We assume it is a string.</param>
         protected override void NewItem(string path, string itemTypeName, object newItemValue)
         {
+            if (!ShouldProcess(path, "New-Item"))
```
chrsmith: `$"New-Item '{path}'"`?

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
```diff
@@ -672,6 +686,7 @@ public IContentWriter GetContentWriter(string path)
             request.UploadAsync();
             IContentWriter contentWriter = new GcsContentWriter(inputStream);
             TelemetryReporter.ReportSuccess(nameof(GoogleCloudStorageProvider), nameof(GetContentWriter));
+            BucketModels.Clear();
```
chrsmith: As per-usual, assumne the next person reading the code isn't that smart. (Like me.)

Why are we clearing the bucket models? Is it just because we don't want to update the internal model with the results of the operation? If so, that's fine... but let's leave a brief comment to explain as much.

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
```diff
@@ -717,6 +736,10 @@ public object ClearContentDynamicParameters(string path)
         /// non-empty bucket.</param>
         protected override void RemoveItem(string path, bool recurse)
         {
+            if (!ShouldProcess(path, "Remove"))
```
chrsmith: `Remove-Item`, right?

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
chrsmith: Please add the file path here, and below in `Clear-Content` and `Remove-Item`. Using (cross-plat) PowerShell, it shows the path:

````
PS /Users/chrsmith/src/google-cloud-powershell> New-Item test.txt -WhatIf                                                                                                                                   
What if: Performing the operation "Create File" on target "Destination: /Users/chrsmith/src/google-cloud-powershell/test.txt".
````

We don't need to be as verbose, but we should at least include the file path.

File: Google.PowerShell/Provider/GoogleCloudStorageProvider.cs [hunk 9]
```diff
@@ -768,6 +791,7 @@ private void RemoveBucket(GcsPath gcsPath, bool removeObjects)
             if (removeObjects)
             {
                 DeleteObjects(gcsPath);
+                Thread.Sleep(100);
```
chrsmith: Like before, is this absolutely necessary? Was adding a delay here necessary to prevent errors when deleting the bucket, e.g. it would throw a 409? If not, let's remove it. (Or, catch the 409 exception and try again once more.)

Static sleeps are a code smell.

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
```diff
@@ -893,7 +917,9 @@ private Bucket NewBucket(GcsPath gcsPath, NewGcsBucketDynamicParameters dynamicP
         }
 
         private void ObsoleteBucketCache()
-        { }
+        {
+            BucketCache.Obsolete();
```
chrsmith: How many times is `ObsoleteBucketCache` called? Given that the function name is 100% transparent, why not replace its callsites to simply `BucketCache.Obsolete`? YAGNI

Google.PowerShell.IntegrationTests/Provider/Storage.Provider.Tests.ps1
ILMTitan: Done.

Google.PowerShell.IntegrationTests/Provider/Storage.Provider.Tests.ps1
ILMTitan: I did find they were required.

Google.PowerShell/Provider/CacheItem.cs
ILMTitan: Done.

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
ILMTitan: Added a comment about why this might occur and that ignoring those objects is intended.

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
ILMTitan: Line 495 is 111 characters long. I believe we cut lines of at 120.

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
ILMTitan: As it is written, the output is this:

    PS gs:\jimwp-test-bucket\> new-item x -whatif
    What if: Performing the operation "New-Item" on target "jimwp-test-bucket\x".

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
ILMTitan: It would fail occasionally. Changing to the "catch 409 => retry" method.

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
ILMTitan: Done.

PR #347, opened 2016-10-17
Normalize example syntax

chrsmith:
We've been using the https://github.com/red-gate/XmlDoc2CmdletDoc tool to put cmdlet documentation in-line with the source code. This PR normalizes the syntax we use for writing out examples.

Previously we were inconsistent in how we encoded examples. Sometimes we had text before the code snippet, sometimes afterwards. Sometimes we had `<code>` elements inside of `<para>` elements, other times we had it the other way around.

With this PR, all of our examples will render correctly in the new cmdlet reference website. ([Current demo](chrsmith.github.io/google-cloud-powershell/#/).)

The exact rules for how things are encoded are documented [here](https://github.com/red-gate/XmlDoc2CmdletDoc#examples-1) and [here](https://msdn.microsoft.com/en-us/library/bb736335.aspx). But ultimately it boils down to:

- **SHOULD NOT** put `<para>` blocks before the `<code>` block. These are interpreted as "introduction" text. But the way Microsoft outputs introduction text is immediately before the cmdlet example, and by convention is only "PS C:\>". We should just ignore that.
- **MUST** only have one `<code>` block. Subsequent `<code>` blocks in the `<example>` will be ignored.
- **MUST NOT** use HTML-escape characters, e.g. `&gt;` to represent `>`. These don't get converted correctly. Just use `>` in the `<code>` block.
- **MUST NOT** include `<para>` blocks inside of a `<code>` block. If you need newlines, just put them inside of the `<code>` element.
- **MAY** have multiple `<para>` elements after the `<code>` block.

### DIFF
--- a/Google.PowerShell/Compute/GceHealthCheckCmdlets.cs
+++ b/Google.PowerShell/Compute/GceHealthCheckCmdlets.cs
@@ -13,30 +13,30 @@ namespace Google.PowerShell.ComputeEngine
     /// <summary>
     /// <para type="synopsis">
     /// Adds a Google Compute Engine health check.
-    /// </para>
+    ///</para>
     /// <para type="descritpion">
     /// Adds a Google Compute Engine health check. Use this cmdlet to create both HTTP and HTTPS health checks.
     /// https://cloud.google.com/compute/docs/load-balancing/health-checks
-    /// </para>
+    ///</para>
     /// <example>
-    ///     <code> PS C:&gt; Add-GceHealthCheck "my-health-check" -Project "my-project" -Http </code>
-    ///     <para> Adds an HTTP health check to the project named "my-project". </para>
+    ///   <code>PS C:\> Add-GceHealthCheck "my-health-check" -Project "my-project" -Http</code>
+    ///   <para>Adds an HTTP health check to the project named "my-project".</para>
     /// </example>
     /// <example>
-    ///     <code> PS C:&gt; Add-GceHealthCheck "my-health-check" -Https </code>
-    ///     <para> Adds an HTTPS health check to the project in the Cloud SDK config. </para>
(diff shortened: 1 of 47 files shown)


chrsmith:
@quoctruong @ILMTitan PTAL. Not a functional change, but moving forward all cmdlet examples should follow this style. I can give you the gory details about how our previous, inconsistent approach wouldn't render correctly.

Google.PowerShell/Compute/GceHealthCheckCmdlets.cs
```diff
@@ -13,30 +13,30 @@ namespace Google.PowerShell.ComputeEngine
     /// <summary>
     /// <para type="synopsis">
     /// Adds a Google Compute Engine health check.
-    /// </para>
+    ///</para>
```
chrsmith: Yes, thanks. That appears to be some find-replace collateral damage.

chrsmith:
Done. PTAL.

Google.PowerShell/Compute/GceInstanceCmdlets.cs
```diff
@@ -16,34 +16,34 @@ namespace Google.PowerShell.ComputeEngine
     /// <summary>
     /// <para type="synopsis">
     /// Gets information about one or more Google Compute Engine VM instances.
-    /// </para>
+    ///</para>
```
quoctruong: There is no space before ```</para>``` in this file too.

chrsmith: Another good catch. Sorry for missing this.

chrsmith:
PTAL

chrsmith:
Yup, I'll take care of that and submit.

PR #348, opened 2016-10-17
New cmdlet reference website

chrsmith:
This PR contains an entirely new cmdlet reference website, to replace what we currently have at googlecloudplatform.github.io/google-cloud-powershell/.

This summer, @marceloyuela prototyped a new Angular-based website for rendering cmdlet documentation. Rather than using Jekyll to generate static pages, we would load cmdlet documentation as a large JSON object, and allow for snappier navigation.

I started with the prototype, implemented the mocks put together by @hray2003, and 158 commits latter produced a demo at chrsmith.github.io/google-cloud-powershell/.

The design is actually quite simple. Here are the major components.

- GenerateWebsiteData.ps1 - I did a lot of work refactoring the way we extract cmdlet documentation and write it to JSON. The hardest part was wiring parameter sets to parameters and vice versa.
- ContentCtrl - This controller loads all the JSON documentation (currently < 1MiB) and depending on the URL route, loads a different HTML template to render the specific product (Cloud Storage), resource (GceBucket), or cmdlet (New-GceBucket).
- CmdletExplorer - This directive is the left-nav of the website.
- ParametersWidget, SyntaxWidget - These directives render a cmdlet's parameters and syntax (parameter set). In order to have users click a parameter set and show/hide parameters, we do some schannigans with the parent scope (ContentCtrl) and broadcasting a message.
- Styles.css - To best implement the mocks while still avoiding a dependency on Google's internal build system/tools, I had to write all the CSS myself. There are a few warts to get elements vertically aligned, etc. But my understanding is that is par for the course.


chrsmith:
@iantalarico Could you please review this PR, since most of it is JavaScript and Angular? If you need me to stop by in-person and explain it, I'm happy to. Also, I'm sorry for making you do this.

@ILMTitan @quoctruong FYI.

Rui and Yan have a few suggestions I'll be working on, and depending on how long they take I'll ninja them into this PR or save them for later.

website/index.html
```diff
@@ -1,37 +1,84 @@
...
-              .ready(function() {
-                angular.bootstrap(document, ['gcloud']);
-              });
-          });
-      }());
-    </script></body>
+      <div id="top-nav-container">
+        <div id="top-nav">
+          <div class="top-navSection psr-navText">
+            <a href="/google-cloud-powershell/#/">
```
iantalarico: Might be nice to have a link generator util in the controller to make changes easier and so "/google-cloud-powershell/" isnt hard coded everywhere

website/js/app/app.js
```diff
@@ -0,0 +1,124 @@
...
+            return input.substring(splitPoint + 1);
+        };
+    });
+
+    // Filter for rewriting raw text as HTML. To be passed to the ng-bind-html directive, which
+    // also takes care of sanitization.
+    // - Converts arrays of strings to separate <p> elements.
+    // - Puts quoted strings in <code class="text"> elements.
+    // - Puts references to cmdlets in <code class="cmdlet"> elements.
+    app.filter('applyHtmlStyling', function() {
```
iantalarico: Can this not be done in a template? This is a little hairy.

website/js/app/controllers/cmdlet-explorer.js
```diff
@@ -0,0 +1,65 @@
...
+
+  this.isResourceSelected = function(resource) {
+    return (this.selectedResource == resource);
+  };
+
+  // Returns whether or not we should highlight the product, resource, or cmdlet. We only highlight
+  // a row if we don't know what is "below" it. e.g. if a cmdlet is selected, we don't highlight
+  // the parent resource and product. 
+  this.isHighlighted = function(kind, name) {
+    if (kind == 'product') {
```
iantalarico: Should these strings be constants?

website/js/app/controllers/content-product.ng
```diff
@@ -0,0 +1,235 @@
+﻿<div
```
iantalarico: Is the text here generated?

website/js/app/controllers/content.js
```diff
@@ -0,0 +1,182 @@
...
+ */
+app.controller('ContentController', function($scope, $http, $routeParams) {
+  // The header to be displayed on the content page. This isn't referenced in
+  // .ng files because of a quirk of the current design. (How the header's
+  // background spans the entire page, not just the "content" section.)
+  this.contentHeader = '';
+  // Error message to display in case of an error loading data.
+  this.errorMessage = null;
+
+  // Current product, resource, and cmdlet. Some may be null. e.g.
```
iantalarico: They may be null but start as empty string? Is that correct?

website/js/app/controllers/content.js
```diff
@@ -0,0 +1,182 @@
...
+
+  // Populate member variables with the right data based on routeParams and cmdletDocumentation.
+  // Call this after this.cmdletDocumentation has been set.
+  this._loadContent = function() {
+    if (!this.cmdletDocumentation) return;
+
+    // Hierarchy of resources from Home > Product > Resource > Cmdlet.
+    this.contentHeader =
+        this.currentCmdlet || this.currentResource ||
+        this.currentProduct || "Google Cloud Tools for PowerShell";
```
iantalarico: There are a fair amount of strings just laying around.  Maybe make them all constants?

Tools/GenerateWebsiteData.ps1
```diff
@@ -1,74 +1,264 @@
...
 $cmdlets = Get-Command -Module "Google.PowerShell"
 
-# Mapping between cmdlet name and the Cloud resource. This is used to organize
-# the cmdlets on the generated website.
-$apiMappings = @{
-    "Gcs" = "Google Cloud Storage"
-    "Gce" = "Google Compute Engine"
-    "GcSql" = "Google Cloud SQL"
-    "Gcd" = "Google Cloud DNS"
+function ConvertToString($obj)
```
chrsmith: This function can be removed. Deleted.

Tools/GenerateWebsiteData.ps1
```diff
@@ -1,74 +1,264 @@
...
+        }
+
+
+        if ($matchingParameterSet -eq $null) {
+            Write-Warning "Unable to map syntax to parameter set for $cmdletName"
+            exit
+        }
+
+        Write-Host "`tMarking syntax for parameter set $($matchingParameterSet.HackName)"
+        $syntaxObj | Add-Member "parameterSet" $matchingParameterSet.HackName
```
chrsmith: /shrug, I'm inclined to leave it as is.

If we were to define a constant we would only use it in one place. So it would make things slightly more complicated since it would separate definition from usage. Putting all the definitions in the same place would make them easier to update, but it's unlikely that we would want to rename the "parameterSet" field name.

website/README.md
```diff
@@ -1,46 +1,43 @@
...
 
-To run the website, you need to install Ruby and Jekyll.
+## Running
 
-    choco install ruby -y
-    gem install jekyll
+The easiest way to run the website locally for testing is to launch it using
+the Google App Engine dev appserver. The `app.yaml` file contains the necessary
+URL handling schenanigans so that the website will work the same as when it is
+hosed on GitHub Pages.
```
chrsmith: Done.

website/css/style.css
```diff
@@ -0,0 +1,563 @@
+/**
+reset5 © 2011 opensource.736cs.com MIT
```
chrsmith: Done. Broken into: `reset.css`; `base.css` which contains the header, content, etc.; and 'cmdlet.css', which contains styles specific to the cmdlet page such as syntax, parameters, etc.

website/css/style.css
```diff
@@ -0,0 +1,563 @@
...
+https://code.google.com/archive/p/reset5/
```
chrsmith: TODO added. Adding a CSS/JS compiler would help, but I don't think we are that desperate yet.

website/css/style.css
```diff
@@ -0,0 +1,563 @@
...
+
+/**
+ * Global styles.
+ */
+html,body {
+    background-color: #FAFAFA;
```
chrsmith: Done.

website/css/style.css
```diff
@@ -0,0 +1,563 @@
...
+    font-family: "Open Sans";
+    font-weight: 100;
+}
+
+/**
+ * #left-nav, #left-nav-top, #left-nav-bottom
+ * The left nav pane, browsing cmdlets. Broken into two sections.
+ */
+#left-nav {
+    border-bottom: 1px solid rgba(0,0,0,0.08);
```
chrsmith: It's not a super nit. The preferred way would be the RGB hex, as that is the most concise. 

However, I need to use `rgba` in order to specify an opacity. The mocks use terms like "black, 8%". And converting that to hex is hard.

website/css/style.css
```diff
@@ -0,0 +1,563 @@
...
+    padding: 12px 12px;
+    background-color: rgba(0,0,0, 0.04);
+    margin-bottom: 16px;
+
+    overflow: auto;
+}
+
+
+/** The the applyHtmlStyling filter for more information. */
+code { font-family: "Roboto Mono"; }
```
chrsmith: Done.

website/js/app/app.js
```diff
@@ -0,0 +1,124 @@
...
+
+    // Directive for the left-nav, the table of contents.
+    app.directive('cmdletExplorer', function() {
+        return {
+            restrict: 'E',
+            templateUrl: prefix + 'cmdlet-explorer.ng',
+            controller: 'CmdletExplorerController',
+            controllerAs: 'cmdletExplorerCtrl',
+            scope: {
+                documentationObj: '='
```
chrsmith: None of the bindings need to be two way. My understanding is that this syntax means "the directive has a variable on its scope called `documentationObj` and you set it by setting an attribute on the HTML element called `documentationObj`. Do I have that right?

I skimmed through the docs on directives and didn't see anything about data binding.
https://docs.angularjs.org/guide/directive

Is there another way I can express this to mean "make a one-way copy of the data"?

website/js/app/app.js
chrsmith: Unfortunately it can't. For cmdlet descriptions we regularly reference other cmdlets and included string literals. But it is just typed as a string.

We could "preprocess" those strings before we bind it on the view. e.g. "cmdletInfo.description" vs "cmdletInfo.descriptionInHtml", but we would still need to use ng-bind-html. At least with it being a filter it is easy to express. e.g.:

`<div ng-bind-html="contentCtrl.cmdletInfo.description | applyHtmlStyling"></div>`

website/js/app/controllers/content.js
```diff
@@ -0,0 +1,182 @@
+var app = angular.module('powershellSite');
```
chrsmith: Done.

website/js/app/controllers/content.js
chrsmith: It is not correct. Fixed.

website/js/app/controllers/content.js
chrsmith: Teasing all these strings out into a separate object.

website/index.html
chrsmith: I definitely agree with the point, but I'm not convinced it would be net simpler. e.g. using a filter:

` <a href="{{ '/' | makeLink }}">`

or use a value on $scope

` <a href="{{ linkPrefix }}/">`

Given that it would be one-line to bulk fix using `grep` and `sed` I'm inclined to leave it as-is.

website/js/app/controllers/content-product.ng
chrsmith: No. This was all written by the team in the past.

website/js/app/app.js
iantalarico: Yea its really not documented well in angular.  Here is a pretty good article (http://onehungrymind.com/angularjs-sticky-notes-pt-2-isolated-scope/).  It just reduces the number of watches.

Not really important here.

website/js/app/app.js
iantalarico: Hmm okay, I feel like there should be a better way to do this but probably not... because javascript.

chrsmith:
Fixed the issues raised by the designers. [The demo](chrsmith.github.io/google-cloud-powershell/#/) has been updated. I'll submit tomorrow afternoon (and push live) if nobody has any objections.

- Reworked nav behavior on small screens. Removed the jank top-nav, instead on small screens the left-nav floats above content. Added a new JavaScript file small-screen-menu-behavior.js to handle all the events. (I went with just responding to click events on the HTML elements, opposed to wiring up a new Angular controller, etc.)
- Refactor CSS. I updated the CSS to override element defaults rather than requiring specific classes. This not only cleaned up the HTML, but also fixed some issues with font sizes arising from CSS selectors being overridden by the default stylesheet.
- Word wrap example text, rather than having the unintuitive scroll bars. It isn't great, but everyone disliked the alternative. 
- Make entire table cells clickable on the Homeage and Product pages.
- Gray out type and optional metadata for parameters when not related to the active parameter set.
- Fixed font sizes to actually match the mocks. (Previously everything was larger than it should be by 15% or so.)

website/css/base.css
```diff
@@ -207,24 +169,40 @@ pre {
...
+    border: 1px solid rgba(0,0,0,0.08);
+    min-width: 100%;
+}
 
-/** The the applyHtmlStyling filter for more information. */
-code {
-    font-family: "Roboto Mono";
+table.psr-linkTable,
+.psr-linkTable th {
+    padding: 16px 12px 16px 12px;
```
iantalarico: Nit: this can just be 16px 12px

And below

website/js/app/small-screen-menu-behavior.js
```diff
@@ -0,0 +1,56 @@
+// When the window is small, #header-icon becomes a toggleable menu icon which will show/hide
+// #left-nav. (Which will then float on top of #content-container.) This file implements the
+// behavior of the menu icon.
+(function() {
```
iantalarico: Discussed offline how to do this with angular.  Not required though.

chrsmith: Ack. I've added a TODO with some notes, but will leave it as-is. Evidently you need to still let angular know about window resizes, since otherwise it wouldn't be aware that window.clientWidth is updated. (Or something.) A cursory search reveled this wasn't as simple as it sounds.

chrsmith:
Stop tweaking things. But one last feature:

By reusing the cached `cmdletdocumentation` object attached to `$rootScope`, I can now support cmdlet-cross references. e.g.
https://chrsmith.github.io/google-cloud-powershell/#/google-cloud-sql/GcSqlInstance/Add-GcSqlInstance mentions cmdlets in the parameter descriptions, which are proper hyperlinks to the mentioned cmdlets.

Issue #334, opened 2016-10-10
Incorrect behavior when gcloud has two accounts


chrsmith:
When gcloud has two different accounts available, it appears that the active gcloud config account is NOT being used by PowerShell. This should "just work" because of the we are using the Application Default Credentials when initializing the Google API client libraries.

However, this is the behavior I am seeing:

````
PS C:\src\github.com\chrsmith\google-cloud-powershell\Google.PowerShell\bin\Debug> gcloud auth list
Credentialed Accounts:
 - chrsmith@google.com ACTIVE
 - chrsmithdemos@gmail.com
To set the active account, run:
    $ gcloud config set account `ACCOUNT`
````

The active account is my @google.com one, so it should have access to `gcloud-powershell-testing`. However, that doesn't appear to be the case. (The extra text "Handler was created" and "Abnormal response" is from hooking up an ILogger to the client libraries.)

````
PS C:\src\github.com\chrsmith\google-cloud-powershell\Google.PowerShell\bin\Debug> Get-GcsBucket -Project gcloud-powershell-testing
Handler was created. SupportsRedirectConfiguration=True, SupportsAutomaticDecompression=True
Add parameter should not get null values. type=Query, name=key
An abnormal response wasn't handled. Status code is Forbidden
Abnormal response is being returned. Status Code is Forbidden
Get-GcsBucket : Google.Apis.Requests.RequestError
Forbidden [403]
Errors [
    Message[Forbidden] Location[ - ] Reason[forbidden] Domain[global]
]
At line:1 char:1
+ Get-GcsBucket -Project gcloud-powershell-testing
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Get-GcsBucket], GoogleApiException
    + FullyQualifiedErrorId : Google.GoogleApiException,Google.PowerShell.CloudStorage.GetGcsBucketCmdlet
````

However, if I use the Cloud Storage provider, I get the list of accounts for all of the projects associated with my @gmail.com account.

````
PS C:\src\github.com\chrsmith\google-cloud-powershell\Google.PowerShell\bin\Debug> dir gs:\ | select Name

Name
----
blog-posts
chrsmith-demos.appspot.com
chrsmith-pictures
database-snapshots-prod
shopbot9000
staging.chrsmith-demos.appspot.com
[...]

PR #349, opened 2016-10-17
Fix authentication issue

quoctruong:
Currently, we are relying on application_default_credentials.json for authentication. However, this file is no longer generated by default on the new Google Cloud SDK. Also, a change in the active user is also not reflected in this file (see https://github.com/GoogleCloudPlatform/google-cloud-powershell/issues/334).

There are 2 parts to the fix:
1. Get the access token for the current active user with "gcloud auth print-access-token"
2. Get a new token if the old one expired or if the active user has changed.

We detect who is the active user by first figuring out what is the active Google Cloud configurations. The hierarchy of this is:
1. There is a file named properties which is located where Google Cloud SDK is installed. This file contains machine wide Google Cloud configurations, including who is the active user.
2. The current user config directory
   a. $env:APPDATA\gcloud\
   b. If $env:CLOUDSDK_CONFIG is set, then this takes precedence over the location in (a)
When we know the current user config directory, we can look into the active_config file in this directory to figure out who is the active user.
3. $env:CLOUDSDK_ACTIVE_CONFIG_NAME
4. Flags passed to gcloud (ignored this since it is not applicable to our cmdlet).

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/349)

### DIFF
--- a/Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
+++ b/Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
@@ -1,0 +1,138 @@
+﻿using NUnit.Framework;
+using Google.Apis.Util;
+using Google.PowerShell.Common;
+using System;
+using System.Reflection;
+using System.Threading;
+
+namespace Google.PowerShell.Tests.Common
+{
+    [TestFixture]
+    /// <summary>
+    /// Assumes that the Cloud SDK is installed on the local machine and initialized.
+    /// 
+    /// </summary>
+    internal class TestActiveUserCredential
+    {
+        private PropertyInfo TokenProperty = typeof(ActiveUserCredential).GetProperty(
+                "Token",
+                BindingFlags.NonPublic | BindingFlags.Static);
+        private CancellationToken CancelToken = new CancellationToken();
+
+        /// <summary>
+        /// This test checks that an ActiveUserToken is created when a new
+        /// ActiveUserCredential is created.
+        /// </summary>
+        [Test]
+        public void TestActiveUserCredentialConstructor()
+        {
+            ActiveUserCredential activeUserCred = new ActiveUserCredential();
+            ActiveUserToken activeUserToken = TokenProperty.GetValue(null, null) as ActiveUserToken;
+            Assert.IsNotNull(activeUserToken, "ActiveUserToken should be created.");
+        }
+
+        /// <summary>
+        /// This test checks that a call to RefreshTokenAsync will give us a new token.
+        /// </summary>
+        [Test]
+        public void TestRefreshTokenAsync()
+        {
+            ActiveUserCredential activeUserCred = new ActiveUserCredential();
+            ActiveUserToken activeUserToken = TokenProperty.GetValue(null, null) as ActiveUserToken;
+
+            bool refreshed = activeUserCred.RefreshTokenAsync(CancelToken).Result;
+
+            Assert.IsTrue(refreshed, "RefreshTokenAsync should return true.");
+
+            ActiveUserToken refreshedActiveUserToken = TokenProperty.GetValue(null, null) as ActiveUserToken;
+
+            Assert.IsTrue(
+                !Equals(activeUserToken.AccessToken, refreshedActiveUserToken.AccessToken),
+                "Refreshed ActiveUserToken should have a different access token.");
+        }
+
+        /// <summary>
+        /// This test checks that GetAccessTokenForRequestAsync works.
+        /// </summary>
+        [Test]
+        public void TestGetAccessTokenForRequestAsync()
+        {
+            ActiveUserCredential activeUserCred = new ActiveUserCredential();
+            ActiveUserToken activeUserToken = TokenProperty.GetValue(null, null) as ActiveUserToken;
+
+            string accessToken = activeUserCred.GetAccessTokenForRequestAsync(null, CancelToken).Result;
+
+            // The access token returned by GetAccessTokenForRequestAsync should come from token.
+            Assert.IsTrue(
+                Equals(activeUserToken.AccessToken, accessToken),
+                "GetAccessTokenForRefreshAsync returns the wrong access token.");
+
+            // The next call to GetAccessTokenForRequestAsync should returns the same token.
+            accessToken = activeUserCred.GetAccessTokenForRequestAsync(null, CancelToken).Result;
+            Assert.IsTrue(
+                Equals(activeUserToken.AccessToken, accessToken),
+                "GetAccessTokenForRefreshAsync returns the wrong access token.");
+        }
+
+        /// <summary>
+        /// This test checks that GetAccessTokenForRequestAsync returns a new token
+        /// when the old one expired due to time.
+        /// </summary>
+        [Test]
+        public void TestGetAccessTokenExpiredByTime()
+        {
+            ActiveUserCredential activeUserCred = new ActiveUserCredential();
+            ActiveUserToken activeUserToken = TokenProperty.GetValue(null, null) as ActiveUserToken;
+
+            // Make the token expired by time.
+            activeUserToken.Issued = DateTime.Now.AddSeconds(-activeUserToken.ExpiresInSeconds.Value);
+            Assert.IsTrue(activeUserToken.IsExpired(SystemClock.Default), "ActiveUserToken should be expired");
+
+            string newAccessToken = activeUserCred.GetAccessTokenForRequestAsync(null, CancelToken).Result;
+            Assert.IsFalse(
+                Equals(activeUserToken.AccessToken, newAccessToken),
+                "GetAccessTokenForRefreshAsync should returns a new access token when the old one expired due to time."
+                );
+
+            // The next call to GetAccessTokenForRequestAsync should returns the same token.
+            string anotherNewAccessToken = activeUserCred.GetAccessTokenForRequestAsync(null, CancelToken).Result;
+            Assert.IsTrue(
+                Equals(anotherNewAccessToken, newAccessToken),
+                "GetAccessTokenForRefreshAsync should returns a new access token when the old one expired due to time."
+                );
+        }
+
+        /// <summary>
+        /// This test checks that GetAccessTokenForRequestAsync returns a new token
+        /// when the active user is changed.
+        /// </summary>
+        [Test]
+        public void TestGetAcessTokenExpiredByChangingUser()
+        {
+            ActiveUserCredential activeUserCred = new ActiveUserCredential();
+            ActiveUserToken activeUserToken = TokenProperty.GetValue(null, null) as ActiveUserToken;
+
+            // Make the token expired by changing the active user
+            FieldInfo activeUserField = typeof(ActiveUserToken).GetField(
+                "activeUser",
+                BindingFlags.NonPublic | BindingFlags.Instance);
+            activeUserField.SetValue(activeUserToken, "A new user");
+
+            Assert.IsTrue(activeUserToken.IsExpired(SystemClock.Default), "ActiveUserToken should be expired");
+
+            string newAccessToken = activeUserCred.GetAccessTokenForRequestAsync(null, CancelToken).Result;
+            Assert.IsFalse(
+                Equals(activeUserToken.AccessToken, newAccessToken),
+                "GetAccessTokenForRefreshAsync should returns a new access token when the old one expired due to time."
+                );
+
+            // The next call to GetAccessTokenForRequestAsync should returns the same token.
+            string anotherNewAccessToken = activeUserCred.GetAccessTokenForRequestAsync(null, CancelToken).Result;
+            Assert.IsTrue(
+                Equals(anotherNewAccessToken, newAccessToken),
+                "GetAccessTokenForRefreshAsync should returns a new access token when the old one expired due to time."
+                );
+        }
+    }
+}
+
--- a/Google.PowerShell.Tests/Common/TestActiveUserToken.cs
+++ b/Google.PowerShell.Tests/Common/TestActiveUserToken.cs
@@ -12,31 +12,31 @@ namespace Google.PowerShell.Tests.Common
     public class TestActiveUserToken
     {
         private string credentialJson = @"{
-            ""access_token"": ""access token""
-        }";
+            'access_token': 'access token'
+        }".Replace('\'', '\"');
 
         private string credentialJsonWithNullExpiry = @"{
-            ""access_token"": ""access token"",
-            ""token_expiry"": null
-        }";
+            'access_token': 'access token',
+            'token_expiry': null
+        }".Replace('\'', '\"');
 
         private string credentialJsonWithExpiry = @"{
-            ""access_token"": ""access token"",
-            ""token_expiry"": {
-                ""year"": 2016,
-                ""month"": 10,
-                ""day"": 18,
-                ""hour"": 20,
-                ""minute"": 36,
-                ""second"": 4,
-                ""microsecond"": 497000
+            'access_token': 'access token',
+            'token_expiry': {
+                'year': 2016,
+                'month': 10,
+                'day': 18,
+                'hour': 20,
+                'minute': 36,
+                'second': 4,
+                'microsecond': 497000
             }
-        }";
+        }".Replace('\'', '\"');
 
         [Test]
         public void TestCredentialJsonWithTokenExpiry()
         {
-            ActiveUserToken token = new ActiveUserToken(credentialJsonWithExpiry, () => { return "user"; });
+            ActiveUserToken token = new ActiveUserToken(credentialJsonWithExpiry, "user");
             Assert.IsTrue(Equals(token.AccessToken, "access token"), "AccessToken does not match the value in JSON string.");
             Assert.IsTrue(
                 Equals(token.ExpiredTime, new DateTime(2016, 10, 18, 20, 36, 4, 497, DateTimeKind.Utc)),
@@ -46,27 +46,10 @@ namespace Google.PowerShell.Tests.Common
         [Test]
         public void TestCredentialJsonWithNullTokenExpiry()
         {
-            ActiveUserToken token = new ActiveUserToken(credentialJsonWithNullExpiry, () => { return "user"; });
+            ActiveUserToken token = new ActiveUserToken(credentialJsonWithNullExpiry, "user");
             Assert.IsTrue(
                 Equals(token.ExpiredTime, DateTime.MaxValue),
                 "ExpiredTime should be set to DateTime.MaxValue if token_expiry is null");
-        }
-
-        /// <summary>
-        /// This test checks that if the active user is changed, the token should be considered as expired.
-        /// </summary>
-        [Test]
-        public void TestTokenExpiredByActiveUser()
-        {
-            ActiveUserToken token = new ActiveUserToken(credentialJson, () => { return "user"; });
-            token.ExpiredTime = DateTime.UtcNow.AddHours(1);
-
-            Assert.IsFalse(token.IsExpiredOrInvalid());
-
-            token.getActiveUser = () => { return "differentUser"; };
-            Assert.IsTrue(
-                token.IsExpiredOrInvalid(),
-                "Token should have expired when the active user was changed,");
         }
 
         /// <summary>
@@ -75,15 +58,15 @@ namespace Google.PowerShell.Tests.Common
         [Test]
         public void TestTokenExpiredByTime()
         {
-            ActiveUserToken token = new ActiveUserToken(credentialJson, () => { return "user"; });
+            ActiveUserToken token = new ActiveUserToken(credentialJson, "user");
             // If we don't set the access token, the token will be treated as expired.
             token.AccessToken = "Random Access Token";
             token.ExpiredTime = DateTime.UtcNow.AddHours(1);
 
-            Assert.IsFalse(token.IsExpiredOrInvalid(), "Token should not expire yet.");
+            Assert.IsFalse(token.IsExpired, "Token should not expire yet.");
 
             token.ExpiredTime = DateTime.UtcNow.AddMinutes(-1);
-            Assert.IsTrue(token.IsExpiredOrInvalid(), "Token should have expired 1 minute before.");
+            Assert.IsTrue(token.IsExpired, "Token should have expired 1 minute before.");
         }
     }
 }
--- a/Google.PowerShell.Tests/Common/TestActiveUserToken.cs
+++ b/Google.PowerShell.Tests/Common/TestActiveUserToken.cs
@@ -1,8 +1,6 @@
 ﻿using NUnit.Framework;
-using Google.Apis.Util;
 using Google.PowerShell.Common;
 using System;
-using System.Reflection;
 
 namespace Google.PowerShell.Tests.Common
 {
@@ -11,30 +9,63 @@ namespace Google.PowerShell.Tests.Common
     /// Assumes that the Cloud SDK is installed on the local machine and initialized.
     /// 
     /// </summary>
-    internal class TestActiveUserToken
+    public class TestActiveUserToken
     {
+        private string credentialJson = @"{
+            ""access_token"": ""access token""
+        }";
+
+        private string credentialJsonWithNullExpiry = @"{
+            ""access_token"": ""access token"",
+            ""token_expiry"": null
+        }";
+
+        private string credentialJsonWithExpiry = @"{
+            ""access_token"": ""access token"",
+            ""token_expiry"": {
+                ""year"": 2016,
+                ""month"": 10,
+                ""day"": 18,
+                ""hour"": 20,
+                ""minute"": 36,
+                ""second"": 4,
+                ""microsecond"": 497000
+            }
+        }";
+
+        [Test]
+        public void TestCredentialJsonWithTokenExpiry()
+        {
+            ActiveUserToken token = new ActiveUserToken(credentialJsonWithExpiry, () => { return "user"; });
+            Assert.IsTrue(Equals(token.AccessToken, "access token"), "AccessToken does not match the value in JSON string.");
+            Assert.IsTrue(
+                Equals(token.ExpiredTime, new DateTime(2016, 10, 18, 20, 36, 4, 497, DateTimeKind.Utc)),
+                "ExpiredTime does not match the value in JSON string.");
+        }
+
+        [Test]
+        public void TestCredentialJsonWithNullTokenExpiry()
+        {
+            ActiveUserToken token = new ActiveUserToken(credentialJsonWithNullExpiry, () => { return "user"; });
+            Assert.IsTrue(
+                Equals(token.ExpiredTime, DateTime.MaxValue),
+                "ExpiredTime should be set to DateTime.MaxValue if token_expiry is null");
+        }
+
         /// <summary>
         /// This test checks that if the active user is changed, the token should be considered as expired.
         /// </summary>
         [Test]
         public void TestTokenExpiredByActiveUser()
         {
-            ActiveUserToken token = new ActiveUserToken();
-            // If we don't set the access token, the token will be treated as expired.
-            token.AccessToken = "Random Access Token";
-            token.ExpiresInSeconds = 3600;
-            token.Issued = DateTime.Now;
+            ActiveUserToken token = new ActiveUserToken(credentialJson, () => { return "user"; });
+            token.ExpiredTime = DateTime.UtcNow.AddHours(1);
 
-            Assert.IsFalse(token.IsExpired(SystemClock.Default));
+            Assert.IsFalse(token.IsExpiredOrInvalid());
 
-            FieldInfo activeUserField = typeof(ActiveUserToken).GetField(
-                "activeUser",
-                BindingFlags.NonPublic | BindingFlags.Instance);
-
-            activeUserField.SetValue(token, "A new user");
-
+            token.getActiveUser = () => { return "differentUser"; };
             Assert.IsTrue(
-                token.IsExpired(SystemClock.Default),
+                token.IsExpiredOrInvalid(),
                 "Token should have expired when the active user was changed,");
         }
 
@@ -44,16 +75,15 @@ namespace Google.PowerShell.Tests.Common
         [Test]
         public void TestTokenExpiredByTime()
         {
-            ActiveUserToken token = new ActiveUserToken();
+            ActiveUserToken token = new ActiveUserToken(credentialJson, () => { return "user"; });
             // If we don't set the access token, the token will be treated as expired.
             token.AccessToken = "Random Access Token";
-            token.ExpiresInSeconds = 3600;
+            token.ExpiredTime = DateTime.UtcNow.AddHours(1);
 
-            token.Issued = DateTime.Now.AddSeconds(-1800);
-            Assert.IsFalse(token.IsExpired(SystemClock.Default), "Token should not expire yet.");
+            Assert.IsFalse(token.IsExpiredOrInvalid(), "Token should not expire yet.");
 
-            token.Issued = DateTime.Now.AddSeconds(-3591);
-            Assert.IsTrue(token.IsExpired(SystemClock.Default), "Token should have expired 1 minute before.");
+            token.ExpiredTime = DateTime.UtcNow.AddMinutes(-1);
+            Assert.IsTrue(token.IsExpiredOrInvalid(), "Token should have expired 1 minute before.");
         }
     }
 }
--- a/Google.PowerShell.Tests/Common/TestActiveUserToken.cs
+++ b/Google.PowerShell.Tests/Common/TestActiveUserToken.cs
@@ -35,7 +35,7 @@ namespace Google.PowerShell.Tests.Common
 
             Assert.IsTrue(
                 token.IsExpired(SystemClock.Default),
-                "Token should expire when the active user is changed,");
+                "Token should have expired when the active user was changed,");
         }
 
         /// <summary>
@@ -53,7 +53,7 @@ namespace Google.PowerShell.Tests.Common
             Assert.IsFalse(token.IsExpired(SystemClock.Default), "Token should not expire yet.");
 
             token.Issued = DateTime.Now.AddSeconds(-3591);
-            Assert.IsTrue(token.IsExpired(SystemClock.Default), "Token should expire 1 minute before.");
+            Assert.IsTrue(token.IsExpired(SystemClock.Default), "Token should have expired 1 minute before.");
         }
     }
 }
--- a/Google.PowerShell.Tests/Common/TestActiveUserToken.cs
+++ b/Google.PowerShell.Tests/Common/TestActiveUserToken.cs
@@ -1,0 +1,60 @@
+﻿using NUnit.Framework;
+using Google.Apis.Util;
+using Google.PowerShell.Common;
+using System;
+using System.Reflection;
+
+namespace Google.PowerShell.Tests.Common
+{
+    [TestFixture]
+    /// <summary>
+    /// Assumes that the Cloud SDK is installed on the local machine and initialized.
+    /// 
+    /// </summary>
+    internal class TestActiveUserToken
+    {
+        /// <summary>
+        /// This test checks that if the active user is changed, the token should be considered as expired.
+        /// </summary>
+        [Test]
+        public void TestTokenExpiredByActiveUser()
+        {
+            ActiveUserToken token = new ActiveUserToken();
+            // If we don't set the access token, the token will be treated as expired.
+            token.AccessToken = "Random Access Token";
+            token.ExpiresInSeconds = 3600;
+            token.Issued = DateTime.Now;
+
+            Assert.IsFalse(token.IsExpired(SystemClock.Default));
+
+            FieldInfo activeUserField = typeof(ActiveUserToken).GetField(
+                "activeUser",
+                BindingFlags.NonPublic | BindingFlags.Instance);
+
+            activeUserField.SetValue(token, "A new user");
+
+            Assert.IsTrue(
+                token.IsExpired(SystemClock.Default),
+                "Token should expire when the active user is changed,");
+        }
+
+        /// <summary>
+        /// This test checks that if enough time has passed, the token should be considered as expired.
+        /// </summary>
+        [Test]
+        public void TestTokenExpiredByTime()
+        {
+            ActiveUserToken token = new ActiveUserToken();
+            // If we don't set the access token, the token will be treated as expired.
+            token.AccessToken = "Random Access Token";
+            token.ExpiresInSeconds = 3600;
+
+            token.Issued = DateTime.Now.AddSeconds(-1800);
+            Assert.IsFalse(token.IsExpired(SystemClock.Default), "Token should not expire yet.");
+
+            token.Issued = DateTime.Now.AddSeconds(-3591);
+            Assert.IsTrue(token.IsExpired(SystemClock.Default), "Token should expire 1 minute before.");
+        }
+    }
+}
+
--- a/Google.PowerShell.Tests/Common/TestActiveUserToken.cs
+++ b/Google.PowerShell.Tests/Common/TestActiveUserToken.cs
@@ -59,8 +59,6 @@ namespace Google.PowerShell.Tests.Common
         public void TestTokenExpiredByTime()
         {
             ActiveUserToken token = new ActiveUserToken(credentialJson, "user");
-            // If we don't set the access token, the token will be treated as expired.
-            token.AccessToken = "Random Access Token";
             token.ExpiredTime = DateTime.UtcNow.AddHours(1);
 
             Assert.IsFalse(token.IsExpired, "Token should not expire yet.");
--- a/Google.PowerShell/Common/ActiveUserCredential.cs
+++ b/Google.PowerShell/Common/ActiveUserCredential.cs
@@ -1,0 +1,184 @@
+﻿/*
+Copyright 2016 Google Inc
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+using System;
+using System.Net;
+using System.Net.Http;
+using System.Threading;
+using System.Threading.Tasks;
+
+using Google.Apis.Auth.OAuth2.Flows;
+using Google.Apis.Auth.OAuth2.Responses;
+using Google.Apis.Http;
+using Google.Apis.Logging;
+using Google.PowerShell.Common;
+
+namespace Google.Apis.Auth.OAuth2
+{
+    /// <summary>
+    /// OAuth 2.0 credential for accessing protected resources using an access token.
+    /// This class will get the access token from "gcloud auth print-access-token".
+    /// </summary>
+    public class ActiveUserCredential : ICredential, IHttpExecuteInterceptor, IHttpUnsuccessfulResponseHandler
+    {
+        private static ActiveUserToken token;
+        private static object lockObject = new object();
+
+        /// <summary>
+        /// We just need a default flow for things like flow.AccessMethod.Intercept
+        /// or flow.RevokeTokenAsync.
+        /// </summary>
+        private readonly IAuthorizationCodeFlow flow = new GoogleAuthorizationCodeFlow(
+                new GoogleAuthorizationCodeFlow.Initializer()
+                {
+                    ClientSecrets = new ClientSecrets()
+                    {
+                        ClientId = "clientId",
+                        ClientSecret = "clientSecrets"
+                    }
+                });
+
+        /// <summary>
+        /// UserId is not important, it can be anything.
+        /// </summary>
+        private readonly string userId = "userId";
+
+        /// <summary>Gets or sets the token response which contains the access token.</summary>
+        public static ActiveUserToken Token
+        {
+            get
+            {
+                lock (lockObject)
+                {
+                    return token;
+                }
+            }
+            set
+            {
+                lock (lockObject)
+                {
+                    token = value;
+                }
+            }
+        }
+
+        /// <summary>Gets the authorization code flow.</summary>
+        public IAuthorizationCodeFlow Flow
+        {
+            get { return flow; }
+        }
+
+        /// <summary>Constructs a new credential instance.</summary>
+        public ActiveUserCredential()
+        {
+            if (Token == null)
+            {
+                Token = GCloudWrapper.GetAccessToken().Result;
+            }
+        }
+
+        #region IHttpExecuteInterceptor
+
+        /// <summary>
+        /// Default implementation is to try to refresh the access token if there is no access token or if we are 1 
+        /// minute away from expiration. If token server is unavailable, it will try to use the access token even if 
+        /// has expired. If successful, it will call <see cref="IAccessMethod.Intercept"/>.
+        /// </summary>
+        public async Task InterceptAsync(HttpRequestMessage request, CancellationToken taskCancellationToken)
+        {
+            var accessToken = await GetAccessTokenForRequestAsync(request.RequestUri.ToString(), taskCancellationToken).ConfigureAwait(false);
+            flow.AccessMethod.Intercept(request, Token.AccessToken);
+        }
+
+        #endregion
+
+        #region IHttpUnsuccessfulResponseHandler
+
+        public async Task<bool> HandleResponseAsync(HandleUnsuccessfulResponseArgs args)
+        {
+            if (args.Response.StatusCode == HttpStatusCode.Unauthorized)
+            {
+                return !Object.Equals(Token.AccessToken, flow.AccessMethod.GetAccessToken(args.Request))
+                    || await RefreshTokenAsync(args.CancellationToken).ConfigureAwait(false);
+            }
+
+            return false;
+        }
+
+        #endregion
+
+        #region IConfigurableHttpClientInitializer
+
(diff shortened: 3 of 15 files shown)


Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
```diff
@@ -0,0 +1,138 @@
...
+                BindingFlags.NonPublic | BindingFlags.Static);
+        private CancellationToken CancelToken = new CancellationToken();
+
+        /// <summary>
+        /// This test checks that an ActiveUserToken is created when a new
+        /// ActiveUserCredential is created.
+        /// </summary>
+        [Test]
+        public void TestActiveUserCredentialConstructor()
+        {
```
chrsmith: Do we want to explicitly null it out before the test?

Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
```diff
@@ -0,0 +1,138 @@
...
+namespace Google.PowerShell.Tests.Common
+{
+    [TestFixture]
+    /// <summary>
+    /// Assumes that the Cloud SDK is installed on the local machine and initialized.
+    /// 
+    /// </summary>
+    internal class TestActiveUserCredential
+    {
+        private PropertyInfo TokenProperty = typeof(ActiveUserCredential).GetProperty(
```
chrsmith: To be consistent with the .NET naming guidelines, this should be `tokenProperty` (lowercase T). Same with `cancelToken` below.

Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// This test checks that GetAccessTokenForRequestAsync returns a new token
+        /// when the old one expired due to time.
+        /// </summary>
+        [Test]
+        public void TestGetAccessTokenExpiredByTime()
+        {
+            ActiveUserCredential activeUserCred = new ActiveUserCredential();
+            ActiveUserToken activeUserToken = TokenProperty.GetValue(null, null) as ActiveUserToken;
+
+            // Make the token expired by time.
```
chrsmith: "expired by time" isn't especially cromulent. Perhaps "Force the access token to expire."

Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
```diff
@@ -0,0 +1,138 @@
...
+            activeUserToken.Issued = DateTime.Now.AddSeconds(-activeUserToken.ExpiresInSeconds.Value);
+            Assert.IsTrue(activeUserToken.IsExpired(SystemClock.Default), "ActiveUserToken should be expired");
+
+            string newAccessToken = activeUserCred.GetAccessTokenForRequestAsync(null, CancelToken).Result;
+            Assert.IsFalse(
+                Equals(activeUserToken.AccessToken, newAccessToken),
+                "GetAccessTokenForRefreshAsync should returns a new access token when the old one expired due to time."
+                );
```
chrsmith: When spilling function parameters across multiple lines, put the final `);` on the same line as the last parameter.

Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// This test checks that GetAccessTokenForRequestAsync returns a new token
+        /// when the active user is changed.
+        /// </summary>
+        [Test]
+        public void TestGetAcessTokenExpiredByChangingUser()
+        {
+            ActiveUserCredential activeUserCred = new ActiveUserCredential();
+            ActiveUserToken activeUserToken = TokenProperty.GetValue(null, null) as ActiveUserToken;
+
+            // Make the token expired by changing the active user
```
chrsmith: This comment and the error message below (line 121) aren't 100% correct. It isn't that the token has expired (ran out of time). It is simply no longer valid. By changing the name, the hash/whatever it uses to validate that token is no longer correct. Perhaps we sould just say "Invalidate the token by changing the active user."?

Google.PowerShell.Tests/Common/TestActiveUserToken.cs
```diff
@@ -0,0 +1,60 @@
...
+            ActiveUserToken token = new ActiveUserToken();
+            // If we don't set the access token, the token will be treated as expired.
+            token.AccessToken = "Random Access Token";
+            token.ExpiresInSeconds = 3600;
+
+            token.Issued = DateTime.Now.AddSeconds(-1800);
+            Assert.IsFalse(token.IsExpired(SystemClock.Default), "Token should not expire yet.");
+
+            token.Issued = DateTime.Now.AddSeconds(-3591);
+            Assert.IsTrue(token.IsExpired(SystemClock.Default), "Token should expire 1 minute before.");
```
chrsmith: "should have expired"

Google.PowerShell/Common/ActiveUserCredential.cs
```diff
@@ -0,0 +1,172 @@
...
+namespace Google.PowerShell.Common
+{
+    /// <summary>
+    /// OAuth 2.0 credential for accessing protected resources using an access token.
+    /// This class will get the access token from "gcloud auth print-access-token".
+    /// </summary>
+    public class ActiveUserCredential : ICredential, IHttpExecuteInterceptor, IHttpUnsuccessfulResponseHandler
+    {
+        private static ActiveUserToken token;
+        private static object lockObject = new object();
```
chrsmith: Nit: Use `new Object()` instead of the lower-case `object` keyword.

Google.PowerShell/Common/ActiveUserCredential.cs
```diff
@@ -0,0 +1,172 @@
...
+
+namespace Google.PowerShell.Common
+{
+    /// <summary>
+    /// OAuth 2.0 credential for accessing protected resources using an access token.
+    /// This class will get the access token from "gcloud auth print-access-token".
+    /// </summary>
+    public class ActiveUserCredential : ICredential, IHttpExecuteInterceptor, IHttpUnsuccessfulResponseHandler
+    {
+        private static ActiveUserToken token;
```
chrsmith: The .NET naming guidelines for static fields (which everyone unanimously hates) dictates this be called `s_token` and `s_lockObject`. Sorry.

Also, please add a detailed comment about what `token` is, how it is used, and what invariants the `ActiveUserCredential` class expects. e.g. "This is a cache of the last call to `gcloud auth print-access-token`. If the current user hasn't changed, and the token hasn't expired, we reuse it rather than launching gcloud again."

Google.PowerShell/Common/ActiveUserCredential.cs
```diff
@@ -0,0 +1,172 @@
...
+        private static object lockObject = new object();
+
+        /// <summary>
+        /// We just need a default flow for things like flow.AccessMethod.Intercept
+        /// or flow.RevokeTokenAsync.
+        /// </summary>
+        private readonly IAuthorizationCodeFlow flow = new GoogleAuthorizationCodeFlow(
```
chrsmith: Delete this private field? See comment below on the property.

Google.PowerShell/Common/ActiveUserCredential.cs
```diff
@@ -0,0 +1,172 @@
...
+                {
+                    ClientSecrets = new ClientSecrets()
+                    {
+                        ClientId = "clientId",
+                        ClientSecret = "clientSecrets"
+                    }
+                });
+
+        /// <summary>UserId is not important, it can be anything.</summary>
+        private readonly string userId = "userId";
```
chrsmith: Given that this field is only accessed in one method, why not make it a local variable in the `RevokeTokenAsync` method? i.e. do we really need it as a class field?

Google.PowerShell/Common/ActiveUserCredential.cs
```diff
@@ -0,0 +1,172 @@
...
+                {
+                    token = value;
+                }
+            }
+        }
+
+        /// <summary>Gets the authorization code flow.</summary>
+        public IAuthorizationCodeFlow Flow
+        {
+            get { return flow; }
```
chrsmith: Can we avoid the need for a local variable flow by just using a default value on an autoprop?

````
public IAuthorizationCodeFlow Flow { get; } = new GoogleAuthorizationCodeFlow(
               new GoogleAuthorizationCodeFlow.Initializer()
               {
                    ClientSecrets = new ClientSecrets()
                    {
                        ClientId = "clientId",
                        ClientSecret = "clientSecrets"
                    }
                });
````

Comment (chrsmith) [inline Google.PowerShell/Common/ActiveUserCredential.cs]: However, I don't understand how this is actually used. The comment "Gets the authorization code flow." states what the property is, but not why it needs to exist.

What is it used for? Who needs it? Is it just this class, or does it actually need to be public and used outside of this class?

File: Google.PowerShell/Common/ActiveUserCredential.cs [hunk 12]
```diff
@@ -0,0 +1,172 @@
...
+        private static ActiveUserToken Token
+        {
+            get
+            {
+                lock (lockObject)
+                {
+                    return token;
+                }
+            }
+            set
```
chrsmith: Does the setter need to be public? Why not protected? Can we use the InteralsVisibleTo attribute to enable unit testing, but not harm the design?

Google.PowerShell/Common/ActiveUserCredential.cs
```diff
@@ -0,0 +1,172 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace Google.PowerShell.Common
+{
+    /// <summary>
+    /// OAuth 2.0 credential for accessing protected resources using an access token.
+    /// This class will get the access token from "gcloud auth print-access-token".
+    /// </summary>
+    public class ActiveUserCredential : ICredential, IHttpExecuteInterceptor, IHttpUnsuccessfulResponseHandler
```
chrsmith: The type name isn't especially useful. `ActiveUserCredential` makes me think it is a data type. However, we  use this class as a substitute for how Apiary authenticates RPC calls. Perhaps we should call this "AuthicateWithSdkCredentialsExecutor"? That name is obviously terrible, but does communicate more of how it is supposed to be used.

Google.PowerShell/Common/ActiveUserCredential.cs
```diff
@@ -0,0 +1,172 @@
...
+        {
+            get { return flow; }
+        }
+
+        /// <summary>Constructs a new credential instance.</summary>
+        public ActiveUserCredential()
+        {
+            if (Token == null)
+            {
+                Token = GCloudWrapper.GetAccessToken().Result;
```
chrsmith: The problem with this is that it blocks. Perhaps we should just leave this as a TODO for now, but I'd imagine having `Token` only provide a getter, and it simply returns `tokenRequestTask`. (`Task<ActiveUserToken> tokenRequestTask`.)

When you access Token, it would call `tokenRequestTask.Result`, meaning that we don't block in the construction of `ActiveUserCredential` until the credential is returned.

Just a thought.

Google.PowerShell/Common/ActiveUserCredential.cs
```diff
@@ -0,0 +1,172 @@
...
+        #region IHttpExecuteInterceptor
+
+        /// <summary>
+        /// Default implementation is to try to refresh the access token if there is no access token or if we are 1 
+        /// minute away from expiration. If token server is unavailable, it will try to use the access token even if 
+        /// has expired. If successful, it will call <see cref="IAccessMethod.Intercept"/>.
+        /// </summary>
+        public async Task InterceptAsync(HttpRequestMessage request, CancellationToken taskCancellationToken)
+        {
+            var accessToken = await GetAccessTokenForRequestAsync(request.RequestUri.ToString(), taskCancellationToken).ConfigureAwait(false);
```
chrsmith: Please introduce some local variables. It is unclear what type `accessToken` has. Perhaps:

```
Foo x = GetAccessTokenForRequestAsync(request.RequestUri.ToString(), taskCancellationToken);
Bar accessToken = x.ConfigureAwait(false);
````

What do you think?

File: Google.PowerShell/Common/ActiveUserCredential.cs [hunk 16]
```diff
@@ -0,0 +1,172 @@
...
+            {
+                return false;
+            }
+
+            await flow.RevokeTokenAsync(userId, Token.AccessToken, taskCancellationToken).ConfigureAwait(false);
+            // We don't set the token to null, cause we want that the next request (without reauthorizing) will fail).
+            return true;
+        }
+    }
+}
```
chrsmith: New line at end of file.

Google.PowerShell/Common/ActiveUserToken.cs
```diff
@@ -0,0 +1,74 @@
+﻿using Google.Apis.Util;
+using System;
+
+namespace Google.PowerShell.Common
+{
+    /// <summary>
+    /// OAuth 2.0 model for a successful access token response as specified in 
+    /// http://tools.ietf.org/html/rfc6749#section-5.1.
+    /// </summary>
+    public class ActiveUserToken
```
chrsmith: Can we make this class immutable? e.g. converting all properties into `{ get; protected set; }`. The idea of modifying the fields of this seems super sketchy. Perhaps we should just have a constructor that parses the values out of a Json object?

e.g. `new ActiveUserToken(userTokenJson)`?

Google.PowerShell/Common/ActiveUserToken.cs
```diff
@@ -0,0 +1,74 @@
...
+    {
+        /// <summary>
+        /// The current active user that corresponds to this token.
+        /// </summary>
+        private string activeUser = CloudSdkSettings.GetSettingsValue("account");
```
chrsmith: Reading from the Cloud SDK settings here, and again in `IsExpired` too tightly couples the Cloud SDK's implementation and what would otherwise be a straight-forward data object.

Could we instead have `ActuiveUserToken` take an `Action<string>` object (delegate) to compute the current user name on demand? That would make it possible to test `ActiveUserToken` logic without actually accessing Cloud SDK configuration files on disk.

As for complicating usage, IMHO it isn't too bad. And you can always include a static factory method for the common case.

e.g. 
````
public static ActiveUserToken NewCloudSdkUserToken() {
    return new ActiveUserToken(() => CloudSdkSettings.GetSettingsValue("account"));
}
````

File: Google.PowerShell/Common/CloudSdkSettings.cs [hunk 19]
```diff
@@ -21,6 +21,18 @@ public class CommonProperties
             public const string Region = "region";
         }
 
+        /// <summary>
+        /// Environment variable which points to the location of the current configuration file.
+        /// This overrides the user configuration file as well as the globa lcinstallation properties file.
```
chrsmith: "global installation properties"

Google.PowerShell/Common/CloudSdkSettings.cs
```diff
@@ -21,6 +21,18 @@ public class CommonProperties
...
+        /// </summary>
+        private const string CloudSdkConfigVariable = "CLOUDSDK_CONFIG";
+
+        /// <summary>
+        /// Environment variable which stores the current active config.
+        /// This overrides value found in active_config file.
```
chrsmith: Environment variable which stores the current active config. Overrides the value found in the active_config file *if present*.

Google.PowerShell/Common/CloudSdkSettings.cs
```diff
@@ -36,25 +48,47 @@ public class CommonProperties
         /// <summary>Name of the file containing the anonymous client ID.</summary>
         private const string ClientIDFileName = ".metricsUUID";
 
+        private static string _installationPropertiesPath;
```
chrsmith: s_installationPropertiesPath

Google.PowerShell/Common/CloudSdkSettings.cs
```diff
@@ -67,21 +101,45 @@ public static string GetCurrentConfigurationName()
             }
         }
 
+        /// <summary>
+        /// Returns the current config directory.
+        /// If CLOUDSDK_CONFIG is set, we will use that as the config directory.
```
chrsmith: "If the CLOUDSDK_CONFIG *environment variable* is set..."

Google.PowerShell/Common/CloudSdkSettings.cs
```diff
@@ -67,21 +101,45 @@ public static string GetCurrentConfigurationName()
...
+        /// If not, we will use the config directory from AppData.
+        /// </summary>
+        public static string GetCurrentConfigurationDirectory()
+        {
+            string cloudConfigPath = Environment.GetEnvironmentVariable(CloudSdkConfigVariable);
+
```
chrsmith: Delete this line.

Google.PowerShell/Common/CloudSdkSettings.cs
```diff
@@ -67,21 +101,45 @@ public static string GetCurrentConfigurationName()
...
         /// <summary> 
         /// Returns the file path to the current Cloud SDK configuration set's property file. Returns null on any
         /// sort of error.
         /// </summary>
         public static string GetCurrentConfigurationFilePath()
         {
-            string appDataFolder = Environment.GetEnvironmentVariable(AppdataEnvironmentVariable);
-            if (appDataFolder == null || !Directory.Exists(appDataFolder))
+            string cloudConfigDir = GetCurrentConfigurationDirectory();
+
```
chrsmith: Delete this line.

Google.PowerShell/Common/CloudSdkSettings.cs
```diff
@@ -94,10 +152,27 @@ public static string GetCurrentConfigurationFilePath()
...
+                userConfigSetting = GetSettingsValueFromFile(settingName, InstallationPropertiesPath);
+            }
+
+            return userConfigSetting;
+        }
+
+        /// <summary>
+        /// Retrieves the setting with the given name from a config file.
+        /// </summary>
+        private static string GetSettingsValueFromFile(string settingName, string configFile)
```
chrsmith: This is a style nit, but personally I assume function parameters go from most important to least important. So putting the file name before the setting name seems more intuitive to me. But YMMV.

Google.PowerShell/Common/GcloudWrapper.cs
```diff
@@ -0,0 +1,109 @@
...
+
+            if (!string.IsNullOrWhiteSpace(processOutput.StandardError))
+            {
+                throw new InvalidOperationException($"Command {actualCommand} failed with error: processOutput.StandardError");
+            }
+
+            throw new InvalidOperationException($"Command {actualCommand} failed.");
+        }
+    }
+}
```
chrsmith: Newline at end of file.

Google.PowerShell/Common/GcloudWrapper.cs
```diff
@@ -0,0 +1,109 @@
...
+using System;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Google.PowerShell.Common
+{
+    /// <summary>
+    /// This class wraps the gcloud command and offers up some of its services.
```
chrsmith: "wraps the gcloud command" isn't especially specific. How about:

"This class shell executes "gcloud {command} --format=json", to allow delegation to Cloud SDK implementation. Used for things like credential management."

Google.PowerShell/Common/GcloudWrapper.cs
```diff
@@ -0,0 +1,109 @@
...
+                gCloudInfoJson = gCloudInfoJson.SelectToken("config.paths.installation_properties_path");
+                
+                if (gCloudInfoJson.Type == JTokenType.String)
+                {
+                    return gCloudInfoJson.Value<string>();
+                }
+            }
+            catch (NullReferenceException)
+            {
+                // Throw exception at the end.
```
chrsmith: Why not:

````
catch (Exception ex) {
    throw new FileNotFoundException("Installation Properties file for Google Cloud SDK cannot be found.", ex);
}
````
?

File: Google.PowerShell/Common/GcloudWrapper.cs [hunk 29]
```diff
@@ -0,0 +1,109 @@
...
+        /// </summary>
+        /// <returns></returns>
+        public static async Task<string> GetInstallationPropertiesPath()
+        {
+            string gCloudInfoOutput = await GetGCloudCommandOutput("info");
+            JToken gCloudInfoJson = JObject.Parse(gCloudInfoOutput);
+
+            try
+            {
+                // SelectToken will throw NullReferenceException if the token cannot be found.
```
chrsmith: Is there a way to check if the token exists? Ultimately we should avoid having the exception be thrown if we can. There is a huge perf hit and we can also provide a more accurate error message. (e.g. "Got configuration value, but couldn't find installation_properties_path."

Google.PowerShell/Common/GcloudWrapper.cs
```diff
@@ -0,0 +1,109 @@
...
+{
+    /// <summary>
+    /// This class wraps the gcloud command and offers up some of its services.
+    /// </summary>
+    public static class GCloudWrapper
+    {
+        /// <summary>
+        /// Returns the global installation properties path of GoogleCloud SDK.
+        /// </summary>
+        /// <returns></returns>
```
chrsmith: Delete this line.

Google.PowerShell/Common/GcloudWrapper.cs
```diff
@@ -0,0 +1,109 @@
...
+        public static async Task<string> GetInstallationPropertiesPath()
```
chrsmith: How often do we call this? Don't we need to call this nearly every time we look up a Cloud SDK setting? It might be worth-while to wrap/cache the result?

Google.PowerShell/Common/GcloudWrapper.cs
```diff
@@ -0,0 +1,109 @@
...
+                tokenJson = tokenJson.SelectToken("token_response");
+
+                ActiveUserToken token = tokenJson.ToObject<ActiveUserToken>();
+                token.Issued = issuedTime;
+
+                return token;
+            }
+            catch (NullReferenceException)
+            {
+                // Throw exception at the end.
```
chrsmith: Same as before. Let's provide better exception messages if the value we need isn't found. It will make it much easier to start diagnosing issues right away. Or at the very least, include the exception as an inner exception of InvalidDataException.

Google.PowerShell/Common/GcloudWrapper.cs
```diff
@@ -0,0 +1,109 @@
...
+            }
+
+            throw new InvalidDataException("Failed to get access token from gcloud auth print-access-token.");
+        }
+
+        private static async Task<string> GetGCloudCommandOutput(string command, IDictionary<string, string> environment = null)
+        {
+            var actualCommand = $"gcloud {command} --format=json";
+            // This code depends on the fact that gcloud.cmd is a batch file.
+            Debug.Write($"Executing gcloud command: {actualCommand}");
```
chrsmith: Delete this line.

Google.PowerShell/Common/GcloudWrapper.cs
```diff
@@ -0,0 +1,109 @@
...
+            }
+            catch (Newtonsoft.Json.JsonException)
+            {
+                // Throw exception at the end.
+            }
+
+            throw new InvalidDataException("Failed to get access token from gcloud auth print-access-token.");
+        }
+
+        private static async Task<string> GetGCloudCommandOutput(string command, IDictionary<string, string> environment = null)
```
chrsmith: This method seems important enough to put an XML doc comment on.

How does it work? What assumptions does it make? what is the `environment` parameter used for?

Google.PowerShell/Common/GcloudWrapper.cs
```diff
@@ -0,0 +1,109 @@
...
+        {
+            var actualCommand = $"gcloud {command} --format=json";
+            // This code depends on the fact that gcloud.cmd is a batch file.
+            Debug.Write($"Executing gcloud command: {actualCommand}");
+            ProcessOutput processOutput = await ProcessUtils.GetCommandOutput("cmd.exe", $"/c {actualCommand}", environment);
```
chrsmith: I'm going to go out on a limb here and assume HORRIBLE THINGS HAPPEN if `actualCommand` has spaces in it. We probably need to quote it. Right?

chrsmith: Does this popup any UI? Or will it only execute in the background.

Google.PowerShell/Common/ProcessUtils.cs
```diff
@@ -0,0 +1,185 @@
...
+            {
+                while (!stream.EndOfStream)
+                {
+                    var line = stream.ReadLine();
+                    handler(null, new OutputHandlerEventArgs(line, outputStream));
+                }
+            });
+        }
+    }
+}
```
chrsmith: New line.

Google.PowerShell/Common/ProcessUtils.cs
```diff
@@ -0,0 +1,185 @@
...
+        public OutputHandlerEventArgs(string line, OutputStream stream)
+        {
+            Line = line;
+            OutputStream = stream;
+        }
+    }
+
+    /// <summary>
+    /// This class defines helper methods for starting sub-processes and getting the output from
+    /// the processes, including a helper to parse the output as json.
```
chrsmith: JSON is an acronym. All caps.

Google.PowerShell/Common/ProcessUtils.cs
```diff
@@ -0,0 +1,185 @@
...
+    /// </summary>
+    public static class ProcessUtils
```
chrsmith: Having written similar types of wrappers in the past, can we add some unit tests if possible? Even if the command is "echo %TMP%", it would be useful to have non-zero coverage of this tricky-to-get-right code.

Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
```diff
@@ -0,0 +1,138 @@
...
+using System;
+using System.Reflection;
+using System.Threading;
+
+namespace Google.PowerShell.Tests.Common
+{
+    [TestFixture]
+    /// <summary>
+    /// Assumes that the Cloud SDK is installed on the local machine and initialized.
+    ///
```
quoctruong: Deleted

Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
quoctruong: Changed.

Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
quoctruong: Added logic to null it out before each test.

Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
quoctruong: Changed to suggested comment.

Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
quoctruong: Fixed.

Google.PowerShell.Tests/Common/TestActiveUserCredential.cs
quoctruong: Changed the comment to use invalidate.

Google.PowerShell.Tests/Common/TestActiveUserToken.cs
quoctruong: Fixed.

Google.PowerShell/Common/ActiveUserCredential.cs
quoctruong: Changed the class name.

Google.PowerShell/Common/ActiveUserCredential.cs
quoctruong: Fixed.

Google.PowerShell/Common/ActiveUserCredential.cs
quoctruong: I no longer use Object here (SemaphoreSlim instead).

Google.PowerShell/Common/ActiveUserCredential.cs
quoctruong: Deleted.

Google.PowerShell/Common/ActiveUserCredential.cs
quoctruong: Deleted this field.

Google.PowerShell/Common/ActiveUserCredential.cs
quoctruong: Remove this properties. We don't need to expose the Token object.

Google.PowerShell/Common/ActiveUserCredential.cs
quoctruong: Added comment.

Google.PowerShell/Common/ActiveUserCredential.cs
quoctruong: Remove the constructor altogether.

Google.PowerShell/Common/ActiveUserCredential.cs
quoctruong: Changed to suggested.

Google.PowerShell/Common/ActiveUserCredential.cs
quoctruong: Added a new line.

Google.PowerShell/Common/ActiveUserToken.cs
quoctruong: Fixed.

Google.PowerShell/Common/ActiveUserToken.cs
quoctruong: Fixed.

Google.PowerShell/Common/CloudSdkSettings.cs
quoctruong: Fixed.

Google.PowerShell/Common/CloudSdkSettings.cs
quoctruong: Fixed.

Google.PowerShell/Common/CloudSdkSettings.cs
quoctruong: Fixed.

Google.PowerShell/Common/CloudSdkSettings.cs
quoctruong: Fixed.

Google.PowerShell/Common/CloudSdkSettings.cs
quoctruong: Deleted.

Google.PowerShell/Common/CloudSdkSettings.cs
quoctruong: Deleted.

Google.PowerShell/Common/CloudSdkSettings.cs
quoctruong: Fixed.

Google.PowerShell/Common/GcloudWrapper.cs
quoctruong: Fixed.

Google.PowerShell/Common/GcloudWrapper.cs
quoctruong: Deleted.

Google.PowerShell/Common/GcloudWrapper.cs
quoctruong: This is cached in CloudSdkSettings class.

Google.PowerShell/Common/GcloudWrapper.cs
quoctruong: Moved logic into ActiveUserToken. Also, the comment in the previous commit was wrong, there won't be any NullReferenceException thrown so I remove that try catch logic.

Google.PowerShell/Common/GcloudWrapper.cs
quoctruong: Fixed.

Google.PowerShell/Common/GcloudWrapper.cs
quoctruong: Fixed.

Google.PowerShell/Common/GcloudWrapper.cs
quoctruong: Added comments.

Google.PowerShell/Common/GcloudWrapper.cs
quoctruong: Deleted.

Google.PowerShell/Common/GcloudWrapper.cs
quoctruong: Added double quotes. It does not pop up any UI because we set NoNewWindow to true in ProcessUtils.

Google.PowerShell/Common/GcloudWrapper.cs
quoctruong: Added.

Google.PowerShell/Common/ProcessUtils.cs
quoctruong: Fixed.

Google.PowerShell/Common/ProcessUtils.cs
quoctruong: Added tests.

Google.PowerShell/Common/ProcessUtils.cs
quoctruong: Fixed.

Google.PowerShell/Common/ActiveUserToken.cs
```diff
@@ -12,37 +14,131 @@ public class ActiveUserToken
...
-        private string activeUser = CloudSdkSettings.GetSettingsValue("account");
+        private string activeUser;
 
-        /// <summary>Gets or sets the access token issued by the authorization server.</summary>
-        [Newtonsoft.Json.JsonPropertyAttribute("access_token")]
-        public string AccessToken { get; set; }
+        /// <summary>
+        /// This function will return the current active user.
+        /// </summary>
+        internal Func<string> getActiveUser;
```
chrsmith: Is it possible to make this simply `private`? Same with the setter on `AccessToken` and others.

Google.PowerShell/Common/ActiveUserToken.cs
```diff
@@ -12,37 +14,131 @@ public class ActiveUserToken
...
+        {
+            this.getActiveUser = getActiveUser;
+            activeUser = this.getActiveUser();
+
+            JToken parsedCredentialJson = JObject.Parse(userCredentialJson);
+            JToken accessTokenJson = parsedCredentialJson.SelectToken("access_token");
+
+            if (accessTokenJson == null || accessTokenJson.Type != JTokenType.String)
+            {
+                throw new InvalidDataException("Credential Json should contain access token key.");
```
chrsmith: JSON is an acronym, all caps.

Google.PowerShell/Common/ActiveUserToken.cs
```diff
@@ -12,37 +14,131 @@ public class ActiveUserToken
...
+            }
+
+            AccessToken = accessTokenJson.Value<string>();
+
+            JToken tokenExpiryJson = parsedCredentialJson.SelectToken("token_expiry");
+
+            // Token from GCE does not have expiry.
```
chrsmith: "Token from GCE" isn't quite a technical definition. I believe it is "Service account credentials do not expire."

Google.PowerShell/Common/ActiveUserToken.cs
```diff
@@ -12,37 +14,131 @@ public class ActiveUserToken
...
 
-        /// <summary>The date and time that this token was issued. <remarks>
+        /// <summary>
+        /// The date and time that this token was issued.
+        /// This is set in UTC time.
         /// It should be set by the CLIENT after the token was received from the server.
-        /// </remarks> 
         /// </summary>
-        public DateTime Issued { get; set; }
+        public DateTime ExpiredTime { get; internal set; }
```
chrsmith: I don't understand the name here. "ExpiredTime" is the date/time the token was issued? Why not use the old name "Issued"?

chrsmith: If it is instead, the time when the token expires, perhaps "Expiration" or "Expiry" would be clearer?

Google.PowerShell/Common/ActiveUserToken.cs
```diff
@@ -12,37 +14,131 @@ public class ActiveUserToken
...
+
+        /// <summary>
+        /// Constructs a new token by parsing the userTokenJson.
+        /// GetActiveUser will be using CloudSdkSettings.GetSettingsValue("account").
+        /// </summary>
+        public ActiveUserToken(string userTokenJson) : this(userTokenJson, () => CloudSdkSettings.GetSettingsValue("account"))
```
chrsmith: Looking at this code again, I don't like the idea of passing in the delegate to get the "current" user name. As it breaks the abstraction of the `ActiveUserToken`.

An `ActiveUserToken` contains all the information about a user token. It's name, access token, and when it expires. The "getCurrentUser" delegate is only needed when we check if the token is no longer valid... but that isn't a property of the `ActiveUserToken` class, but rather the outside world.

What do you think about removing the "getActiveUser" property entirely, and only exposing an `IsExpired` property? Whether or not the currently Cloud SDK user matches the `ActiveUserToken`'s `TokenUser` property, is a problem we should push onto callers.

Google.PowerShell/Common/ActiveUserToken.cs
```diff
@@ -12,37 +14,131 @@ public class ActiveUserToken
...
+            }
+
+            return ExpiredTime.AddSeconds(-60) <= DateTime.UtcNow;
         }
     }
+
+    /// <summary>
+    /// Represents a token expiry object returned from gcloud auth print-access-token.
+    /// </summary>
+    internal class TokenExpiry
```
chrsmith: This is only used as a shortcut when parsing the expiration from the JSON object, right? If so, why don't we make this a private inner-class of `ActiveUserToken`. The goal is to restrict the scope of this type name as much as possible, since it isn't used anywhere else.

Also, given the basic nature of the time, I'd personally be OK with removing the newlines between its properties and putting the attribute/properties on the same line to conserve space. e.g.

````
[JsonProperty("microsecond")] internal int MicroSecond { get; set; }
[JsonProperty("second")]      internal int Second { get; set; }
[JsonProperty("minute")]      internal int Minute { get; set; }
````

Just a matter of style. I'd suggest making the code as compact as possible, since it isn't as important.

File: Google.PowerShell/Common/AuthenticateWithSdkCredentialsExecutor.cs [hunk 46]
```diff
@@ -36,13 +36,14 @@ public class AuthenticateWithSdkCredentialsExecutor : ICredential, IHttpExecuteI
         /// rather than launching gcloud again.
```
chrsmith: In the class declaration, put the `: ICredential, IHttpExecuteInterceptor, ` etc. on a new line to avoid going over 120 chars?

Google.PowerShell/Common/AuthenticateWithSdkCredentialsExecutor.cs
```diff
@@ -125,13 +97,20 @@ public void Initialize(ConfigurableHttpClient httpClient)
...
         public virtual async Task<string> GetAccessTokenForRequestAsync(string authUri = null, CancellationToken cancellationToken = default(CancellationToken))
         {
-            if (Token.IsExpired(flow.Clock))
+            if (s_token == null)
+            {
+                await RefreshTokenAsync(cancellationToken).ConfigureAwait(false);
+                return s_token.AccessToken;
+            }
+
+            if (s_token.IsExpiredOrInvalid())
```
chrsmith: Following up with a comment I made earlier, it makes sense that the `AuthenticateWithSdkCredentialsExecutor` class knows about the Cloud SDK and so we can put the "has the active user changed" logic here. Perhaps:

````
string currentCloudSdkUser = CloudSDKConfig.GetProperty("user");
if (s_token.IsExpired || (s_token.User != currentCloudSdkUser)) {
    // refresh the token...
}
````

File: Google.PowerShell/Common/AuthenticateWithSdkCredentialsExecutor.cs [hunk 48]
```diff
@@ -142,10 +121,16 @@ public void Initialize(ConfigurableHttpClient httpClient)
         /// </summary>
```
chrsmith: End the summary comment above with a period.

Google.PowerShell/Common/AuthenticateWithSdkCredentialsExecutor.cs
```diff
@@ -142,10 +121,16 @@ public void Initialize(ConfigurableHttpClient httpClient)
...
+            await s_tokenLock.WaitAsync();
+            try
+            {
+                s_token = await GCloudWrapper.GetAccessToken(taskCancellationToken);
+                return true;
+            }
+            finally
+            {
+                s_tokenLock.Release();
+            }
```
chrsmith: `return false;`? It seems strange to have a method that has a return type of `bool` and yet who only returns `true` or throws an exception.

Google.PowerShell/Common/AuthenticateWithSdkCredentialsExecutor.cs
```diff
@@ -36,13 +36,14 @@ public class AuthenticateWithSdkCredentialsExecutor : ICredential, IHttpExecuteI
...
 
         /// <summary>
-        /// We just need a default flow for things like flow.AccessMethod.Intercept
-        /// or flow.RevokeTokenAsync.
+        /// Returns the authorization code flow. 
+        /// This is used to revoke token (in RevokeTokenAsync)
+        /// and intercept request with the access token (in InterceptAsync).
         /// </summary>
-        private readonly IAuthorizationCodeFlow flow = new GoogleAuthorizationCodeFlow(
+        public IAuthorizationCodeFlow Flow { get; } = new GoogleAuthorizationCodeFlow(
```
chrsmith: Given that `Flow` is only used inside of this class, perhaps makes sense tom make it private?

Google.PowerShell/Common/AuthenticateWithSdkCredentialsExecutor.cs
```diff
@@ -102,8 +75,7 @@ public AuthenticateWithSdkCredentialsExecutor()
         {
             if (args.Response.StatusCode == HttpStatusCode.Unauthorized)
             {
-                return !Object.Equals(Token.AccessToken, flow.AccessMethod.GetAccessToken(args.Request))
-                    || await RefreshTokenAsync(args.CancellationToken).ConfigureAwait(false);
+                return await RefreshTokenAsync(args.CancellationToken).ConfigureAwait(false);
```
chrsmith: What is the `IHttpUnsuccessfulResponseHandler` interface for and what is the significance of  `HandleResponseAsync` 's return value?

If we successfully refresh the auth token, does it retry the request? What if you legitimately access a resource you don't have access to?

We should probably add a comment to the top of the method explaining its role, even if it is just a copy of the interface's documentation. Do you agree?

Google.PowerShell/Common/GcloudWrapper.cs
```diff
@@ -56,43 +51,28 @@ public static class GCloudWrapper
...
-            catch (NullReferenceException)
-            {
-                // Throw exception at the end.
-            }
-            catch (Newtonsoft.Json.JsonException)
-            {
-                // Throw exception at the end.
-            }
+            string userCredentialJson = await GetGCloudCommandOutput("auth print-access-token");
+            cancellationToken.ThrowIfCancellationRequested();
```
chrsmith: I don't think we should wire in the cancellation token if we don't actually honor it. `GetGCloudCommandOutput` will wait until the process exits regardless of the cancellation state. So we should either:

- Be honest with ourselves, and not even add the cancellation token as a parameter.
- Wire the cancellation token down into ProcessUtils where we call `WaitForExit` and kill the process as soon as cancellation is requested. (I did some Googling and you can see an example here https://gist.github.com/mxpv/3798871).

I'm inclined to go with the former, and not bother with wiring in cancellation if we won't actually support it.

quoctruong:
@chrsmith PTAL

Google.PowerShell/Common/ActiveUserToken.cs
```diff
@@ -0,0 +1,118 @@
...
+    /// </summary>
+    public class ActiveUserToken
+    {
+        /// <summary>
+        /// The user that this token corresponds to.
+        /// </summary>
+        public string User;
+
+        /// <summary>The access token issued by the authorization server.</summary>
+        public string AccessToken { get; internal set; }
```
chrsmith: Make this setter `private` as well as the one on line 27. (Unless there is a reason for other types to modify the field within the same assembly?)

Google.PowerShell/Common/ActiveUserToken.cs
```diff
@@ -0,0 +1,118 @@
...
+
+                return ExpiredTime.AddSeconds(-60) <= DateTime.UtcNow;
+            }
+        }
+
+        /// <summary>
+        /// Constructs a new token by parsing userCredentialJson.
+        /// User will be set to the current active user.
+        /// </summary>
+        public ActiveUserToken(string userCredentialJson) :
```
chrsmith: Given that this constructor is only used in one place, and it coules this type with the `CloudSdkSettings`, why don't we remove this. And force callers to specify the user.

Google.PowerShell/Common/AuthenticateWithSdkCredentialsExecutor.cs
```diff
@@ -0,0 +1,159 @@
...
+        /// </summary>
+        private static ActiveUserToken s_token;
+        private static readonly SemaphoreSlim s_tokenLock = new SemaphoreSlim(1);
+
+        /// <summary>
+        /// Returns the authorization code flow. 
+        /// This is used to revoke token (in RevokeTokenAsync)
+        /// and intercept request with the access token (in InterceptAsync).
+        /// </summary>
+        public IAuthorizationCodeFlow Flow { get; } = new GoogleAuthorizationCodeFlow(
```
chrsmith: Let's make this property private, since (if I am reading the code correctly) it is not used outside of the class. Also, perhaps it should just be class field? The ` { get; } ` isn't buying us anything. (We'd need to rename the field to be a lowercase `flow` though.

Google.PowerShell/Common/AuthenticateWithSdkCredentialsExecutor.cs
```diff
@@ -0,0 +1,159 @@
...
+
+        #endregion
+
+        #region IHttpUnsuccessfulResponseHandler
+
+        public async Task<bool> HandleResponseAsync(HandleUnsuccessfulResponseArgs args)
+        {
+            if (args.Response.StatusCode == HttpStatusCode.Unauthorized)
+            {
+                return await RefreshTokenAsync(args.CancellationToken).ConfigureAwait(false);
```
chrsmith: Could you add the doc comment from `IHttpUnsuccessfulResponseHandler.HandleResponseAsync` here? I don't know the purpose of returning the result of `RefreshTokenAsync` here, and so I'm inclined to assume we are doing the wrong thing.

Google.PowerShell/Common/AuthenticateWithSdkCredentialsExecutor.cs
```diff
@@ -0,0 +1,159 @@
...
+                await RefreshTokenAsync(cancellationToken).ConfigureAwait(false);
+                return s_token.AccessToken;
+            }
+
+            string currentCloudSdkUser = CloudSdkSettings.GetSettingsValue("account");
+            if (s_token.IsExpired || s_token.User != currentCloudSdkUser)
+            {
+                if (!await RefreshTokenAsync(cancellationToken).ConfigureAwait(false))
+                {
+                    throw new InvalidOperationException("The access token has expired but we can't refresh it");
```
chrsmith: "Token has expired *or the user has changed*, but we can't refresh it."

Google.PowerShell/Common/CloudSdkSettings.cs
```diff
@@ -36,25 +48,47 @@ public class CommonProperties
         /// <summary>Name of the file containing the anonymous client ID.</summary>
         private const string ClientIDFileName = ".metricsUUID";
 
+        private static string s_installationPropertiesPath;
```
chrsmith: Add a doc comment explaining what this field will hold. e.g. "Backing field for the InstallationPropertiesPath property."

Google.PowerShell/Common/CloudSdkSettings.cs
```diff
@@ -36,25 +48,47 @@ public class CommonProperties
...
+        /// </summary>
+        public static string InstallationPropertiesPath
+        {
+            get
+            {
+                if (string.IsNullOrWhiteSpace(s_installationPropertiesPath))
+                {
+                    s_installationPropertiesPath = GCloudWrapper.GetInstallationPropertiesPath().Result;
+                }
+
```
chrsmith: Delete this line.

Google.PowerShell/Common/GcloudWrapper.cs
```diff
@@ -0,0 +1,89 @@
...
+        public static async Task<ActiveUserToken> GetAccessToken(CancellationToken cancellationToken)
+        {
+            // We get the issued time before the command so we won't be too late
+            // when it comes to token expiry.
+            DateTime issuedTime = DateTime.Now;
+
+            string userCredentialJson = await GetGCloudCommandOutput("auth print-access-token");
+            cancellationToken.ThrowIfCancellationRequested();
+
+            return new ActiveUserToken(userCredentialJson);
```
chrsmith: Let's make the current user lookup explicit here, and use the remaining constructor of `ActiveUsertoken`. e.g.

````
string currentUser = CloudSdkSettings.GetProperty("user");
return new ActiveUsertoken(userCredentialJson, currentUser);
````

PR #353, opened 2016-10-19
Make New-GcsObject's -ObjectName parameter optional when uploading a file

### LINKED ISSUE #94: Should New-GcsObject make "name" optional for file upload?
Having to specify the filename of the file you are passing to it seems weird.

chrsmith:
When uploading a file, we can use the file's name as the default for ObjectName instead of making the parameter mandatory. Fixes #94.

Also tweaks some documentation values. (e.g. "UploadFolder" is a more descriptive parameter set name than "ContentsFromFolder" IMHO.)


chrsmith:
@quoctruong PTAL

PR #363, opened 2016-10-26
Update files for 0.2.0 release; Add function in buildandpackage script to simply the process

quoctruong:
The AssemblyInfo.cs, GoogleCloud.psd1 and cmdletsFull.json changes are all generated using BuildAndpackage.ps1 script

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/363)


ILMTitan:
Shouldn't this be the 0.1.10 release?

quoctruong:
@ILMTitan Updated the release. PTAL.

Tools/BuildAndPackage.ps1
```diff
@@ -14,18 +26,33 @@ $slnFile = Join-Path $projectRoot "gcloud-powershell.sln"
...
 
 # Purge the existing bin directory.
 Write-Host -ForegroundColor Cyan "*** Purging the bindir***"
 if (Test-Path $binDir) {
     Remove-Item $binDir -Recurse -Force
 }
 
+# Change the module manifest version
+$moduleManifestFile = Join-Path $projectRoot "Google.PowerShell\ReleaseFiles\GoogleCloud.psd1"
+$moduleManifestContent = Get-Content $moduleManifestFile -Raw
```
ILMTitan: I think I would have preferred something that used Import-PowerShellDataFile, but this looks ok.

ILMTitan:
I have one thought about how to load the manifest as an object rather than a string, but otherwise LGTM.

Tools/BuildAndPackage.ps1
```diff
@@ -14,18 +26,33 @@ $slnFile = Join-Path $projectRoot "gcloud-powershell.sln"
 $binDir = Join-Path $projectRoot "Google.PowerShell\bin\"
 $debugDir = Join-Path $binDir "Debug"
 
+
```
chrsmith: Unnecessary vertical whitespace.

In general, avoid having two successive lines of vertical whitespace except in very rare circumstances. (The JavaScript style guide mandates it for closure classes, and IIRC, Python classes.)

Tools/BuildAndPackage.ps1
```diff
@@ -1,3 +1,8 @@
+[CmdletBinding()]
+Param(
```
chrsmith: Let's put a TODO to confirm the value is correct. If we incorrectly use a lower version, we could publish bits that won't get updated correctly. Perhaps we can use the GitHub API to check for the latest release, or confirm we have the previous release on Cloud Storage, etc.

PR #369, opened 2016-11-07
Fix module manifest file to explicitly list out cmdlets; Also uses RootModule instead of NestedModules

quoctruong:
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/369)


chrsmith:
The change looks good. But I'd rather not LGTM this as it would get out of date the second we add our next cmdlet.

How did you update the module manifest? If it is something we can do in an automatic fashion, perhaps it is something we should do as part of "BuildAndPackage". But if we did that, we still wouldn't be able to load new cmdlets for testing before we ship...

quoctruong:
I updated the module manifest by getting all the cmdlets from GoogleCloud module and populate that field. I think what we can do is have a manifest that load every cmdlets by default for testing and things like that. And then when we actually package up the module manifest for testing, we create a new manifest?

quoctruong:
@chrsmith PTAL

Google.PowerShell/ReleaseFiles/GoogleCloudPowerShell.psd1
```diff
@@ -7,16 +7,16 @@
...
 
 @{
 
 # Script module or binary module file associated with this manifest.
-# RootModule = ''
+RootModule = '..\PowerShell\GoogleCloud\Google.PowerShell.dll'
 
 # Version number of this module.
-ModuleVersion = '0.1.0'
+ModuleVersion = '0.1.10.0'
```
iantalarico: Should there be an extra .0 added? or should this be "0.1.10"? or should it be "0.1.01"?  The latter seems right to me

Google.PowerShell/ReleaseFiles/GoogleCloudPowerShell.psd1
```diff
@@ -64,22 +64,22 @@ ProcessorArchitecture = 'None'
...
-FormatsToProcess = @("..\PowerShell\GoogleCloud\GoogleCloudPlatform.Format.ps1xml")
+FormatsToProcess = '..\PowerShell\GoogleCloud\GoogleCloudPlatform.Format.ps1xml'
 
 # Modules to import as nested modules of the module specified in RootModule/ModuleToProcess
-NestedModules = @('..\PowerShell\GoogleCloud\Google.PowerShell.dll')
+NestedModules = @()
 
 # Functions to export from this module
-FunctionsToExport = '*'
+# FunctionsToExport = @()
```
iantalarico: Why are these few items commented out? Should they just be deleted?  If not can you add a comment?

Tools/BuildAndPackage.ps1
```diff
@@ -75,17 +120,29 @@ Write-Host -ForegroundColor Cyan "*** HACK: Creating GoogleCloudPowerShell modul
...
+New-ModuleManifest -Path "$gcpsDir\GoogleCloudPowerShell.psd1" @gCloudPowerShellManifest
 
 # Package the bits. Requires setting up the right directory structure.
 Write-Host -ForegroundColor Cyan "*** Packaging the bits ***"
 
 New-Item -ItemType Directory $powerShellDir
 Copy-Item -Recurse $debugDir $powerShellDir
 
-# The binaries in a folder named "Debug". Rename that to "GooglePowerShell".
+# The binaries are in a folder named "Debug". Move them to GoogleCloud\$normalizedVersion folder.
```
iantalarico: Might be outside the scope of this review but does this mean they are being built with debug instead of as release?

quoctruong: Yes, they are always being built as Debug.

Google.PowerShell/ReleaseFiles/GoogleCloudPowerShell.psd1
quoctruong: There are automatically generated from New-ModuleManifest cmdlet and when we don't supply anything, they will be generated like this.

Google.PowerShell/ReleaseFiles/GoogleCloudPowerShell.psd1
quoctruong: It should be 4 parts versioning.

Tools/BuildAndPackage.ps1
iantalarico: Should they be? Or should they be built as Release?

quoctruong: I was going to make a separate pull request to change them to release. I can do it in this PR too.

iantalarico: It's okay to do in another PR.  I'd rather not have PR creep.

quoctruong:
Thanks!

PR #376, opened 2016-11-10
Fix error messages when objects or buckets are not found.

quoctruong:
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/376)

### DIFF
--- a/Google.PowerShell/Storage/GcsObject.cs
+++ b/Google.PowerShell/Storage/GcsObject.cs
@@ -437,8 +437,20 @@ namespace Google.PowerShell.CloudStorage
 
             ObjectsResource.GetRequest getReq = Service.Objects.Get(Bucket, ObjectName);
             getReq.Projection = ObjectsResource.GetRequest.ProjectionEnum.Full;
-            Object gcsObject = getReq.Execute();
-            WriteObject(gcsObject);
+            try
+            {
+                Object gcsObject = getReq.Execute();
+                WriteObject(gcsObject);
+            }
+            catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.NotFound)
+            {
+                ErrorRecord errorRecord = new ErrorRecord(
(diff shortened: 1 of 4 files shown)


Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -832,10 +841,25 @@ protected override void ProcessRecord()
...
+                CheckForError(result);
+            }
+        }
+
+        private void CheckForError(IDownloadProgress downloadProgress)
+        {
+            if (downloadProgress.Status == DownloadStatus.Failed || downloadProgress.Exception != null)
+            {
+                string exceptionMessage = downloadProgress.Exception.Message;
+                if (exceptionMessage != null && exceptionMessage.IndexOf("Not Found", System.StringComparison.OrdinalIgnoreCase) >= 0)
```
iantalarico: Is this a safe check? Shouldn't we check and error code? 

exceptionMessage.IndexOf("Not Found"

quoctruong:
@iantalarico PTAL

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -849,8 +849,8 @@ private void CheckForError(IDownloadProgress downloadProgress)
         {
             if (downloadProgress.Status == DownloadStatus.Failed || downloadProgress.Exception != null)
             {
-                string exceptionMessage = downloadProgress.Exception.Message;
-                if (exceptionMessage != null && exceptionMessage.IndexOf("Not Found", System.StringComparison.OrdinalIgnoreCase) >= 0)
+                GoogleApiException googleApiException = downloadProgress.Exception as GoogleApiException;
```
iantalarico: Do you need to check if this is a GoogleApiException or will it always be?

quoctruong: I have a check in line 853 so if it is not a GoogleApiException, we will just throw whatever exception message that it has.

quoctruong:
Thanks!

PR #377, opened 2016-11-14
Merge Find-GcsObject into Get-GcsObject; Fix error message for Remove-GcsObject when object not found

quoctruong:
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/377)

### DIFF
--- a/Google.PowerShell/Storage/GcsObject.cs
+++ b/Google.PowerShell/Storage/GcsObject.cs
@@ -417,9 +417,12 @@ namespace Google.PowerShell.CloudStorage
     /// </para>
     /// <para type="description">
     /// An optional delimiter may be provided. If used, will return results in a
-    /// directory-like mode, delimited by the given string. e.g. with objects "1,
-    /// "2", "subdir/3" and delimited "/"; "subdir/3" would not be returned.
-    /// (There is no way to just return "subdir" in the previous example.)
+    /// directory-like mode, delimited by the given string. This means that the names
+    /// of all objects returned will not, aside from the prefix, contain the delimiter.
+    /// For example, with objects "1", "2", "subdir/3", "subdir/subdir/4",
+    /// if the delimiter is "/", only "1" and "2" will be returned.
+    /// If the delimiter is "/" and the prefix is "subdir/", only "subdir/3"
+    /// will be returned.
     /// </para>
     /// <para type="description">
     /// To gets a specific Cloud Storage Object by name, use the -ObjectName parameter.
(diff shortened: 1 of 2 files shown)


Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -398,17 +398,51 @@ private static string ConvertLocalToGcsFolderPath(string localFilePath)
...
     /// </para>
     /// <para type="description">
-    /// Returns the give Storage object's metadata.
+    /// If no parameter besides -Bucket is provided, all objects in the bucket are returned.
+    /// If a given prefix string is provided, returns all Cloud Storage objects identified
+    /// by the prefix string.
+    /// </para>
+    /// <para type="description">
+    /// An optional delimiter may be provided. If used, will return results in a
+    /// directory-like mode, delimited by the given string. e.g. with objects "1,
```
iantalarico: The description and example are a little hard to follow.

Is this well documented for GCS?

Can you add a more detailed example?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -419,37 +453,90 @@ public class GetGcsObjectCmdlet : GcsCmdlet
...
+                // When used with WriteObject, expand the IEnumerable rather than
+                // returning the IEnumerable itself. IEnumerable<T> vs. IEnumerable<IEnumerable<T>>.
+                const bool enumerateCollection = true;
+
+                // First page.
+                Objects gcsObjects = listReq.Execute();
+                WriteObject(gcsObjects.Items, enumerateCollection);
+
+                // Keep paging through results as necessary.
+                while (!Stopping && gcsObjects.NextPageToken != null)
```
iantalarico: I'm guessing this is a powershell thing I'm missing but where is "Stopping" coming from?

quoctruong: It's a property of the PowerShell cmdlet class. It is true if the cmdlet execution stopped (like if user press Ctrl + C).

Google.PowerShell/Storage/GcsObject.cs
quoctruong: Is this description better? "Returns results in a directory-like mode. items will contain only objects whose names, aside from the prefix, do not contain delimiter. Objects whose names, aside from the prefix, contain delimiter will have their name, truncated after the delimiter, returned in prefixes. Duplicate prefixes are omitted." This is from [here](https://cloud.google.com/storage/docs/json_api/v1/objects/list).

iantalarico: Ah yea that is much more clear, thanks :)

quoctruong:
Thanks!

PR #387, opened 2016-11-28
Update BuildAndPackage script to compress archive correctly

quoctruong:
Context:
If the folder structure is something like "PowerShell\GoogleCloud\1.0.0.0\<Files>", then Compress-Archive will create an empty "PowerShell\GoogleCloud" folder in the ZipArchive. This causes a lot of test failures when we submit to the GoogleCloudSDK.
So in this PR, I first copied the Compress-Archive cmdlets from PowerShell repository and then change the cmdlet so that it will not create the empty folder.
The first commit is just copying the code from PowerShell repository. All the changes are made in the second commit.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/387)

### DIFF
--- a/Tools/ArchiveCmdlets/ArchiveResources.psd1
+++ b/Tools/ArchiveCmdlets/ArchiveResources.psd1
@@ -1,0 +1,25 @@
+﻿# Localized ArchiveResources.psd1
+
+ConvertFrom-StringData @'
+###PSLOC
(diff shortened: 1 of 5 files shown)


ILMTitan:
LGTM. I would suggest you not squash the changes, so what was changed is visable.

quoctruong:
Thanks!

PR #392, opened 2016-11-30
Add Google Cloud PubSub Topics cmdlets.

quoctruong:
Add New-GcpsTopic, Get-GcpsTopic and Remove-GcpsTopic.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/392)

### DIFF
--- a/Google.PowerShell.IntegrationTests/PubSub/PubSub.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/PubSub/PubSub.Tests.ps1
@@ -35,6 +35,10 @@ Describe "Get-GcpsTopic" {
     It "should throw an error for non-existent topic" {
         { Get-GcpsTopic -Topic "non-existent-topic-name" -ErrorAction Stop } | Should Throw "does not exist"
     }
+
+    It "should throw error for invalid topic name" {
+        { Get-GcpsTopic -Topic "!!" -ErrorAction Stop } | Should Throw "Invalid resource name"
+    }
 }
 
 Describe "New-GcpsTopic" {
@@ -75,13 +79,13 @@ Describe "New-GcpsTopic" {
             gcloud beta pubsub topics delete $topicName --quiet 2>$null
         }
     }
+
+    It "should throw error for invalid topic name" {
+        { New-GcpsTopic -Topic "!!" -ErrorAction Stop } | Should Throw "Invalid resource name"
+    }
 }
 
 Describe "Remove-GcpsTopic" {
-    It "should throw error for non-existent topic" {
-        { Remove-GcpsTopic -Topic "non-existent-topic-powershell-testing" -ErrorAction Stop } | Should Throw "does not exist"
-    }
-
     It "should work" {
         $r = Get-Random
         $topicName = "gcp-creating-topic-$r"
@@ -97,5 +101,35 @@ Describe "Remove-GcpsTopic" {
         # Remove an array of topics.
         Remove-GcpsTopic -Topic $secondTopicName, $thirdTopicName
         { Get-GcpsTopic -Topic $secondTopicName, $thirdTopicName -ErrorAction Stop } | Should Throw "does not exist"
+    }
+
+    It "should work with pipeline" {
+        $r = Get-Random
+        $topicName = "gcp-creating-topic-$r"
+        New-GcpsTopic -Topic $topicName
+        Get-GcpsTopic -Topic $topicName | Should Not BeNullOrEmpty
+
+        # Remove through pipeline
+        Get-GcpsTopic -Topic $topicName | Remove-GcpsTopic
+        { Get-GcpsTopic -Topic $topicName -ErrorAction Stop } | Should Throw "does not exist"
+    }
+
+    It "should throw error for non-existent topic" {
+        { Remove-GcpsTopic -Topic "non-existent-topic-powershell-testing" -ErrorAction Stop } | Should Throw "does not exist"
+    }
+
+    It "should throw error for invalid topic name" {
+        { Remove-GcpsTopic -Topic "!!" -ErrorAction Stop } | Should Throw "Invalid resource name"
+    }
+
+    It "should not remove topic if -WhatIf is used" {
+        $r = Get-Random
+        $topicName = "gcp-creating-topic-$r"
+        New-GcpsTopic -Topic $topicName
+        Get-GcpsTopic -Topic $topicName | Should Not BeNullOrEmpty
+
+        # Topic should not be removed.
+        Remove-GcpsTopic -Topic $topicName -WhatIf
+        Get-GcpsTopic -Topic $topicName | Should Not BeNullOrEmpty
     }
 }
--- a/Google.PowerShell.IntegrationTests/PubSub/PubSub.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/PubSub/PubSub.Tests.ps1
@@ -1,0 +1,101 @@
+﻿. $PSScriptRoot\..\GcloudCmdlets.ps1
+Install-GCloudCmdlets
+
+$project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
+
+Describe "Get-GcpsTopic" {
+    $r = Get-Random
+    $script:topicName = "gcps-testing-topic-$r"
+    $script:secondTopicName = "gcps-testing-topic2-$r"
+    $previousCount = (Get-GcpsTopic).Count
+    gcloud beta pubsub topics create $script:topicName 2>$null
+    gcloud beta pubsub topics create $script:secondTopicName 2>$null
+
+    AfterAll {
+        gcloud beta pubsub topics delete $topicName --quiet 2>$null
+        gcloud beta pubsub topics delete $secondTopicName --quiet 2>$null
+    }
+
+    It "should work without any parameters" {
+        $topics = Get-GcpsTopic
+        $topics.Count - $previousCount | Should Be 2
+    }
+
+    It "should work with -Topic parameter" {
+        $firstTopic = Get-GcpsTopic -Topic $topicName
+        $firstTopic.Name | Should BeExactly "projects/$project/topics/$topicName"
+
+        # Should work with an array of topics names.
+        $topics = Get-GcpsTopic -Topic $topicName, $secondTopicName
+        $topics.Count | Should Be 2
+        $topics.Name -contains "projects/$project/topics/$topicName" | Should Be $true
+        $topics.Name -contains "projects/$project/topics/$secondTopicName" | Should Be $true
+    }
+
+    It "should throw an error for non-existent topic" {
+        { Get-GcpsTopic -Topic "non-existent-topic-name" -ErrorAction Stop } | Should Throw "does not exist"
+    }
+}
+
+Describe "New-GcpsTopic" {
+    It "should work" {
+        $r = Get-Random
+        $topicName = "gcp-creating-topic-$r"
+        $secondTopicName = "gcp-creating-topic2-$r"
+        $thirdTopicName = "gcp-creating-topic3-$r"
+
+        try {
+            New-GcpsTopic -Topic $topicName
+            (Get-GcpsTopic -Topic $topicName).Name | Should BeExactly "projects/$project/topics/$topicName"
+
+            # Should work with an array of topics names.
+            New-GcpsTopic -Topic $secondTopicName, $thirdTopicName
+            $topics = Get-GcpsTopic -Topic $secondTopicName, $thirdTopicName
+            $topics.Count | Should Be 2
+            $topics.Name -contains "projects/$project/topics/$secondTopicName" | Should Be $true
+            $topics.Name -contains "projects/$project/topics/$thirdTopicName" | Should Be $true
+        }
+        finally {
+            gcloud beta pubsub topics delete $topicName --quiet 2>$null
+            gcloud beta pubsub topics delete $secondTopicName --quiet 2>$null
+            gcloud beta pubsub topics delete $thirdTopicName --quiet 2>$null
+        }
+    }
+
+    It "should throw an error for topic that is already created" {
+        $r = Get-Random
+        $topicName = "gcp-creating-topic-$r"
+
+        try {
+            New-GcpsTopic -Topic $topicName
+            (Get-GcpsTopic -Topic $topicName).Name | Should BeExactly "projects/$project/topics/$topicName"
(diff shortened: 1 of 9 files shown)


Google.PowerShell/PubSub/GcpsCmdlet.cs
```diff
@@ -0,0 +1,49 @@
+﻿using Google.Apis.Pubsub.v1;
```
chrsmith: We have a tool to automatically insert copyright headers on all files. You might want to rerun that on a separate PR, since it will probably fix up a bunch of other files in the repo.

Google.PowerShell/PubSub/GcpsCmdlet.cs
```diff
@@ -0,0 +1,49 @@
...
+using Google.PowerShell.Common;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Google.PowerShell.PubSub
+{
+    /// <summary>
+    /// Base class for Google Cloud PubSub Cmdlets.
```
chrsmith: lowercase c for cmdlets?

Google.PowerShell/PubSub/GcpsCmdlet.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            Service = new PubsubService(GetBaseClientServiceInitializer());
+        }
+
+        /// <summary>
+        /// Prefix projects/{project name}/topics/{topic} to topicName if not present.
+        /// </summary>
+        protected string PrefixProjectToTopic(string topicName, string project)
+        {
+            if (string.Equals(topicName, "_deleted-topic_", System.StringComparison.OrdinalIgnoreCase))
```
chrsmith: Why not just `topicName == "_deleted-topic_"`? Or just let this case be handled by the `return topicName;` at the very end? Calling out this case specifically might make sense, but in that case we might want to add a comment to explain why.

Google.PowerShell/PubSub/GcpsCmdlet.cs
```diff
@@ -0,0 +1,49 @@
...
+            {
+                topicName = $"projects/{project}/topics/{topicName}";
+            }
+            return topicName;
+        }
+
+        /// <summary>
+        /// Prefix projects/{project name}/subscriptions/{subscriptions} to subscriptionName if not present.
+        /// </summary>
+        protected string PrefixProjectToSubscription(string subscriptionName, string project)
```
chrsmith: Perhaps rename this to "GetProjectPrefixForSubscription"? "PrefixProject" doesn't quite make sense at first glance. I don't feel too strongly about it, so feel free to push back.

Google.PowerShell/PubSub/GcpsTopic.cs
```diff
@@ -0,0 +1,234 @@
...
+using System.Net;
+
+namespace Google.PowerShell.PubSub
+{
+    /// <summary>
+    /// <para type="synopsis">
+    /// Creates new Google Cloud PubSub topics.
+    /// </para>
+    /// <para type="description">
+    /// Creates one or more Gooogle Cloud PubSub topics. Will throw errors if the topics already exist.
```
chrsmith: "throw exceptions" or "raise errors"?

Google.PowerShell/PubSub/GcpsTopic.cs
```diff
@@ -0,0 +1,234 @@
...
+                Topic topic = new Topic() { Name = formattedTopicname };
+                ProjectsResource.TopicsResource.CreateRequest request = Service.Projects.Topics.Create(topic, formattedTopicname);
+                try
+                {
+                    Topic response = request.Execute();
+                    WriteObject(response);
+                }
+                catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.Conflict)
+                {
+                    ErrorRecord errorRecord = new ErrorRecord(
```
chrsmith: This is pretty inconsistent with the way we have been doing things in the past. (Either letting the exception bubble up, or just calling `WriteError` with a string.)

Should we be trying to clean up other places in the code to use this more elaborate pattern? Should we try to simplify it some how? e.g. `WriteResourceExistsError("topic", topicName, "TopicAlreadyExists");`?

Google.PowerShell/PubSub/GcpsTopic.cs
```diff
@@ -0,0 +1,234 @@
...
+        {
+            if (Topic != null && Topic.Length > 0)
+            {
+                foreach (string topicName in Topic)
+                {
+                    string formattedTopicName = PrefixProjectToTopic(topicName, Project);
+                    try
+                    {
+                        ProjectsResource.TopicsResource.GetRequest getRequest = Service.Projects.Topics.Get(formattedTopicName);
+                        WriteObject(getRequest.Execute());
```
chrsmith: Do we want to return the full response object? Wouldn't we instead want to write:

````
TopicResource.GetResponse getResponse = getRequest.Execute();
WriteObject(getResponse.Topics, true /* expand enumerable */);
````

Right? Or am I making invalid assumptions?

File: Google.PowerShell/PubSub/GcpsTopic.cs [hunk 8]
```diff
@@ -0,0 +1,234 @@
...
+                    }
+                }
+            }
+            else
+            {
+                ProjectsResource.TopicsResource.ListRequest listRequest = Service.Projects.Topics.List($"projects/{Project}");
+                do
+                {
+                    ListTopicsResponse response = listRequest.Execute();
+
```
chrsmith: Remove this line of whitespace? The code seems straight forward enough to not need it.

Google.PowerShell/PubSub/GcpsTopic.cs
```diff
@@ -0,0 +1,234 @@
...
+        /// <summary>
+        /// <para type="description">
+        /// The names of the topics to be retrieved.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, Position = 0)]
+        [ValidateNotNullOrEmpty]
+        [Alias("Name")]
+        public string[] Topic { get; set; }
```
chrsmith: Should this be "Topics"? Or is the pattern in PowerShell not to pluralize things? It just seems strange to specify multiple items for a parameter named singular "Topic".

Google.PowerShell/PubSub/GcpsTopic.cs
```diff
@@ -0,0 +1,234 @@
...
+    ///   <code>PS C:\> Remove-GcpsTopic -Topic "topic1", "topic2" -Project "my-project"</code>
+    ///   <para>
+    ///   This command removes 2 topics ("topic1" and "topic2") in the project "my-project".
+    ///   </para>
+    /// </example>
+    /// <para type="link" uri="(https://cloud.google.com/pubsub/docs/publisher#delete)">
+    /// [Deleting a Topic]
+    /// </para>
+    /// </summary>
+    [Cmdlet(VerbsCommon.Remove, "GcpsTopic")]
```
chrsmith: Since this deletes data, we should add `SupportsShouldProcess = true` and then call `ShouldProcess` depending on the Force/Confirm flag.

Google.PowerShell/PubSub/GcpsTopic.cs
```diff
@@ -0,0 +1,234 @@
...
+        /// default to the Cloud SDK's DefaultProject property.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        [ConfigPropertyName(CloudSdkSettings.CommonProperties.Project)]
+        public string Project { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The names of the topics to be removed.
```
chrsmith: Rename to `Topics`? Change text to "topic(s)"? Something just doesn't seem right about the tense here, though if I am wrong about the PowerShell idiom, do whatever is the standard.

Google.PowerShell/PubSub/GcpsTopic.cs
```diff
@@ -0,0 +1,234 @@
...
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, Position = 0)]
+        [ValidateNotNullOrEmpty]
+        [Alias("Name")]
+        public string[] Topic { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            if (Topic != null && Topic.Length > 0)
```
chrsmith: With the `ValidateNotNullOrEmpty` attribute, isn't this the only valid code path? e.g. the `else` clause won't ever be hit?

Google.PowerShell/PubSub/GcpsTopic.cs
```diff
@@ -0,0 +1,234 @@
...
+        public string Project { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The names of the topics to be removed.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, Position = 0, ValueFromPipelineByPropertyName = true)]
+        [Alias("Name")]
+        public string[] Topic { get; set; }
```
chrsmith: This should be `ValidateNotNullOrEmpty` right?

Google.PowerShell.IntegrationTests/PubSub/PubSub.Tests.ps1
```diff
@@ -0,0 +1,101 @@
...
+
+        try {
+            New-GcpsTopic -Topic $topicName
+            (Get-GcpsTopic -Topic $topicName).Name | Should BeExactly "projects/$project/topics/$topicName"
+            { New-GcpsTopic -Topic $topicName -ErrorAction Stop } | Should Throw "already exists" 
+        }
+        finally {
+            gcloud beta pubsub topics delete $topicName --quiet 2>$null
+        }
+    }
```
chrsmith: It seems like we should have a few more unit tests, just to confirm the behavior is reasonable. (For this and other cmdlets.) For example, what is the error you get if you use an invalid topic name? Do we crash with an obtuse error, or is it reasonable?

quoctruong: Added more tests.

Google.PowerShell/PubSub/GcpsCmdlet.cs
quoctruong: I'll submit that in a separate PR.

Google.PowerShell/PubSub/GcpsCmdlet.cs
quoctruong: Upon further testing, it seems "_deleted-topic_" is not recognized by the PubSub API at all so I removed it.

Google.PowerShell/PubSub/GcpsCmdlet.cs
quoctruong: Renamed.

Google.PowerShell/PubSub/GcpsTopic.cs
quoctruong: Raise errors.

Google.PowerShell/PubSub/GcpsTopic.cs
quoctruong: Added helper functions.

Google.PowerShell/PubSub/GcpsTopic.cs
quoctruong: According to the [convention](https://msdn.microsoft.com/en-us/library/dd878270(v=vs.85).aspx), we should avoid plural names for a parameter unless the value of the parameter is always a multiple-element value.

Google.PowerShell/PubSub/GcpsTopic.cs
quoctruong: This is not true. If the user does not use -Topic parameter at all, then the `else` clause will still be hit. The `ValidateNotNullOrEmpty` attribute will only take effect if the user do something like -Topic $emptyArray.

Google.PowerShell/PubSub/GcpsTopic.cs
quoctruong: The `getResponse` actually is not a list so it does not contain `Topics`. The ListResponse is the one that contains `Topics`.

Google.PowerShell/PubSub/GcpsTopic.cs
quoctruong: Removed

Google.PowerShell/PubSub/GcpsTopic.cs
quoctruong: Added.

Google.PowerShell/PubSub/GcpsTopic.cs
quoctruong: Same reasoning as above.

Google.PowerShell/PubSub/GcpsTopic.cs
quoctruong: Added.

Google.PowerShell/Common/GcloudCmdlet.cs
```diff
@@ -99,6 +99,40 @@ protected string GetFullPath(string filePath)
         }
 
         /// <summary>
+        /// Converts a hashtable to a dictionary
```
iantalarico: Can you elaborate this comment?

Google.PowerShell/PubSub/GcpsCmdlet.cs
iantalarico: Why not just add them now?

Google.PowerShell/PubSub/GcpsTopic.cs
```diff
@@ -0,0 +1,231 @@
...
+        /// default to the Cloud SDK's DefaultProject property.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        [ConfigPropertyName(CloudSdkSettings.CommonProperties.Project)]
+        public string Project { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The names of the topics to be created. Topics must not have already existed.
```
iantalarico: "...must not have already existed." Make me think it cannot ever has existed.  Is that correct?  If so then I would be more specific otherwise I would use ""...must not exist.""

quoctruong:
@iantalarico PTAL. I didn't add the copyright headers yet because it will affect other files not in this PR.

iantalarico:
Okay SGTM, just be sure to do that soon

quoctruong:
Thanks! I'll submit another PR today for that.

PR #400, opened 2016-12-15
Switch to using "gcloud config config-helper" instead of "gcloud auth print-access-token"

quoctruong:
Some context:
Currently, we are using "gcloud auth print-access-token" to get an access token from the user. The problem with this is that internally, we have to keep track of when there is a change in the config and call this again. This is very complicated and does not take into account internal releases version of Gcloud.

The Cloud SDK team introduces "gcloud config config-helper" which will print out the current config, access token as well as a sentinel file. This sentinel file will be changed whenever there is a change in the current gcloud configuration. However, it does not take into account changes in CLOUDSDK_* environment variables. Hence, we have to keep track of the timestamp of the sentinel file as well as CLOUDSDK_* environment variables to detect a change in the active config. This, however, will be a lot more accurate and easier to maintain than the previous approach.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/400)


Google.PowerShell/Common/ActiveUserConfig.cs
```diff
@@ -0,0 +1,211 @@
...
+        /// </summary>
+        internal ActiveUserConfig(string activeConfigJson)
+        {
+            JToken parsedConfigJson = JObject.Parse(activeConfigJson);
+
+            // Parse the sentinels section to get the sentinel file.
+            JToken sentinelJson = parsedConfigJson.SelectToken("sentinels.config_sentinel");
+            if (sentinelJson == null)
+            {
+                throw new FileNotFoundException("Sentinel file for current active configuration could not be found.");
```
iantalarico: Is this the correct exception? This is more invalid json or param right?

Google.PowerShell/Common/ActiveUserConfig.cs
```diff
@@ -0,0 +1,211 @@
...
+            }
+            SentinelFile = sentinelJson.Value<string>();
+            CachedFingerPrint = GetCurrentConfigurationFingerPrint();
+
+            // Parse the credential section to get the access token.
+            JToken parsedCredentialJson = parsedConfigJson.SelectToken("credential");
```
iantalarico: Can this be null?

Google.PowerShell/Common/ActiveUserConfig.cs
```diff
@@ -0,0 +1,211 @@
...
+            UserToken = new TokenResponse(parsedCredentialJson);
+
+            // Parse the properties section to get properties of the current configuration.
+            PropertiesJson = parsedConfigJson.SelectToken("configuration.properties");
```
iantalarico: Can this be null?

Google.PowerShell/Common/ActiveUserConfig.cs
```diff
@@ -0,0 +1,211 @@
...
+            }
+            return result;
+        }
+
+        /// <summary>
+        /// Search the JToken and its children recursively for a key that matches the given key.
+        /// If such a key is found, set the value to the ref variable value and returns true.
+        /// Otherwise, returns false.
+        /// </summary>
+        internal static bool TryGetPropertyValue(JToken propertiesJson, string key, ref string value)
```
iantalarico: Maybe make this an extension?

Google.PowerShell/Common/JsonUtils.cs
```diff
@@ -0,0 +1,53 @@
...
+                // the same name as key, then we are done. Otherwise, recursively call this method
+                // with the child token to continue the search.
+                foreach (JProperty childProperty in propertiesJson.Children<JProperty>())
+                {
+                    if (string.Equals(childProperty.Name, key, StringComparison.OrdinalIgnoreCase))
+                    {
+                        value = childProperty.Value?.Value<string>();
+                        return true;
+                    }
+                    if (TryGetPropertyValue(childProperty.Value, key, ref value))
```
iantalarico: Can't this just be childProperty.Value.TryGetPropertyValue?

Google.PowerShell/Common/JsonUtils.cs
```diff
@@ -0,0 +1,53 @@
+﻿using Newtonsoft.Json.Linq;
+using System;
+
+namespace Google.PowerShell.Common
+{
+    /// <summary>
+    /// Static class that contains extension static to manipulate JToken.
+    /// </summary>
+    public static class JsonUtils
```
iantalarico: Maybe JsonExtensions? Up to you though

Google.PowerShell/Common/JsonUtils.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            if (propertiesJson == null)
+            {
+                return false;
+            }
+            if (propertiesJson.Type == JTokenType.Array)
+            {
+                foreach (JToken childToken in propertiesJson.Children())
+                {
+                    if (TryGetPropertyValue(childToken, key, ref value))
```
iantalarico: Can't this just be childToken.TryGetPropertyValue(?

Google.PowerShell.Tests/Common/TestActiveUserConfig.cs
```diff
@@ -62,16 +62,19 @@ public void TestTryGetPropertyValue()
...
+            Assert.IsTrue(success, "Failed to get property region from config JSON.");
+            Assert.AreEqual(zone, result, "TryGetPropertyValue returns wrong zone property from config JSON.");
 
             string project = "gcloud-powershell-testing";
-            ActiveUserConfig.TryGetPropertyValue(userConfig.PropertiesJson, "project", ref result);
-            Assert.AreEqual(project, result, "Failed to get property project from config JSON.");
+            success = userConfig.PropertiesJson.TryGetPropertyValue("project", ref result);
+            Assert.IsTrue(success, "Failed to get property region from config JSON.");
+            Assert.AreEqual(project, result, "TryGetPropertyValue returns wrong project property from config JSON.");
         }
```
iantalarico: Should you add a test for TryGetPropertyValue returning false?

quoctruong:
@iantalarico PTAL

quoctruong:
Thanks!

PR #411, opened 2017-01-04
Fix a bug where -RemoveDisk in Set-GceInstance cmdlet does not work with Disk object

quoctruong:
This is because we are using the Name property from the Disk object. However, this property gives us the persistent name of the Disk whereas what we need is the DeviceName of the disk (what the VM sees the Disk as).

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/411)


quoctruong:
Thanks!

PR #418, opened 2017-01-10
Change Get-GcLogEntry so -Filter will work with other parameters

quoctruong:
Also factor out the logic for constructing filter so we can reuse it in other log cmdlets.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/418)


quoctruong:
Thanks!

PR #428, opened 2017-01-18
Add integration between New-GcsObject and the provider

quoctruong:
With this change, New-GcsObject will try to get the context of the current PowerShell location if it is inside a Google Cloud Storage Provider location. The cmdlet can then automatically fill out -Bucket as well as apply prefix to the object created.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/428)

### DIFF
--- a/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
@@ -23,9 +23,9 @@ Describe "New-GcsObject" {
         $filename = [System.IO.Path]::GetTempFileName()
         "Hello, World" | Out-File $filename -Encoding utf8
 
-        $objectName = "43b75bdd-8869-496e-8c0d-3c12b49dcb18.txt"
+        $objectName = Get-Random
 
-        $newObj = New-GcsObject $objectName -File $filename $bucket
+        $newObj = New-GcsObject $bucket $objectName -File $filename
         Remove-Item $filename
 
         $newObj.Name | Should Be $objectName
@@ -44,12 +44,12 @@ Describe "New-GcsObject" {
         $filename = [System.IO.Path]::GetTempFileName()
         "Hello, World" | Out-File $filename -Encoding utf8
 
-        $objectName = "43b75bdd-8869-496e-8c0d-3c12b49dcb18.txt"
+        $objectName = Get-Random
 
         $currentLocation = Resolve-Path .\
         try {
             cd "gs:\$bucket"
-            $newObj = New-GcsObject $objectName -File $filename
+            $newObj = New-GcsObject -ObjectName $objectName -File $filename
             Remove-Item $filename
 
             $newObj.Name | Should Be $objectName
@@ -96,7 +96,7 @@ Describe "New-GcsObject" {
         $filename = [System.IO.Path]::GetTempFileName()
         "Hello, World" | Out-File $filename -Encoding utf8
 
-        $objectName = "43b75bdd-8869-496e-8c0d-3c12b49dcb18.txt"
+        $objectName = Get-Random
 
         $currentLocation = Resolve-Path .\
         $folderName = "gcps-testing"
@@ -104,7 +104,7 @@ Describe "New-GcsObject" {
             cd "gs:\$bucket"
             mkdir $folderName
             cd $folderName
-            $newObj = New-GcsObject $objectName -File $filename
+            $newObj = New-GcsObject -ObjectName $objectName -File $filename
             Remove-Item $filename
 
             $newObj.Name | Should Be "$folderName/$objectName"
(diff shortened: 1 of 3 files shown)


Google.PowerShell/Common/GcloudCmdlet.cs
```diff
@@ -53,6 +53,34 @@ public static BaseClientService.Initializer GetBaseClientServiceInitializer()
             };
         }
 
+        protected Tuple<string, ProviderInfo> GetFullPath(string filePath)
```
iantalarico: Can you add a comment for this?

Google.PowerShell/Common/GcloudCmdlet.cs
```diff
@@ -53,6 +53,34 @@ public static BaseClientService.Initializer GetBaseClientServiceInitializer()
...
+                    return new Tuple<string, ProviderInfo>(result[0], provider);
+                }
+            }
+            catch (ItemNotFoundException itemEx)
+            {
+                // This exception is thrown if the path does not exist.
+                // We can still resolve the path with GetUnresolvedProviderPathFromPSPath.
+                ProviderInfo provider = null;
+                PSDriveInfo psDrive = null;
+                string path = SessionState.Path.GetUnresolvedProviderPathFromPSPath(itemEx.ItemName, out provider, out psDrive);
```
iantalarico: This is a little confusing.  The path does not exist but we can still find it?

Can you clarify what is happening here?

iantalarico: Can path be null?

iantalarico: Can GetUnresolvedProviderPathFromPSPath throw?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -149,9 +163,10 @@ private class ParameterSetNames
         /// The name of the bucket to upload to. Will also accept a Bucket object.
         /// </para>
         /// </summary>
-        [Parameter(Position = 0, Mandatory = true)]
+        [Parameter(Position = 2, Mandatory = false)]
```
iantalarico: Is this cmdlet GA? Do the positional and mandatory changes need to happen in a new version as to not break users?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -258,12 +273,35 @@ protected override void ProcessRecord()
             string objContentType = null;
             Stream contentStream = null;
 
+            // Check whether our current location is in gs:\ (i.e., we are in the Google Cloud Storage provider).
+            // If so, then we can resolve the path to get the bucket and folder name (if we are in one).
+            if (SessionState?.Path?.CurrentLocation?.Provider?.ImplementingType == typeof(GoogleCloudStorageProvider))
```
iantalarico: Might be worth checking if bucket and ObjectNamePrefix are set already to avoid doing unneeded work

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
```diff
@@ -40,20 +40,127 @@ Describe "New-GcsObject" {
         $obj.Metadata | Should Be $null
     }
 
+    It "should upload a file content with GCS Provider" {
+        $filename = [System.IO.Path]::GetTempFileName()
+        "Hello, World" | Out-File $filename -Encoding utf8
+
+        $objectName = "43b75bdd-8869-496e-8c0d-3c12b49dcb18.txt"
```
iantalarico: Should this be different each test?

Same below

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
```diff
@@ -40,20 +40,127 @@ Describe "New-GcsObject" {
...
+
+        $currentLocation = Resolve-Path .\
+        try {
+            cd "gs:\$bucket"
+            $newObj = New-GcsObject $objectName -File $filename
+            Remove-Item $filename
+
+            $newObj.Name | Should Be $objectName
+            $newObj.Size | Should Be 17
```
iantalarico: Will this always be 17? even on different machines? OS? 

Same below

Google.PowerShell/Storage/GcsObject.cs
quoctruong: We have to check the possible prefix to append to the front of ObjectNamePrefix anyway.

Google.PowerShell/Storage/GcsObject.cs
quoctruong: The mandatory changes won't break users. I've changed the position back to 0.

Google.PowerShell/Common/GcloudCmdlet.cs
quoctruong: It can throw but we'll just let the caller handle that.

quoctruong:
Thanks!

PR #432, opened 2017-01-24
Fix comments for PredefinedAcl

quoctruong:
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/432)

### DIFF
--- a/Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
+++ b/Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
@@ -134,11 +134,18 @@ namespace Google.PowerShell.CloudStorage
 
             /// <summary>
             /// <para type="description">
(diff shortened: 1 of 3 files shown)


Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
```diff
@@ -134,23 +134,37 @@ public class NewGcsBucketDynamicParameters
 
             /// <summary>
             /// <para type="description">
-            /// Default ACL for the bucket. e.g. "publicRead", "private", etc.
+            /// Default ACL for the bucket.
```
iantalarico: Not required but I find both of these are a little hard to parse. It would be nice if enum values and "*** permission" were quoted.

quoctruong:
Thanks!

PR #433, opened 2017-01-25
Fix an issue where dynamic parameter is not generated for website.

quoctruong:
Dynamic Parameter is not generated for the website because they don't show up in Get-Help. They do, however, show up in Get-Command.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/433)

### DIFF
--- a/Tools/GenerateWebsiteData.ps1
+++ b/Tools/GenerateWebsiteData.ps1
@@ -8,6 +8,8 @@ Param(
     [string]$configuration = "Release"
 )
 
+$script:defaultParameterSetName = "Default"
+
 # Unload the module if already loaded. (Weird things happen when debugging...)
 if (Get-Module Google.PowerShell) {
     Write-Warning "Previous version of Google.PowerShell loaded. Unloadig the module."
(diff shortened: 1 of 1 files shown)


Tools/GenerateWebsiteData.ps1
```diff
@@ -8,6 +8,8 @@ Param(
     [string]$configuration = "Release"
 )
 
+$script:defaultParameterSetName = "Default"
```
chrsmith: Do we need this? It only seems to be used in one place, and there it is just read?

Tools/GenerateWebsiteData.ps1
```diff
@@ -77,6 +79,58 @@ function GetLinks ($docObj) {
     return $null
 }
 
+# Generate a PSObject that will be used to populate the cmdlet JSON file based on $parameter.
```
chrsmith: Minor nit: it would be nice to know the type of `$parameter`. e.g. "based on $parameter (WidgetSproketFactory)."

Tools/GenerateWebsiteData.ps1
```diff
@@ -77,6 +79,58 @@ function GetLinks ($docObj) {
...
+function New-DynamicParamObject($parameter) {
+    if ($parameter.Position -ge 0) {
```
chrsmith: This is a matter of style, but personally I prefer to declare the variable with a default value and then change it after the fact. Rather than set its value inside the if-body, and then use it outside of the if body. Because the way it is used now it isn't 100% clear where $position is declared. (Since it is never declared, just assigned to.)

Tools/GenerateWebsiteData.ps1
```diff
@@ -147,9 +201,21 @@ function AnnotateParametersAndSyntaxObjects($cmdletName, $docObj) {
             # the parameter set are honored. It is possible to have one parameter set be a proper
             # subset of another.
             ForEach ($paramSetParam in $parameterSet.Parameters) {
+                # If the parameter is a dynamic parameter, then it won't show up in Get-Help -Full.
+                # So we will have to generate an object that corresponds to this dynamic parameter and add it to $syntaxObj.parameter.
+                if ($paramSetParam.IsDynamic) {
+                    $dynamicParameterSyntaxObject = New-DynamicParamObject $paramSetParam
+
```
chrsmith: For readability, I'd remove line 208 and 210. The comment on line 204 does a good job explaining things, and a 4-line if statement body is pretty legible IMHO.

Tools/GenerateWebsiteData.ps1
```diff
@@ -77,6 +79,58 @@ function GetLinks ($docObj) {
     return $null
 }
 
+# Generate a PSObject that will be used to populate the cmdlet JSON file based on $parameter.
+function New-DynamicParamObject($parameter) {
```
chrsmith: Is this function name consistent with the rest of the code? Later we define `AddDynamicParameterToDocObj`. Should that instead be `Add-DynamicParameterToDocObj`?

Tools/GenerateWebsiteData.ps1
```diff
@@ -195,6 +288,8 @@ $productInfoLookup = @{
     "Gce"   = @{ name = "Google Compute Engine"; shortName = "google-compute-engine"; resources = @() }
     "GcSql" = @{ name = "Google Cloud SQL";      shortName = "google-cloud-sql"     ; resources = @() }
     "Gcd"   = @{ name = "Google Cloud DNS";      shortName = "google-cloud-dns"     ; resources = @() }
+    "Gcps"   = @{ name = "Google Cloud PubSub";      shortName = "google-cloud-pubsub"     ; resources = @() }
```
chrsmith: In a perfect world these two changes would be in a separate CL so that the change description would match the code being updated. However, don't worry about it.

quoctruong:
Thanks!

Issue #438, opened 2017-01-27
Cross-Platform bits not working on OS X


quoctruong:
We get an error like this:

get-gcsbucket : One or more errors occurred. (Could not find file 
'/Users/quoct/.config/gcloud/config_sentinel'.)
At line:1 char:1
+ get-gcsbucket
+ ~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [], AggregateException
    + FullyQualifiedErrorId : System.AggregateException

This is most likely because the GetLastWriteTime throws a FileNotFoundException in OS X but not in Windows when the file does not exist.

chrsmith:
While we should handle this edge case, we should also log a bug on the .NET
Core project if this is an inconsistency in the .NET APIs.

quoctruong:
This issue is fixed with https://github.com/GoogleCloudPlatform/google-cloud-powershell/pull/440. I've opened a bug on the .NET repo at https://github.com/dotnet/corefx/issues/15633. However, it seems that they intentionally wanted to throw on Linux (see https://github.com/dotnet/corefx/issues/1728)

PR #440, opened 2017-01-30
Fix a bug where the initialization throws error when sentinel file does not exist.

quoctruong:
This is to address https://github.com/GoogleCloudPlatform/google-cloud-powershell/issues/438

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/440)


quoctruong:
Thanks!

PR #450, opened 2017-02-02
Report project number in telemetry

quoctruong:
The main logic is in this file https://github.com/GoogleCloudPlatform/google-cloud-powershell/pull/450/files#diff-9b70d977cd9295395d28146529744e1f.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/450)

### DIFF
--- a/Google.PowerShell/Common/GcloudCmdlet.cs
+++ b/Google.PowerShell/Common/GcloudCmdlet.cs
@@ -25,6 +25,11 @@ namespace Google.PowerShell.Common
         /// <summary>Placeholder for an unknown cmdlet name when reporting telemetry.</summary>
         private const string UnknownCmdletName = "unknown-cmdlet";
 
+        /// <summary>
+        /// The project that is used by the cmdlet. This value will be used in reporting usage.
+        /// </summary>
+        public virtual string Project { get; set; }
+
         public GCloudCmdlet()
         {
             if (CloudSdkSettings.GetOptIntoUsageReporting())
@@ -270,7 +275,7 @@ namespace Google.PowerShell.Common
 
             if (_cmdletInvocationSuccessful)
             {
-                _telemetryReporter.ReportSuccess(cmdletName, parameterSet);
+                _telemetryReporter.ReportSuccess(cmdletName, parameterSet, Project);
             }
             else
             {
@@ -278,7 +283,7 @@ namespace Google.PowerShell.Common
                 // cmdlet threw? If so, use that to determine a more appropriate error code.
                 // We report 1 instead of 0 so that the data can be see in Google Analytics.
                 // (null vs. 0 is ambiguous in the UI.)
-                _telemetryReporter.ReportFailure(cmdletName, parameterSet, 1);
+                _telemetryReporter.ReportFailure(cmdletName, parameterSet, Project, 1);
             }
         }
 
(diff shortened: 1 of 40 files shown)


Google.PowerShell/Common/GcloudCmdlet.cs
```diff
@@ -25,6 +25,11 @@ public abstract class GCloudCmdlet : PSCmdlet, IDisposable
         /// <summary>Placeholder for an unknown cmdlet name when reporting telemetry.</summary>
         private const string UnknownCmdletName = "unknown-cmdlet";
 
+        /// <summary>
+        /// The project that is used by the cmdlet. This value will be used in reporting usage.
+        /// </summary>
+        public virtual string Project { get; set; }
```
chrsmith: It is worth calling out a few other things:

- Can it be ignored? What happens if the derived class leaves it as null? What value will be reported to telemetry? Will we fall back to whatever is in the Cloud SDK "active project" value?

- We should explicitly say that "will be converted into hash, and then used in reporting usage". i.e. we want to make clear we don't send project names to Google, and rather anonymized versions of the data. (I believe that is the expected format in the telemetry reporting.)

Google.PowerShell/Common/ReportCmdletResults.cs
```diff
@@ -174,32 +180,33 @@ public GoogleAnalyticsCmdletReporter(string clientID, AnalyticsEventCategory ana
...
+                "projectNumber", (projectNumber ?? "null"),
                 "errorCode", (errorCode.HasValue ? errorCode.ToString() : "null"));
 
             s_reporter.Value.ReportEvent(
                 source: "virtual.powershell",
                 eventType: "powershell",
                 eventName: eventData.Name,
                 userLoggedIn: true,
-                projectNumber: null,
+                projectNumber: projectNumber,
```
chrsmith: Something seems wrong here. Projects can be identified both by name ("bobs-taco-truck") and also by a number (29576092874). I assume we need to do some sort of conversion between the two?

Alternatively, should we just return `projectName.GetHashCode` instead? i.e. an anonymous variant of the project name?

@ivannaranjo , can you chime in here? We aren't reporting project info in our telemetry. What is the right way to wire this in?

quoctruong: We actually hash the project name before reporting it.  I can update the parameter to projectName instead of projectNumber to accurately reflect that. Will that be sufficient?

ivannaranjo: We need to report the hash of the project number, that is how the backend expects it to be.

quoctruong:
@chrsmith PTAL

Google.PowerShell/Common/ReportCmdletResults.cs
chrsmith: @ivannaranjo just to make sure I have that right, _hash of the project number_? e.g. `"96463850128".GetHashCode()`?

I assume if we don't do that exactly, and instead just hash the project name, that will be OK. Right? Since on our backend logs the entire reason we hash things is so that it is anonymized, and if the anonymization method we use isn't compatible, that only breaks cases where you are joining different data streams. Something that shouldn't be done anyways, right?

@quoctruong As long as the code is consistent (variable names and comments) around projectName vs. projectNumber vs. projectXXHash, I'm happy with this PR.

quoctruong: We actually have a function to produce the hash [here ](https://github.com/GoogleCloudPlatform/google-cloud-powershell/blob/77fb00fd3ab621dcee85e72c67c44649d26ca5aa/Google.PowerShell/Common/AnalyticsReporter.cs#L444). This uses the same logic that the visual studio one is using.

chrsmith:
LGTM :shipit:

quoctruong:
Thanks!

PR #452, opened 2017-02-08
Fix formating for get-gceinstance

quoctruong:
Also, separate the format for bucket and object (because bucket does not have ContentType and Size properties.
image5

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/452)

### DIFF
--- a/Google.PowerShell/ReleaseFiles/GoogleCloudPlatform.Format.ps1xml
+++ b/Google.PowerShell/ReleaseFiles/GoogleCloudPlatform.Format.ps1xml
@@ -1,18 +1,45 @@
 ﻿<Configuration>
-  <SelectionSets>
-    <SelectionSet>
-      <Name>GoogleCloudStorageTypes</Name>
-      <Types>
-        <TypeName>Google.Apis.Storage.v1.Data.Bucket</TypeName>
-        <TypeName>Google.Apis.Storage.v1.Data.Object</TypeName>
-      </Types>
-    </SelectionSet>
-  </SelectionSets>
   <ViewDefinitions>
     <View>
-      <Name>Google Cloud Storage Table</Name>
+      <Name>Google Cloud Storage Bucket Table</Name>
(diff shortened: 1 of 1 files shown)


Google.PowerShell/ReleaseFiles/GoogleCloudPlatform.Format.ps1xml
```diff
@@ -53,5 +80,63 @@
...
+                    $match = $reg.Match($_.Zone);
+                    if ($match.Success) {
+                        $match.Groups[1].Value;
+                    } else {
+                        $_.Zone;
+                    }
+                </ScriptBlock>
+              </TableColumnItem>
+              <TableColumnItem>
+                <ScriptBlock>$_.CreationTimestamp</ScriptBlock>
```
ILMTitan: Is there a reason to prefer ScriptBlock over PropertyName?

Google.PowerShell/ReleaseFiles/GoogleCloudPlatform.Format.ps1xml
```diff
@@ -53,5 +80,63 @@
...
+              </TableColumnItem>
+              <TableColumnItem>
+                <PropertyName>CpuPlatform</PropertyName>
+              </TableColumnItem>
+              <TableColumnItem>
+                <ScriptBlock>
+                    $reg = [Regex]"https:.*/zones/[^/]*/machineTypes/([^/]*)";
+                    $match = $reg.Match($_.MachineType);
+                    if ($match.Success) {
+                        $match.Groups[1].Value;
```
ILMTitan: Instead of using `[Regex]`, I think you could use 
```
if($_.MachineType -match $reg) {
    $Matches[1]
 } else ...
```

quoctruong:
Fixed the file based on the suggestion. Thanks!

PR #453, opened 2017-02-08
Initial commit for project cmdlet

ahandley:
Initial commit of Get-GcbqProject cmdlet and integration tests for BigQuery.  

One comment is that I was not able to get as much automated test coverage as I would like for this cmdlet.  I looked into the Pantheon test user form page, but it was talking specifically about javascript, and I could not see a way to access that from this setting.  If anyone has ideas or knowledge of how to make a test user and programmatically add and remove projects (to test the different possible returns from the list API), please let me know and I'll add that in.  This shouldn't be an issue under any of the other cmdlet sections, because those have full API control support for integration testing.


Google.PowerShell.IntegrationTests/packages.config
```diff
@@ -0,0 +1,10 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Google.Apis" version="1.20.0" targetFramework="net452" />
+  <package id="Google.Apis.Auth" version="1.20.0" targetFramework="net452" />
+  <package id="Google.Apis.Bigquery.v2" version="1.20.0.744" targetFramework="net452" />
+  <package id="Google.Apis.Core" version="1.20.0" targetFramework="net452" />
+  <package id="log4net" version="2.0.3" targetFramework="net452" />
+  <package id="Newtonsoft.Json" version="7.0.1" targetFramework="net452" />
+  <package id="Zlib.Portable.Signed" version="1.11.0" targetFramework="net452" />
+</packages>
```
quoctruong: Add a line at the end of this file.

Google.PowerShell/BigQuery/GcbqCmdlet.cs
```diff
@@ -0,0 +1,21 @@
+﻿// Copyright 2015-2016 Google Inc. All Rights Reserved.
+// Licensed under the Apache License Version 2.0.
+
+using Google.Apis.Bigquery.v2;
+using Google.PowerShell.Common;
+
+namespace Google.PowerShell.BigQuery
+{
+    /// <summary>
+    /// Base class for Google Cloud PubSub cmdlets.
```
quoctruong: Should be "Google Cloud BigQuery".

Google.PowerShell/BigQuery/GcbqProject.cs
```diff
@@ -0,0 +1,77 @@
...
+        public uint? MaxResults { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// Page token, returned by a previous call, to request the next page of results.
+        /// </para>
+        /// </summary>
+        [Alias("pt")]
+        [Parameter(Mandatory = false)]
+        public string PageToken { get; set; }
```
quoctruong: I don't think you need PageToken and MaxResult for this cmdlet. Most likely users won't be using these parameters.

Google.PowerShell.IntegrationTests/BigQuery/GcbqCmdlets.ps1
```diff
@@ -0,0 +1,16 @@
+﻿. $PSScriptRoot\..\GcloudCmdlets.ps1
```
quoctruong: Do we need this file? The Test-Funct does not seem to be used anywhere.

Google.PowerShell/BigQuery/GcbqProject.cs
```diff
@@ -0,0 +1,77 @@
...
+    public class GetGcbqProject : GcbqCmdlet
+    {
+        /// <summary>
+        /// <para type="description">
+        /// The maximum number of results to return.
+        /// </para>
+        /// </summary>
+        [Alias("mr")]
+        [Parameter(Mandatory = false)]
+        public uint? MaxResults { get; set; }
```
ILMTitan: Don't use this. As written, it will only make the request page size smaller, requiring more requests but iterating all the projects nevertheless. In the rare case that a user does want to limit the number of returned projects, they can pipe this Cmdlet to `Select-Object` like so:
```
Get-GcbqProject | Select-Object -First 100
```
That will cause `WriteObject` to throw an exception once 100 projects have been returned, short circuiting further requests. In an interactive environment, `Ctrl-C` will set `Stopping` to `true`.

Google.PowerShell.IntegrationTests/BigQuery/GcbqCmdlets.ps1
ahandley: I can take out the Test-Funct, since that was just a reminder to myself about syntax.  As for the file, I saw that pattern being used to define constants and functions in between the different sets of cmdlet tests in other sections.  It should start to become useful for future sets, such as tables and datasets.  Is it better to leave it until then or take it out now and include it later when it's explicitly needed?

Google.PowerShell/BigQuery/GcbqCmdlet.cs
ahandley: Just noticed there's a problem with the comments there.  Fixed that.  

Was your comment talking about the using statement?

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqProject.Tests.ps1
```diff
@@ -0,0 +1,24 @@
+﻿. $PSScriptRoot\..\BigQuery\GcbqCmdlets.ps1
+$project, $configName = Set-GCloudConfig
```
ILMTitan: This should be 
`$project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig`
or 
`$_, $_, $oldActiveConfig, $configName = Set-GCloudConfig`
if you are never going to care about the default project or zone.

Google.PowerShell/BigQuery/GcbqProject.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// Lists all projects to which you have been granted any project role.
+    /// </para>
+    /// <para type="description">
+    /// Queries to find which projects you have been granted any project role in and returns them in the form of ProjectsData objects.
+    /// The flag -MaxResults can be used to specify the maximum number of results retrieved at once, and -PageToken can be used to
+    /// specify an offset into the project list at which to start.
+    /// </para>
+    /// <example>
+    ///   <code>PS C:\> Get-GcbqProjects -MaxResults 20</code>
+    ///   <para>This command will list up to the first 20 BigQuery projects that you've been granted any roles for.</para>
```
iantalarico: I believe these are Google Cloud Console projects.  I don't think BigQuery has the notion itself of projects.

Google.PowerShell/BigQuery/GcbqProject.cs
```diff
@@ -0,0 +1,77 @@
...
+    /// </example>
+    /// <example>
+    ///   <code>PS C:\> Get-GcbqProjects -MaxResults 20 -PageToken 20</code>
+    ///   <para>This command will list the second page of up to 20 BigQuery projects that you've been granted any roles for.</para>
```
iantalarico: Same here

Google.PowerShell/BigQuery/GcbqProject.cs
```diff
@@ -0,0 +1,77 @@
...
+        protected override void ProcessRecord()
+        {
+            //check for valid value of MaxResults
+            if (MaxResults <= 0)
+            {
+                ThrowTerminatingError(new ErrorRecord(new Exception("MaxResults only takes positive numbers."), 
+                    "400", ErrorCategory.InvalidArgument, MaxResults));
+            }
+
+            //create request with flag results.  nulls should be handled
```
iantalarico: Nit: Cap Create and add space after //

Full sentence for the second half too

Google.PowerShell/BigQuery/GcbqProject.cs
```diff
@@ -0,0 +1,77 @@
...
+            ProjectsResource.ListRequest request = Service.Projects.List();
+            request.MaxResults = MaxResults;
+            request.PageToken = PageToken;
+
+            //send request and parse response
```
iantalarico: Same here

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqProject.Tests.ps1
```diff
@@ -0,0 +1,21 @@
+﻿. $PSScriptRoot\..\BigQuery\GcbqCmdlets.ps1
+$project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
+
+Describe "Get-GcbqProject" {
+
+	It "should reject non-positive values of maxResults" {
```
ILMTitan: Remove these tests parameters that no longer exist.

PR #462, opened 2017-02-14
New-GcbqDataset cmdlet implemented

ahandley:
Impemented the cmdlet and testing.  Word of warning, I haven't written remove-dataset, so the datasets must be removed manually from the testing project in between runs of the integration testing.  

Specific request from this code review:  Look at the way I implemented the parameters for this cmdlet and let me know if I'm going in the wrong direction with this.

### DIFF
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
@@ -1,0 +1,85 @@
+﻿. $PSScriptRoot\..\BigQuery\GcbqCmdlets.ps1
+$project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
+
+Describe "New-GcbqDataset" {
+
+	It "should take a name, description, and time to make a dataset" {
+		$data = New-GcbqDataset "test_data_id1" -Name "Testdata" -Description "Some interesting data!" -Timeout 86400000
+		$data | Should Not BeNullOrEmpty
+		$data.DatasetReference.DatasetId | Should Be "test_data_id1"
+		$data.FriendlyName | Should Be "Testdata"
+		$data.Description | Should Be "Some interesting data!"
+		$data.DefaultTableExpirationMs | Should Be 86400000
+    }
+
+	It "should accept a dataset object from pipeline" {
+		$data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
+		$data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
+        $data.DatasetReference.DatasetId = "test_data_id2"
+        $data.DatasetReference.ProjectId = $project
+		$data.FriendlyName = "PipeTest"
+		$data.Description = "Some cool data coming hot off the pipeline!"
+		$data | New-GcbqDataset -outvariable newdata
+		$newdata | Should Not BeNullOrEmpty
+		$newdata.DatasetReference.DatasetId | Should Be "test_data_id2"
+		$newdata.FriendlyName | Should Be "PipeTest"
+    }
+
+	It "should accept a dataset object from a command line argument" {
+		$data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
+		$data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
+        $data.DatasetReference.DatasetId = "test_data_id3"
+        $data.DatasetReference.ProjectId = $project
+		$data.FriendlyName = "PipeTest"
+		$data.Description = "Some cool data coming hot off the pipeline!"
+		$newdata = New-GcbqDataset -Dataset $data
+		$newdata | Should Not BeNullOrEmpty
+		$newdata.DatasetReference.DatasetId | Should Be "test_data_id3"
+		$newdata.FriendlyName | Should Be "PipeTest"
+    }
+
+	It "should not work with no arguments (meaning no datasetId)" {
+		$data = New-GcbqDataset | Should Throw
+    }
+
+	It "should reject datasets that do not have unique ids" {
+		$data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
+		$data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
+        $data.DatasetReference.DatasetId = "test_data_id6"
+		$data2 = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
+		$data2.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
+        $data2.DatasetReference.DatasetId = "test_data_id6"
+		$data | New-GcbqDataset
+		{ $data2 | New-GcbqDataset } | Should Throw
+    }
+
+	It "should use the -Project tag correctly" {
+		$data = New-GcbqDataset -Project $project "test_data_id7" -Name "Testdata" -Description "Some interesting data!"
+		$data | Should Not BeNullOrEmpty
+		$data.DatasetReference.ProjectId | Should Be $project
+    }
+
+	It "should throw when you try to add to a project that doesnt exist" {
+		{ New-GcbqDataset -Project $nonExistProject "test_data_id8" -Name "Testdata" -Description "Some interesting data!" } | Should Throw
+    }
+
+	It "should throw when you try to add to a project that is not yours" {
+		{ New-GcbqDataset -Project $accessErrProject "test_data_id9" -Name "Testdata" -Description "Some interesting data!" } | Should Throw
+    }
+
+	#add afterall block that deletes test datasets when remove-x command is done
+}
+
+Describe "Get-GcbqDataset" {
+
+}
+
+Describe "Set-GcbqDataset" {
+
+}
+
+Describe "Remove-GcbqDataset" {
+
+}
+
+Reset-GCloudConfig $oldActiveConfig $configName
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
@@ -39,7 +39,15 @@ Describe "New-GcbqDataset" {
 	}
 
 	It "should not work with no arguments (meaning no datasetId)" {
-		$data = New-GcbqDataset | Should Throw
+		{ New-GcbqDataset } | Should Throw 400
+	}
+
+	It "should reject datasets with malformed ids" {
+		{ New-GcbqDataset "test-?ata-4" } | Should Throw 400
+	}
+
+	It "should reject datasets with empty ids" {
+		{ New-GcbqDataset "" } | Should Throw "Cannot validate argument on parameter 'DatasetId'"
 	}
 
 	It "should reject datasets that do not have unique ids" {
@@ -50,7 +58,7 @@ Describe "New-GcbqDataset" {
 		$data2.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
 		$data2.DatasetReference.DatasetId = "test_data_id6"
 		$data | New-GcbqDataset
-		{ $data2 | New-GcbqDataset } | Should Throw
+		{ $data2 | New-GcbqDataset } | Should Throw 409
 	}
 
 	It "should use the -Project tag correctly" {
@@ -60,14 +68,14 @@ Describe "New-GcbqDataset" {
 	}
 
 	It "should throw when you try to add to a project that doesnt exist" {
-		{ New-GcbqDataset -Project $nonExistProject "test_data_id8" -Name "Testdata" -Description "Some interesting data!" } | Should Throw
+		{ New-GcbqDataset -Project $nonExistProject "test_data_id8" -Name "Testdata" -Description "Some interesting data!" } | Should Throw 404
     }
 
 	It "should throw when you try to add to a project that is not yours" {
-		{ New-GcbqDataset -Project $accessErrProject "test_data_id9" -Name "Testdata" -Description "Some interesting data!" } | Should Throw
+		{ New-GcbqDataset -Project $accessErrProject "test_data_id9" -Name "Testdata" -Description "Some interesting data!" } | Should Throw 400
 	}
 
-	#add afterall block that deletes test datasets when remove-x command is done
+	# TODO (ahandley) add afterall block that deletes test datasets when remove-x command is done
 }
 
 Describe "Get-GcbqDataset" {
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
@@ -10,54 +10,54 @@ Describe "New-GcbqDataset" {
 		$data.FriendlyName | Should Be "Testdata"
 		$data.Description | Should Be "Some interesting data!"
 		$data.DefaultTableExpirationMs | Should Be 86400000
-    }
+	}
 
 	It "should accept a dataset object from pipeline" {
 		$data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
 		$data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
-        $data.DatasetReference.DatasetId = "test_data_id2"
-        $data.DatasetReference.ProjectId = $project
+		$data.DatasetReference.DatasetId = "test_data_id2"
+		$data.DatasetReference.ProjectId = $project
 		$data.FriendlyName = "PipeTest"
 		$data.Description = "Some cool data coming hot off the pipeline!"
 		$data | New-GcbqDataset -outvariable newdata
 		$newdata | Should Not BeNullOrEmpty
 		$newdata.DatasetReference.DatasetId | Should Be "test_data_id2"
 		$newdata.FriendlyName | Should Be "PipeTest"
-    }
+	}
 
 	It "should accept a dataset object from a command line argument" {
 		$data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
 		$data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
-        $data.DatasetReference.DatasetId = "test_data_id3"
-        $data.DatasetReference.ProjectId = $project
+		$data.DatasetReference.DatasetId = "test_data_id3"
+		$data.DatasetReference.ProjectId = $project
 		$data.FriendlyName = "PipeTest"
 		$data.Description = "Some cool data coming hot off the pipeline!"
 		$newdata = New-GcbqDataset -Dataset $data
 		$newdata | Should Not BeNullOrEmpty
 		$newdata.DatasetReference.DatasetId | Should Be "test_data_id3"
 		$newdata.FriendlyName | Should Be "PipeTest"
-    }
+	}
 
 	It "should not work with no arguments (meaning no datasetId)" {
 		$data = New-GcbqDataset | Should Throw
-    }
+	}
 
 	It "should reject datasets that do not have unique ids" {
 		$data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
 		$data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
-        $data.DatasetReference.DatasetId = "test_data_id6"
+		$data.DatasetReference.DatasetId = "test_data_id6"
 		$data2 = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
 		$data2.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
-        $data2.DatasetReference.DatasetId = "test_data_id6"
+		$data2.DatasetReference.DatasetId = "test_data_id6"
 		$data | New-GcbqDataset
 		{ $data2 | New-GcbqDataset } | Should Throw
-    }
+	}
 
 	It "should use the -Project tag correctly" {
 		$data = New-GcbqDataset -Project $project "test_data_id7" -Name "Testdata" -Description "Some interesting data!"
 		$data | Should Not BeNullOrEmpty
 		$data.DatasetReference.ProjectId | Should Be $project
-    }
+	}
 
 	It "should throw when you try to add to a project that doesnt exist" {
 		{ New-GcbqDataset -Project $nonExistProject "test_data_id8" -Name "Testdata" -Description "Some interesting data!" } | Should Throw
@@ -65,7 +65,7 @@ Describe "New-GcbqDataset" {
 
 	It "should throw when you try to add to a project that is not yours" {
 		{ New-GcbqDataset -Project $accessErrProject "test_data_id9" -Name "Testdata" -Description "Some interesting data!" } | Should Throw
-    }
+	}
 
 	#add afterall block that deletes test datasets when remove-x command is done
 }
--- a/Google.PowerShell/BigQuery/GcbqDataset.cs
+++ b/Google.PowerShell/BigQuery/GcbqDataset.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2015-2016 Google Inc. All Rights Reserved.
+﻿// Copyright 2015-2017 Google Inc. All Rights Reserved.
 // Licensed under the Apache License Version 2.0.
 
 using Google.Apis.Bigquery.v2;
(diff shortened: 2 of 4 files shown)


Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,140 @@
+﻿// Copyright 2015-2016 Google Inc. All Rights Reserved.
```
quoctruong: This should be 2015-2017?

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,140 @@
...
+namespace Google.PowerShell.BigQuery
+{
+    /// <summary>
+    /// <para type="synopsis">
+    /// Creates a new empty dataset in the specified project.
+    /// </para>
+    /// <para type="description">
+    /// Creates a new empty dataset in the specified project.  This command can either take a premade Dataset object, 
+    /// or build one using parameters.  Premade Datasets can be passed via the pipeline or with the -Dataset parameter.
+    /// If both are supplied, the premade dataset object will be used and the builder values will be ignored.  If no
```
quoctruong: Since you are using parameter set, powershell will raise an error if both parameter sets are used.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,140 @@
...
+    /// Required build parameters are passed directly after the command, whereas optional parameters are passed as flags.
+    /// -- Builder parameters: 
+    /// DatasetId (required) - unique identifier for the dataset.
+    /// Name - descriptive name for the dataset.
+    /// Description - user-friendly description of the dataset.
+    /// Timeout - default duration in ms for tables in the dataset to exist.
+    /// </para>
+    /// <example>
+    ///   <code>PS C:\> $dataset | New-GcbqDataset -Project $projId </code>
+    ///   <para>This command will take a premade Dataset object from the pipeline and insert it into the Cloud project.
```
quoctruong: Instead of "Cloud project. project $projectId", should be "Cloud project "$projectId"". Also, I think instead of $projId, you should use "my-project" or something like that.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,140 @@
...
+    [Cmdlet(VerbsCommon.New, "GcbqDataset")]
+    public class NewGcbqDataset : GcbqCmdlet
+    {
+        /// <summary>
+        /// <para type="description">
+        /// The project to look for datasets in. If not set via PowerShell parameter processing, will
+        /// default to the Cloud SDK's DefaultProject property.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = "premade")]
```
quoctruong: DataSet name should be Pascal case. Also, I think you should name it "ByObject" and "ByValues" to be consistent with what we've been using.

quoctruong: You should have a private class that stores the string name of the parameter. An example is [this](https://github.com/GoogleCloudPlatform/google-cloud-powershell/blob/c8295d5634c02caf26f3c6db58dd8f87a5dd4be2/Google.PowerShell/Compute/GceAddressCmdlets.cs#L40).

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// <para type="description">
+        /// The dataset object that will be sent to the server to be inserted.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ValueFromPipeline = true, ParameterSetName = "premade")]
+        public Dataset Dataset { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The DatasetId for the datset reference object.  Must be unique within the project, and must be 1024
```
quoctruong: Too much space after the full stop. Also in the other parameters' comments.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = "builder")]
+        public int Timeout { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            Dataset newData;
+            // Check whether they are passing in a completed Dataset object or are building one with params.
+            if (Dataset != null)
```
quoctruong: You should check the name of the parameterset instead. See [this](https://github.com/GoogleCloudPlatform/google-cloud-powershell/blob/c8295d5634c02caf26f3c6db58dd8f87a5dd4be2/Google.PowerShell/Compute/GceAddressCmdlets.cs#L97) for example.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,140 @@
...
+        public string Description { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The default lifetime for tables in the dataset to exist in ms.  Minimum is 3600000 (1hr).  
+        /// When the expiration time is reached, the table will be automatically deleted.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = "builder")]
+        public int Timeout { get; set; }
```
quoctruong: Should this be long? Also, we should add validation for the minimum value.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,140 @@
...
+                newData.DatasetReference.DatasetId = DatasetId;
+                newData.DatasetReference.ProjectId = Project;
+            }
+
+            // Add the new dataset to the project supplied.
+            DatasetsResource.InsertRequest request = Service.Datasets.Insert(newData, Project);
+            Dataset response = request.Execute();
+            if (response != null)
+            {
+                WriteObject(response, true);
```
quoctruong: We should just use WriteObject(response). WriteObject(response, true) is if the response is an Enumerable.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -0,0 +1,85 @@
...
+		$data.DatasetReference.DatasetId | Should Be "test_data_id1"
+		$data.FriendlyName | Should Be "Testdata"
+		$data.Description | Should Be "Some interesting data!"
+		$data.DefaultTableExpirationMs | Should Be 86400000
+    }
+
+	It "should accept a dataset object from pipeline" {
+		$data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
+		$data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
+        $data.DatasetReference.DatasetId = "test_data_id2"
```
quoctruong: Spacing on this line is not consistent with the others. Same for a few other lines in this file.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -0,0 +1,85 @@
...
+        $data.DatasetReference.ProjectId = $project
+		$data.FriendlyName = "PipeTest"
+		$data.Description = "Some cool data coming hot off the pipeline!"
+		$data | New-GcbqDataset -outvariable newdata
```
quoctruong: When you have a Get-GcbqDataset cmdlet, you should do another test with the object you created. For example, "$onlineData = Get-GcbqDataset test_data_id2" then check that $onlineData has the desired attributes.

ahandley: Thanks for all of the comments.  Just fixed all of that, going to re-push.  

@ the additional tests, I'm planning to do another section below once I have all 4 Dataset Operations done that tests how they function as a group / as the developer would use them.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,145 @@
...
+        {
+            // Set up the Dataset based on parameters
+            Dataset newData;
+            switch (ParameterSetName)
+            {
+                case ParameterSetNames.ByObject:
+                    newData = Dataset;
+                    break;
+                case ParameterSetNames.ByValues:
+                    if (DatasetId == null)
```
quoctruong: Instead of this, you can put [ValidateNotNullOrEmpty] attribute on top of DatasetId so PowerShell will automatically throw an error if user supplies a null or empty DatasetId.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -0,0 +1,85 @@
...
+		$data.FriendlyName = "PipeTest"
+		$data.Description = "Some cool data coming hot off the pipeline!"
+		$newdata = New-GcbqDataset -Dataset $data
+		$newdata | Should Not BeNullOrEmpty
+		$newdata.DatasetReference.DatasetId | Should Be "test_data_id3"
+		$newdata.FriendlyName | Should Be "PipeTest"
+	}
+
+	It "should not work with no arguments (meaning no datasetId)" {
+		$data = New-GcbqDataset | Should Throw
```
quoctruong: Can we check for what is thrown here? For example, Should Throw "error message".

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -0,0 +1,85 @@
...
+
+	It "should throw when you try to add to a project that doesnt exist" {
+		{ New-GcbqDataset -Project $nonExistProject "test_data_id8" -Name "Testdata" -Description "Some interesting data!" } | Should Throw
+    }
+
+	It "should throw when you try to add to a project that is not yours" {
+		{ New-GcbqDataset -Project $accessErrProject "test_data_id9" -Name "Testdata" -Description "Some interesting data!" } | Should Throw
+	}
+
+	#add afterall block that deletes test datasets when remove-x command is done
```
quoctruong: You can add a TODO (ahandley) here so you can address it in the future.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,145 @@
...
+                    {
+                        WriteError(new ErrorRecord(new Exception("DatasetId missing."), "Insert failed.",
+                            ErrorCategory.InvalidArgument, DatasetId));
+                        return;
+                    }
+                    newData = new Dataset();
+                    newData.FriendlyName = (Name != null) ? Name : "New Dataset";
```
quoctruong: Should we just leave the FriendlyName and Description as null if user does not set it?

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -0,0 +1,93 @@
+﻿. $PSScriptRoot\..\BigQuery\GcbqCmdlets.ps1
+$project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
+
+Describe "New-GcbqDataset" {
+
+	It "should take a name, description, and time to make a dataset" {
+		$data = New-GcbqDataset "test_data_id1" -Name "Testdata" -Description "Some interesting data!" -Timeout 86400000
```
chrsmith: Nit: Consider creating a local variable to demystify "86400000". e.g. `$oneDay = ...`.

chrsmith: Also, is `Timeout` the best name for this parameter, if the value it sets is `DefaultTableExpirationMs`?

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -0,0 +1,93 @@
...
+		$data.Description | Should Be "Some interesting data!"
+		$data.DefaultTableExpirationMs | Should Be 86400000
+	}
+
+	It "should accept a dataset object from pipeline" {
+		$data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
+		$data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
+		$data.DatasetReference.DatasetId = "test_data_id2"
+		$data.DatasetReference.ProjectId = $project
+		$data.FriendlyName = "PipeTest"
```
chrsmith: If friendly name is supposed to be user generated, consider stressing this a big harder:
- Add spaces.
- Make it a little longer.
- Add a non-ASCII character.

chrsmith: Same with description?

This is unlikely to find a bug. But I do think we should at least include a space.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	It "should use the -Project tag correctly" {
+		$data = New-GcbqDataset -Project $project "test_data_id7" -Name "Testdata" -Description "Some interesting data!"
+		$data | Should Not BeNullOrEmpty
+		$data.DatasetReference.ProjectId | Should Be $project
+	}
+
+	It "should throw when you try to add to a project that doesnt exist" {
+		{ New-GcbqDataset -Project $nonExistProject "test_data_id8" -Name "Testdata" -Description "Some interesting data!" } | Should Throw 404
```
chrsmith: The spacing/indentation seems off here.

Also are other test files using 8-space tabs? That seems off.

chrsmith: Also, please break the long line so that it is easier to read. Most Google style guides require you to split lines before 80 (everything but Java) or 120 characters (Java).

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -0,0 +1,93 @@
...
+    }
+
+	It "should throw when you try to add to a project that is not yours" {
+		{ New-GcbqDataset -Project $accessErrProject "test_data_id9" -Name "Testdata" -Description "Some interesting data!" } | Should Throw 400
+	}
+
+	# TODO (ahandley) add afterall block that deletes test datasets when remove-x command is done
```
chrsmith: The format here (blame the style guides) is `TODO(ahandley): `. And then the comment (and all code comments) should be a complete sentence. Start with a capital letter, end with a period. Also, use `AfterAll` since it is referring to a code construct.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -0,0 +1,93 @@
...
+}
+
+Describe "Get-GcbqDataset" {
+
```
chrsmith: Better to remove these and add TODOs than leave them blank. Otherwise it gives the false impression that we have tests.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+using Google.Apis.Bigquery.v2.Data;
+using Google.PowerShell.Common;
+using System;
+using System.Management.Automation;
+
+namespace Google.PowerShell.BigQuery
+{
+    /// <summary>
+    /// <para type="synopsis">
+    /// Creates a new empty dataset in the specified project.
```
chrsmith: Since new and empty both modify the same noun, add a comma between then. (I believe that is the English rule here.)

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+    /// </para>
+    /// <para type="description">
+    /// Creates a new empty dataset in the specified project from a Dataset Object.  This Dataset can be supplied by object via the 
+    /// pipeline or with the -ByObject parameter, or they can be supplied by value with the flags listed below.  If no Project is 
+    /// specified, the default project will be used.  This cmdlet returns a Dataset object.  Required value parameters are to be
```
chrsmith: The "Required value parameters..." is unnecessary, since `Get-Help <cmdlet name` will provide this same information, along with showing which parameters are optional and which are required.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+    /// specified, the default project will be used.  This cmdlet returns a Dataset object.  Required value parameters are to be 
+    /// passed in order after the command, and optional parameters are passed as flags.
+    ///    ByValue Parameters: 
+    /// DatasetId (required) - unique identifier for the dataset.
+    /// Name - descriptive name for the dataset.
+    /// Description - user-friendly description of the dataset.
+    /// Timeout - default duration in ms for tables in the dataset to exist. 
+    /// </para>
+    /// <example>
+    ///   <code>PS C:\> $dataset | New-GcbqDataset -Project "my-project" </code>
```
chrsmith: Delete trailing space.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+    ///   <para>This command will take a Dataset object from the pipeline and insert it into the Cloud project "my-project". 
+    ///   project $projId</para>
```
chrsmith: Delete this line? Shouldn't be here?

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+    /// </example>
+    /// <example>
+    ///   <code>PS C:\> New-GcbqDataset "test_data_id" -Name "Testdata" -Description "Some interesting data!" -Timeout 86400000</code>
```
chrsmith: Please split this at 80 chars (ideally 60). The code snippets look very bad on the documentation site if they are long. (Use the "`" character.)

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+            public const string ByValues = "ByValue";
+        }
+
+        /// <summary>
+        /// <para type="description">
+        /// The project to look for datasets in. If not set via PowerShell parameter processing, will
+        /// default to the Cloud SDK's DefaultProject property.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.ByObject)]
```
chrsmith: Just one "Parameter" attribute is fine. If you don't specify a ParameterSetName, it will refer to all parameter sets.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// The dataset object that will be sent to the server to be inserted.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ValueFromPipeline = true, ParameterSetName = ParameterSetNames.ByObject)]
+        public Dataset Dataset { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The DatasetId for the datset reference object. Must be unique within the project, and must be 1024 
+        /// characters or less. It must only contain letters (a-z, A-Z), numbers (0-9), or underscores (_).
```
chrsmith: Consider just providing the regular expression, assume the reader can figure that out.

E.g. "Must match the following pattern `[a-zA-Z0-9_]+`."

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// </para>
+        /// </summary>
+        [Parameter(Position = 0, Mandatory = true, ParameterSetName = ParameterSetNames.ByValues)]
+        [ValidatePattern("[a-zA-Z0-9]")]
```
chrsmith: This doesn't match the comment above, needs the underscore character.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// <para type="description">
+        /// The dataset object that will be sent to the server to be inserted.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ValueFromPipeline = true, ParameterSetName = ParameterSetNames.ByObject)]
+        public Dataset Dataset { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The DatasetId for the datset reference object. Must be unique within the project, and must be 1024
```
chrsmith: Consider just saying "Must be unique within the project". The 1 to 1024 characters is kinda of superfluous. (Since I highly doubt people will run into that 1024 limit and be surprised.)

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// </para>
+        /// </summary>
+        [Parameter(Position = 0, Mandatory = true, ParameterSetName = ParameterSetNames.ByValues)]
+        [ValidatePattern("[a-zA-Z0-9]")]
+        [ValidateLength(1, 1024)]
+        public string DatasetId { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The "friendly name" field for the dataset. Used for a descriptive name for the dataset.
```
chrsmith: Remove quotes from "friendly name"? How about just "User-friendly name for the dataset."

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.ByValues)]
+        public string Name { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The "description" field for the dataset. Used for a user-friendly description of the dataset.
```
chrsmith: Just "Description of the dataset."

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.ByValues)]
+        public string Description { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The default lifetime for tables in the dataset to exist in ms. Minimum is 3600000 (1hr).
```
chrsmith: Please grep through PowerShell docs (especially our other cmdlets) and make this consistent with the majority.

I believe most (all?) cmdlets use "seconds" as the unit of timeouts. MS isn't especially useful, and even if that is the implementation detail, we can hide that for our users.

(This is OK, since the output object says "ExpirationMs", so it won't lead to confusion.)

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,141 @@
...
+            // Add the new dataset to the project supplied.
+            DatasetsResource.InsertRequest request = Service.Datasets.Insert(newData, Project);
+            Dataset response = request.Execute();
+            if (response != null)
+            {
+                WriteObject(response);
+            }
+            else
+            {
+                WriteError(new ErrorRecord(new Exception("400"), "Insert request to server failed.",
```
chrsmith: Break this differently. The problem with the code as is, is that you are breaking after providing two parameters to the second type being instantiated. Read: this is harder to parse than it should be.

````
WriteError(new ErrorRecord(
    new Exception("400"),
    "Insert request to server failed.",
    ErrorCategory.InvalidArgument,
    newData));
````

If you are going to spill calls with mutliple args across a few lines, be sure to do it in such a way it is clear which function/constructor is being called. 

This isn't clear in the Java style guide, but generally try to follow that: https://engdoc.corp.google.com/eng/doc/devguide/java/styleguide.shtml?cl=head#s4.5.1-line-wrapping-where-to-break

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
ahandley: Visual Studio defaults to just tabs.  I'm sure I can configure it to just insert spaces when i tab though.  What should the default width be?  (which style guide for that one)

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
ahandley: Fixed.  Changed parameter to -Expiration, converted units to seconds (modeled after the pubsub cmdlets), and fixed the test.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
ahandley: Added a test for that.  Does slightly more complex values.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
ahandley: Got it!

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
ahandley: Got it!

ahandley:
Just going to reply once to all of the formatting changes instead of replying individually. I went through and addressed all of them, either sticking exactly to the changes or rewording the surrounding language.   I'm going to have to push the changes in with the next set because of the git mishap.  It should be fixed as soon as I can get the next changeset patched in, and then I'll go back to small commits.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
chrsmith: Prefer spaces over tabs. That is to be consistent with every language Google uses but Go. Go uses Tabs over spaces, because "spaces for alignment, tabs for indentation".

"Use the default Code Editor settings (smart indenting, four-character indents, tabs saved as spaces)." 4-spaces to deliniate blocks.

https://msdn.microsoft.com/en-us/library/ff926074.aspx

ahandley: Can do.  I'll make sure all of my code gets amended.

PR #465, opened 2017-02-16
Datasetcmds

ahandley:
Sorry for the giant PR.  The new content is the Get- and Remove- cmdlets & tests.  This also includes the changes from the New- cmdlet & tests suggested by Chris.


Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -0,0 +1,364 @@
...
+                {
+                    WriteError(new ErrorRecord(new Exception(response), "Deletion failed.",
+                        ErrorCategory.OperationStopped, response));
+                }
+            }
+        }
+
+        private string GetConfirmationMessage(Dataset d, int? tables)
+        {
+            return d.DatasetReference.DatasetId
```
chrsmith: Use the string interpolation operator.

````
return $"{d.DatasetReference.DatasetId} has {tables} tables and the -Force parameter was not specified...";
````

ahandley: Oh cool.  Had never heard of that.  Got it!

ahandley:
Made the changes and also added in try{}finally{} handling for the tests to clean up the test projects regardless of whether the test passed or not.

PR #466, opened 2017-02-17
Get-GcbqDataset cmdlet PR

ahandley:
Final Dataset cmdlet.  Purposefully ignoring the "patch" vs "update" case because in the c# api, they're functionally the same.  Also streamlines UX.

### DIFF
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
@@ -4,14 +4,13 @@ $project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
 Describe "New-GcbqDataset" {
 
     It "should take a name, description, and time to make a dataset" {
-        $oneDay = 86400
         try {
-            $data = New-GcbqDataset "test_data_id1" -Name "Testdata" -Description "Some interesting data!" -Expiration $oneDay
+            $data = New-GcbqDataset "test_data_id1" -Name "Testdata" -Description "Some interesting data!" -Expiration $oneDaySec
             $data | Should Not BeNullOrEmpty
             $data.DatasetReference.DatasetId | Should Be "test_data_id1"
             $data.FriendlyName | Should Be "Testdata"
             $data.Description | Should Be "Some interesting data!"
-            $data.DefaultTableExpirationMs | Should Be ($oneDay * 1000)
+            $data.DefaultTableExpirationMs | Should Be $oneDayMs
         }
         finally {
             Get-GcbqDataset "test_data_id1" | Remove-GcbqDataset
@@ -236,6 +235,46 @@ Describe "Remove-GcbqDataset" {
     }
 }
 
-#TODO(ahandley): Describe "Set-GcbqDataset".
+Describe "Set-GcbqDataset" {
+
+    It "should update trivial metadata fields via pipeline (not DatabaseId)" {
+        try {
+            New-GcbqDataset "test_dataset_id1" -Name "Testdata" -Description "Some interesting data!" -Expiration $oneDaySec
+            $data = Get-GcbqDataset "test_dataset_id1"
+            $data.FriendlyName = "Some Test Data"
+            $data.Description = "A new description!"
+            $data.DefaultTableExpirationMs = $threeDayMs
+            $data | Set-GcbqDataset -outvariable data
+            $data | Should Not BeNullOrEmpty
+            $data.FriendlyName | Should Be "Some Test Data"
+            $data.Description | Should Be "A new description!"
+            $data.DefaultTableExpirationMs | Should Be $threeDayMs
+        }
+        finally {
+            Get-GcbqDataset "test_dataset_id1" | Remove-GcbqDataset
+        }
+    }
+
+    It "should update trivial metadata fields via parameter (not DatabaseId)" {
+        try {
+            New-GcbqDataset "test_dataset_id1" -Name "Testdata" -Description "Some interesting data!" -Expiration $oneDaySec
+            $data = Get-GcbqDataset "test_dataset_id1"
+            $data.FriendlyName = "Some Test Data"
+            $data.Description = "A new description!"
+            $data.DefaultTableExpirationMs = $threeDayMs
+            $data = Set-GcbqDataset -ByObject $data
+            $data | Should Not BeNullOrEmpty
+            $data.FriendlyName | Should Be "Some Test Data"
+            $data.Description | Should Be "A new description!"
+            $data.DefaultTableExpirationMs | Should Be $threeDayMs
+        }
+        finally {
+            Get-GcbqDataset "test_dataset_id1" | Remove-GcbqDataset
+        }
+    }
+
+    #TODO(ahandley): Add in tests for updating Access field when cmdlet support has been built.
+
+}
 
 Reset-GCloudConfig $oldActiveConfig $configName
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
@@ -1,6 +1,89 @@
 ﻿. $PSScriptRoot\..\BigQuery\GcbqCmdlets.ps1
 $project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
 
+Describe "Get-GcbqDataset" {
+
+    It "should list datasets" {
+        try {
+            New-GcbqDataset "test_id_2"
+            New-GcbqDataset "test_id_3"
+            $a = Get-GcbqDataset
+            $a.Datasets.Count | Should Not Be 0
+        }
+        finally {
+            Get-GcbqDataset "test_id_2" |  Remove-GcbqDataset
+            Get-GcbqDataset "test_id_3" |  Remove-GcbqDataset
+        }
+    }
+
+    It "should get a dataset that exists and has permissions" {
+        try {
+            New-GcbqDataset "test_id_4"
+            $a = Get-GcbqDataset "test_id_4"
+            $a.DatasetReference.DatasetId | Should Be "test_id_4"
+        }
+        finally {
+            Get-GcbqDataset "test_id_4" |  Remove-GcbqDataset
+        }
+    }
+
+    It "should handle when a dataset does not exist" {
+        { Get-GcbqDataset "test_id_5" -ErrorAction Stop } | Should Throw 404
+    }
+
+    It "should handle projects that do not exist" {
+        { Get-GcbqDataset -Project $nonExistProject "test_id_5" -ErrorAction Stop } | Should Throw 404
+    }
+
+    It "should handle projects that the user does not have permissions for" {
+        { Get-GcbqDataset -Project $accessErrProject "test_id_5" } | Should Throw 400
+    }
+}
+
+Describe "Set-GcbqDataset" {
+
+    It "should update trivial metadata fields via pipeline (not DatabaseId)" {
+        try {
+            New-GcbqDataset "test_dataset_id1" -Name "Testdata" -Description "Some interesting data!" -Expiration $oneDaySec
+            $data = Get-GcbqDataset "test_dataset_id1"
+            $data.FriendlyName = "Some Test Data"
+            $data.Description = "A new description!"
+            $data.DefaultTableExpirationMs = $threeDayMs
+
+            $data = $data | Set-GcbqDataset
+            $data | Should Not BeNullOrEmpty
+            $data.FriendlyName | Should Be "Some Test Data"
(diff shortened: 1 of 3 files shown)


Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -136,6 +136,68 @@ protected override void ProcessRecord()
 
     /// <summary>
     /// <para type="synopsis">
+    /// Updates information in an existing dataset.
```
iantalarico: Can you add more context? existing BigQuery Dataset?

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -136,6 +136,68 @@ protected override void ProcessRecord()
...
+    /// </para>
+    /// <para type="description">
+    /// Updates information in an existing dataset. If no Project is specified, the default project will be used.
```
iantalarico: should project be capitalized here?

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -136,6 +136,68 @@ protected override void ProcessRecord()
...
     /// <para type="synopsis">
+    /// Updates information in an existing dataset.
+    /// </para>
+    /// <para type="description">
+    /// Updates information in an existing dataset. If no Project is specified, the default project will be used. 
+    /// This cmdlet returns a Dataset object.
+    /// </para>
+    /// <example>
+    ///   <code>PS C:\> $updatedSet | Set-GcbqDataset</code>
+    ///   <para>This will update the values stored in the cloud for the dataset passed via pipeline.</para>
```
iantalarico: I think the comment here could be more explicit.

This command will update the values stored in the BigQuery dataset for the default project passed.  

I would also add more context as to what " $updatedSet " is.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -136,6 +136,68 @@ protected override void ProcessRecord()
...
+    /// </example>
+    /// <example>
+    ///   <code>PS C:\> Set-GcbqDataset -Project my_project -ByObject $modifedSet</code>
+    ///   <para>This overwrites my_project:my_data_id with the dataset from $modifiedSet.</para>
```
iantalarico: Again here I would be more explicit

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -136,6 +136,68 @@ protected override void ProcessRecord()
...
+            
+            if (response != null)
+            {
+                WriteObject(response);
+            }
+            else
+            {
+                WriteError(new ErrorRecord(
+                    new Exception("400"),
+                    $"Set request for {ByObject.DatasetReference.DatasetId} failed.",
```
iantalarico: what was the invalid argument? Is this always the failure if we are here?

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -136,6 +136,68 @@ protected override void ProcessRecord()
...
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, ValueFromPipeline = true)]
+        public Dataset ByObject { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            Dataset response;
+            var request = new DatasetsResource.UpdateRequest(Service, ByObject, Project, ByObject.DatasetReference.DatasetId);
+            response = request.Execute();
```
iantalarico: Do we need to catch any exception here? 403s?

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: I put it in caps because it was referring to the flag -Project.  should I put the dash in there to make it more clear?

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: What's the best practice here? (@quoct) Should we be catching exceptions in cmdlets or should we let them propagate to the user to see and fix?  The response is written regardless, and the only thing I "catch" is when the response is null and something really went wrong.   I'll look into other cmdlets to see what the general consensus is once i'm done making other changes.

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: I think I left that from a copy-paste because I couldn't find something more descriptive.  WriteError is probably better.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -136,6 +136,68 @@ protected override void ProcessRecord()
...
+        [ConfigPropertyName(CloudSdkSettings.CommonProperties.Project)]
+        override public string Project { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The updated Dataset object.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, ValueFromPipeline = true)]
+        public Dataset ByObject { get; set; }
```
ILMTitan: I think the standard name for this type of parameter would be "InputObject". "ByObject" sounds like the name of a parameter set.

Google.PowerShell.IntegrationTests/BigQuery/GcbqCmdlets.ps1
```diff
@@ -4,5 +4,9 @@ Install-GcloudCmdlets
 # Define variables that will be used in testing
 $nonExistProject = "project-no-exist"
 $accessErrProject = "asdf"
+$oneDaySec = 86400
```
chrsmith: This is fine as-is, but consider the following to be more readable. (1) Don't just include a constant, but the values used to derive it. (2) Use partial values to make it more readable.

````
$oneDaySec = 60 * 60 * 24;
$oneDayMs = $oneDaySec * 1000
$threeDaySec = 3 * $oneDaySec
$threeDayMs = 3 * $oneDayMs
````

File: Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1 [hunk 9]
```diff
@@ -236,6 +235,46 @@ Describe "Remove-GcbqDataset" {
...
+Describe "Set-GcbqDataset" {
+
+    It "should update trivial metadata fields via pipeline (not DatabaseId)" {
+        try {
+            New-GcbqDataset "test_dataset_id1" -Name "Testdata" -Description "Some interesting data!" -Expiration $oneDaySec
+            $data = Get-GcbqDataset "test_dataset_id1"
+            $data.FriendlyName = "Some Test Data"
+            $data.Description = "A new description!"
+            $data.DefaultTableExpirationMs = $threeDayMs
+            $data | Set-GcbqDataset -outvariable data
```
ILMTitan: You can also do `$data = $data | Set-GcbqDataset`. If you want to keep it as is, I would suggest capitalizing `-OutVariable`.

Google.PowerShell.IntegrationTests/BigQuery/GcbqCmdlets.ps1
ahandley: That makes sense.  Changed it!

Google.PowerShell/BigQuery/GcbqDataset.cs
ILMTitan: We have been letting these errors bubble up to the user. There are too many ways it could go wrong for us to create a prettier message for every one. Better to give them all the information they might need to fix the problem was the thought.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -236,6 +235,46 @@ Describe "Remove-GcbqDataset" {
...
+    }
+
+    It "should update trivial metadata fields via parameter (not DatabaseId)" {
+        try {
+            New-GcbqDataset "test_dataset_id1" -Name "Testdata" -Description "Some interesting data!" -Expiration $oneDaySec
+            $data = Get-GcbqDataset "test_dataset_id1"
+            $data.FriendlyName = "Some Test Data"
+            $data.Description = "A new description!"
+            $data.DefaultTableExpirationMs = $threeDayMs
+            $data = Set-GcbqDataset -ByObject $data
```
ILMTitan: I would add a blank line above here to help show visually this is the line being tested.

Google.PowerShell/BigQuery/GcbqDataset.cs
ILMTitan: I agree. Add a line where you get the $updatedSet, and a line where you set a new description before the call to Set-GcbqDataSet.

ahandley: Got it.  Added in both of those two lines, as well as updating the default table expiration time in the second example for added wonder and amazement.  Updated a lot of the wording as well to be more explicit.

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: Alright.  I'll leave it as is then.

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: You seem to be correct.  I misunderstood a previous comment about the set names then.  I'll update that on all 4 cmdlets.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
ahandley: That does seem more elegant.  Thanks.

Google.PowerShell/BigQuery/GcbqDataset.cs
quoctruong: @ahandley If it's a common exception (or if the error is too cryptic) then you should catch it, create your own ErrorRecord and use WriteError.

ahandley: So 403's and 404's?  Also should this be done for all cmdlets?

quoctruong: It really depends on the cmdlet. For example, for cmdlet that deals with existing object, 404 (NotFound) is a common scenario whereas for cmdlet that creates an object, 409 (Conflict) is the common scenario. You can probably deals with 404 here.

ahandley:
Changes made.  Major points:  clairfying wording and examples, added proper exception handling to integration tests, and fixed names of parameters.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -98,33 +99,32 @@ protected override void ProcessRecord()
...
                     lRequest.Filter = Filter;
-                    var lResponse = lRequest.Execute();
-                    if (lResponse != null)
+                    try
                     {
+                        var lResponse = lRequest.Execute();
                         WriteObject(lResponse, true);
                     }
-                    else
+                    catch
```
ILMTitan: This naked catch concerns me. You can't know that it was a 400. What if it was a 500? What if GCP is down? I think the old if else is better in this case.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -98,33 +99,32 @@ protected override void ProcessRecord()
...
                     DatasetsResource.GetRequest gRequest = new DatasetsResource.GetRequest(Service, Project, Dataset);
-                    var gResponse = gRequest.Execute();
-                    if (gResponse != null)
+                    try
                     {
+                        var gResponse = gRequest.Execute();
                         WriteObject(gResponse);
                     }
-                    else
+                    catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.NotFound)
```
ILMTitan: This is a good change.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -98,33 +99,32 @@ protected override void ProcessRecord()
...
                     {
-                        WriteError(new ErrorRecord(
-                            new Exception("400"), 
-                            "Get request to server failed.",
-                            ErrorCategory.InvalidArgument, 
+                        WriteError(new ErrorRecord(new Exception("404"),
```
ILMTitan: Why create a new exception here? You already have a perfectly good one with the correct error code.

ahandley: Ah, I changed that thinking it was the cause of a Pester bug,  Going to put that back to ex.

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: This is true.  Blanket changes are probably bad.  Fixed.

PR #467, opened 2017-02-21
Add ShouldContinue for removing folder

quoctruong:
Also, use path separator based on OS.

image

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/467)

### DIFF
--- a/Google.PowerShell/Provider/BucketModel.cs
+++ b/Google.PowerShell/Provider/BucketModel.cs
@@ -77,8 +77,8 @@ namespace Google.PowerShell.CloudStorage
                 // Find the prefixes (parent folders) of this object.
                 string prefix = gcsObject.Name.TrimEnd(Separator);
                 int lastSeparator = gcsObject.Name.LastIndexOf(Separator);
-                // If the object name is testing/blah.txt, then it has children since / is not at the end.
-                // Otherwise, we don't know yet.
+                // If the object name is testing/blah.txt, then the first prefix "testing" has children
+                // since '/' is not at the end. Otherwise, we don't know yet.
                 bool children = 0 < lastSeparator && lastSeparator < prefix.Length - 1;
                 while (lastSeparator > 0)
                 {
--- a/Google.PowerShell/Provider/BucketModel.cs
+++ b/Google.PowerShell/Provider/BucketModel.cs
@@ -77,7 +77,9 @@ namespace Google.PowerShell.CloudStorage
                 // Find the prefixes (parent folders) of this object.
                 string prefix = gcsObject.Name.TrimEnd(Separator);
                 int lastSeparator = gcsObject.Name.LastIndexOf(Separator);
-                bool children = false;
+                // If the object name is testing/blah.txt, then it has children since / is not at the end.
+                // Otherwise, we don't know yet.
+                bool children = 0 < lastSeparator && lastSeparator < prefix.Length - 1;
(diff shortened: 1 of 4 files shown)


Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
```diff
@@ -859,13 +861,32 @@ protected override object RemoveItemDynamicParameters(string path, bool recurse)
...
         {
+            IEnumerable<Object> children = ListChildren(gcsPath, true);
+            string query = $"The item at the path '{gcsPath.ObjectPath}' has children and the Recurse parameter was not specified."
+                + " If you continue, all children will be removed with the item. Are you sure you want to continue?";
+            string caption = "Confirm";
+
+            // Ask the user for permission to delete children of the folder if -Recurse is not supplied.
+            if (children.Any() && !recurse)
+            {
+                if (ShouldContinue(query, caption))
```
ILMTitan: If we want to force this, we should delete the children first in a loop checked by the four parameter ShouldContinue. This will let the user examine the names of each GcsObject they are deleting, or just say "Yes to All".

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
```diff
@@ -835,6 +836,7 @@ protected override void RemoveItem(string path, bool recurse)
                 case GcsPath.GcsPathType.Object:
                     if (IsItemContainer(path))
                     {
+                        path = path.TrimEnd("/\\".ToCharArray());
```
ILMTitan: String.TrimEnd can take a params array. This could be `path.TrimEnd('/', '\\');`

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
```diff
@@ -859,13 +861,32 @@ protected override object RemoveItemDynamicParameters(string path, bool recurse)
...
+            // Ask the user for permission to delete children of the folder if -Recurse is not supplied.
+            if (children.Any() && !recurse)
+            {
+                if (ShouldContinue(query, caption))
+                {
+                    recurse = true;
+                }
+                else
+                {
+                    return;
```
ILMTitan: Even if we don't delete all of the children, we should try to delete the folder object if it is real. It could contain undesired data and there is already a ShouldProcess gate on it.

quoctruong:
@ILMTitan PTAL. Turns out there is a bug in BucketModel that makes the provider fails to detect whether a folder has children or not. When I fixed the bug, we don't need the extra logic in RemoveFolder so I reverted that.

Google.PowerShell/Provider/BucketModel.cs
```diff
@@ -77,7 +77,9 @@ public void PopulateModel()
                 // Find the prefixes (parent folders) of this object.
                 string prefix = gcsObject.Name.TrimEnd(Separator);
                 int lastSeparator = gcsObject.Name.LastIndexOf(Separator);
-                bool children = false;
+                // If the object name is testing/blah.txt, then it has children since / is not at the end.
```
ILMTitan: Consider changing "it" to "the first prefix". "It" sounds like you are talking about "testing/blah.txt", which probably does not have children.

quoctruong:
Thanks!

PR #471, opened 2017-02-22
Get-GcbqTable cmdlet and integration tests

ahandley:


Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+                    WriteError(new ErrorRecord(ex,
+                        $"Error 404: Table {Table} not found in {Dataset}.",
+                        ErrorCategory.ObjectNotFound,
+                        Table));
+                }
+            }
+        }
+    }
+
+    /// <summary>
```
ILMTitan: Remove these stubs from the PR itself.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+        public string Dataset { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The ID of the dataset that you want to get a descriptor object for. 
+        /// (This will be used if both InputObject and Dataset are provided)
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ValueFromPipeline = true)]
+        public Dataset InputObject { get; set; }
```
quoctruong: You can actually remove this parameter by using PropertyByTypeTransformation. See [this](https://github.com/GoogleCloudPlatform/google-cloud-powershell/blob/master/Google.PowerShell/PubSub/GcpsSubscription.cs#L89). So would just have have a parameter Dataset which is mandatory.

Google.PowerShell/BigQuery/GcbqTable.cs
quoctruong: Shouldn't include the placeholder cmdlets in the PR.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+    /// <para type="link" uri="(https://cloud.google.com/bigquery/docs/reference/rest/v2/tables)">
+    /// [BigQuery Tables]
+    /// </para>
+    /// </summary>
+    [Cmdlet(VerbsCommon.Get, "GcbqTable")]
+    public class GetGcbqTable : GcbqCmdlet
+    {
+        /// <summary>
+        /// <para type="description">
+        /// The project to look for datasets in. If not set via PowerShell parameter processing, it will
```
quoctruong: Should be look for tables in.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// default to the Cloud SDK's DefaultProject property.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        [ConfigPropertyName(CloudSdkSettings.CommonProperties.Project)]
+        override public string Project { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The ID of the dataset that you want to get a descriptor object for.
```
quoctruong: The comment here is not correct. Same for the comment for InputObject parameter. This cmdlet is retrieving the table, not the dataset.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// The ID of the dataset that you want to get a descriptor object for. 
+        /// (This will be used if both InputObject and Dataset are provided)
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ValueFromPipeline = true)]
+        public Dataset InputObject { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The ID of the dataset that you want to get a descriptor object for.
```
quoctruong: The comment here is not correct. Same for the comment for InputObject parameter. This cmdlet is retrieving the table, not the dataset.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+    /// <summary>
+    /// <para type="synopsis">
+    /// Lists all tables in the specified dataset or returns a specific table.
+    /// </para>
+    /// <para type="description">
+    /// Lists all tables in the specified dataset. Requires the READER dataset role. If a table ID is specified, 
+    /// it will return the table resource, which describes the data in the table. Note that this is not the 
+    /// actual data from the table. If no Project is specified, the default project will be used. The dataset 
+    /// may be selected by passing in a Dataset object via pipeline or by passing the dataset ID with the 
+    /// -Dataset parameter. This cmdlet returns a TableList object if no Table was specified, and a Table otherwise.
```
quoctruong: I don't think you should return a TableList, you should get the tables in the TableList and return them. See comment on line 100.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+    /// </para>
+    /// <example>
+    ///   <code>PS C:\> Get-GcbqDataset “my_data” | Get-GcbqTable</code>
```
quoctruong: I don't think we should support pipelining for this cmdlet. It's not very idiomatic in this scenario. Normally pipelining would be something like "Table A" | Remove-GcbqTable.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+            Dataset = (InputObject != null) ? InputObject.DatasetReference.DatasetId : Dataset;
+
+            // Create and execute request.
+            if (Table == null)
+            {
+                TablesResource.ListRequest request = new TablesResource.ListRequest(Service, Project, Dataset);
+                var response = request.Execute();
+                if (response != null)
+                {
+                    WriteObject(response, true);
```
quoctruong: Should be WriteObject(response.Tables, true)?

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+                }
+                else
+                {
+                    WriteError(new ErrorRecord(
+                        new Exception("400"),
+                        "Error 400: List request to server failed.",
```
ILMTitan: I would avoid saying the request failed, or giving an http error code unless you actually get a `GoogleApiException`. The libraries we are using have a nasty tendency to have a null list when it should have an empty list. You can make a valid request, that should return no data, and get back `null`. I would add a verbose message to that effect. An error might lead users to kill their script even though the response was perfectly fine.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+                    ErrorCategory.InvalidArgument,
+                    InputObject));
+                return;
+            }
+            Dataset = (InputObject != null) ? InputObject.DatasetReference.DatasetId : Dataset;
+
+            // Create and execute request.
+            if (Table == null)
+            {
+                TablesResource.ListRequest request = new TablesResource.ListRequest(Service, Project, Dataset);
```
ILMTitan: This can be rewritten as `TablesResource.ListRequest request = Service.Tables.List(Project, Dataset);`

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
```diff
@@ -0,0 +1,58 @@
+﻿. $PSScriptRoot\..\BigQuery\GcbqCmdlets.ps1
+$project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
+
+Describe "Get-GcbqTable" {
```
quoctruong: Is it possible to add code to generate tables and then check for the generated tables in the test? (Also remove the generated tables at the end of the test)

ahandley: I was planning to add that in as I filled in the other cmdlets.  At least, that's what I did with the Datasets.  Are you saying using the bq command or something external to generate/destroy them?

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: got it.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+                var response = request.Execute();
```
ILMTitan: The response has a `NextPageToken`, which means you need to iterate over page tokens. Look at `GetGceNetworkCmdlet.GetAllProjectNetworks()` to see a good example of doing that.

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: fixed

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: fixed.

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: fixed

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+                    WriteError(new ErrorRecord(
+                        new Exception("400"),
+                        "Error 400: List request to server failed.",
+                        ErrorCategory.InvalidArgument,
+                        Dataset));
+                }
+            }
+            else
+            {
+                TablesResource.GetRequest request = new TablesResource.GetRequest(Service, Project, Dataset, Table);
```
ILMTitan: This could be written as `Service.Tables.Get(Project, Dataset, Table);`

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: That does make more sense.  I did that for Projects, but not for Datasets.  Going to fix it here and then go ahead and fix it in the Dataset get cmdlet as well.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
```diff
@@ -0,0 +1,58 @@
...
+
+    It "should list tables from a dataset ID by parameter."{
+        $tables = Get-GcbqTable -Dataset “pshell_testing”
+        $tables.Count | Should Not Be 0
+    }
+
+    It "should get a singular table with a dataset object from pipeline."{
+        $table = Get-GcbqDataset “pshell_testing” | Get-GcbqTable "my_table"
+        $table.TableReference.TableId | Should be "my_table"
+        $table.Description | Should Be "TEST"
```
ILMTitan: Looking at your sample data, both tables have the same description. You should give them different descriptions.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
quoctruong: That works too. Just add a TODO for that.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
```diff
@@ -0,0 +1,58 @@
+﻿. $PSScriptRoot\..\BigQuery\GcbqCmdlets.ps1
+$project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
+
+Describe "Get-GcbqTable" {
+
+    It "should list tables from a dataset object from pipeline."{
+        $tables = Get-GcbqDataset “pshell_testing” | Get-GcbqTable
```
ILMTitan: This sample data is not set up in the test. Does that take time to do? If not much more than querying the data, the test should set up and tear down the sample data itself (in a BeforeAll and an AfterAll).

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: Is it bad to have extra functionality?  It's also supported here by passing in via parameter which is the way that all of the other cmdlets will take in a dataset ID.    (And I can totally see arguments either way, just wondering what the trend is for this codebase)

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,198 @@
...
+        /// The ID of the dataset that you want to get a descriptor object for.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, Position = 0)]
+        public string Table { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            // Select the DatasetID. Favor InputObject if both present. Default if neither present.
+            if (InputObject == null && Dataset == null)
```
ahandley: Quoc said the same.  I'll definitely look into that.  I didn't like the current solution anyway.  Just felt disorganized.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
ahandley: As I mentioned above as well, I was planning to iterate on the tests as I developed stuff like New-GcbqTable and Remove-GcbqTable.  Ideally by the end of this block of cmdlets, I would like to be using all of them anyway instead of external cmds like "bq".  That way the tests will eventually be contained individually and show that the cmdlets work properly together.

Google.PowerShell/BigQuery/GcbqTable.cs
ILMTitan: Note that multiple PropertyByTypeTransformation should run in order, so you can first do Dataset => DatasetReference and then DatasetReference => DatasetId.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
ahandley: got it.

Google.PowerShell/BigQuery/GcbqTable.cs
ILMTitan: The Microsoft [PowerShell Cmdlet guidlines](https://msdn.microsoft.com/en-us/library/dd878270(v=vs.85).aspx#SC02) encourage a pipeline input for every ParameterSet. Piping in the Dataset seems a reasonable use for this cmdlet, as it would let you get all tables for several Datasets (or even all tables with a specific name over several Datasets).

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: Ah, yes.  I did that before on the Project cmdlet.  Disregarded it this time because we said to nix the input parameters on the project one.  I've spliced that code in now.

Google.PowerShell/BigQuery/GcbqTable.cs
ILMTitan: Be careful, in my experience with similar libraries, response.Tables may be null (and never an empty list), and we don't want to put a null on the pipeline when we should put nothing at all.

ahandley: It won't matter since i'm re-coding it to output individual table objects instead of 1+ lists.

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: got it

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: got it

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
ILMTitan: I agree with Quoc, a TODO: is a great way to show that. You can even stub out the BeforeAll and AfterAll and put the TODO in there.

ahandley:
PTAL.  Everything should have been addressed here, and I added some fixed to the Dataset cmdlets in the same vein.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -96,24 +96,28 @@ protected override void ProcessRecord()
             switch (ParameterSetName)
             {
                 case ParameterSetNames.List:
-                    DatasetsResource.ListRequest lRequest = new DatasetsResource.ListRequest(Service, Project);
+                    DatasetsResource.ListRequest lRequest = Service.Datasets.List(Project);
```
ILMTitan: This is getting to be a pretty long case statement. Consider moving most of this code into a function. The standard practice has been to have the function that returns an Enumeration of the objects we want to write to the pipeline, and have the WriteObject call remain in the case statement. Look at how yield return works to help build the Enumeration.

    case ParameterSetNames.List:
        Enumeration<Dataset> datasets = ListProjectDatasets(Project);
        WriteObject(datasets, true);
        break;
    case ParameterSetNames.Get:
        Dataset dataset = GetDataset(Project, Dataset);
        WriteObject(dataset);
        break;

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -386,8 +390,7 @@ public class RemoveGcbqDataset : GcbqCmdlet
         protected override void ProcessRecord()
         {
             // Form and send request.
-            DatasetsResource.DeleteRequest request = 
-                new DatasetsResource.DeleteRequest(Service, Project, InputObject.DatasetReference.DatasetId);
+            DatasetsResource.DeleteRequest request = Service.Datasets.Delete(Project, InputObject.DatasetReference.DatasetId);
```
quoctruong: This line is too long. You should break it up.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -52,65 +53,48 @@ public class GetGcbqTable : GcbqCmdlet
...
             if (Table == null)
             {
-                TablesResource.ListRequest request = new TablesResource.ListRequest(Service, Project, Dataset);
-                var response = request.Execute();
-                if (response != null)
-                {
-                    WriteObject(response, true);
-                }
-                else
+                TablesResource.ListRequest request = Service.Tables.List(Project, Dataset);
```
ILMTitan: Like above, this is probably a good place for a new function.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
```diff
@@ -0,0 +1,56 @@
...
+
+    #TODO(ahandley): Include setup and teardown code in these tests as New- and Remove- are developed.
+
+    It "should list tables from a dataset object from pipeline."{
+        $tables = Get-GcbqDataset “pshell_testing” | Get-GcbqTable
+        $tables.Count | Should BeGreaterThan 0
+    }
+
+    It "should list tables from a dataset object via parameter."{
+        $table = Get-GcbqDataset “pshell_testing”
```
quoctruong: This variable should be $dataset instead of $table.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -96,24 +96,28 @@ protected override void ProcessRecord()
...
+                        {
+                            WriteObject(response.Datasets, true);
+                        }
+                        lRequest.PageToken = response.NextPageToken;
                     }
+                    while (!Stopping && lRequest.PageToken != null);
                     break;
                 case ParameterSetNames.Get:
-                    DatasetsResource.GetRequest gRequest = new DatasetsResource.GetRequest(Service, Project, Dataset);
+                    DatasetsResource.GetRequest gRequest = Service.Datasets.Get(Project, Dataset);
```
ILMTitan: Adding a function here as well will help readability. The two case statements would look similar and short, making them easier to understand together.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -52,65 +53,48 @@ public class GetGcbqTable : GcbqCmdlet
...
+                        WriteObject(response.Tables, true);
+                    }
+                    request.PageToken = response.NextPageToken;
                 }
+                while (!Stopping && request.PageToken != null);
             }
             else
             {
-                TablesResource.GetRequest request = new TablesResource.GetRequest(Service, Project, Dataset, Table);
+                TablesResource.GetRequest request = Service.Tables.Get(Project, Dataset, Table);
```
ILMTitan: Again, adding a function here to match the function above would help readability.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,113 @@
...
+                try
+                {
+                    var response = request.Execute();
+                    WriteObject(response);
+                }
+                catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.NotFound)
+                {
+                    WriteError(new ErrorRecord(ex,
+                        $"Error 404: Table {Table} not found in {Dataset}.",
```
quoctruong: nit: You can use {ex.HttpStatusCode} instead of hard-coding 404 here.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -0,0 +1,113 @@
...
+        /// default to the Cloud SDK's DefaultProject property.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        [ConfigPropertyName(CloudSdkSettings.CommonProperties.Project)]
+        override public string Project { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The ID of the dataset to search.  Can be a string or a dataset object.
```
quoctruong: Double space between "." and "Can be a string".

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
quoctruong: nit: I think your description shouldn't be in all caps. Maybe like "Test" and "Test Set 2".

ahandley:
@ILMTitan - Refactored to functions.  Take a look - should be a lot cleaner.

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: Got this and all of the above ^^.  Thanks

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -96,45 +97,63 @@ protected override void ProcessRecord()
...
+                    foreach(DatasetList.DatasetsData d in response.Datasets)
+                    {
+                        yield return d;
+                    }
+                    WriteObject(response.Datasets, true);
+                }
+                lRequest.PageToken = response.NextPageToken;
+            }
+            while (!Stopping && lRequest.PageToken != null);
+            yield break;
```
ILMTitan: This isn't needed. Reaching the end of the function ends the Enumeration.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -96,45 +97,63 @@ protected override void ProcessRecord()
...
+                        new Exception("List request to server responded with null."),
+                        "List request returned null", ErrorCategory.InvalidArgument, project));
+                }
+                if (response.Datasets != null)
+                {
+                    foreach(DatasetList.DatasetsData d in response.Datasets)
+                    {
+                        yield return d;
+                    }
+                    WriteObject(response.Datasets, true);
```
ILMTitan: Don't `WriteObject` here. You are doing that outside the function.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -96,45 +97,63 @@ protected override void ProcessRecord()
...
-                        WriteObject(gResponse);
-                    }
-                    catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.NotFound)
-                    {
-                        WriteError(new ErrorRecord(ex, 
-                            $"Error 404: Dataset {Dataset} not found in {Project}.",
-                            ErrorCategory.ObjectNotFound, 
-                            Dataset));
-                    }
+                    var dataset = DoGetRequest(Project, Dataset);
```
ILMTitan: This function can return null. Don't write null to the pipeline.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -96,45 +97,63 @@ protected override void ProcessRecord()
...
+            while (!Stopping && lRequest.PageToken != null);
+            yield break;
+        }
+
+        private Dataset DoGetRequest(string project, string dataset)
+        {
+            DatasetsResource.GetRequest gRequest = Service.Datasets.Get(project, dataset);
+            try
+            {
+                var gResponse = gRequest.Execute();
```
ILMTitan: What type is this var? var is best used when the left side would restate what the right side already said. `var aList = new List<SomeOtherClass>();`

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -20,8 +21,8 @@ namespace Google.PowerShell.BigQuery
     /// READER dataset role. The -All flag will include hidden datasets in the search results. The -Filter 
     /// flag allows you to filter results by label. The syntax to filter is "labels.name[:value]". Multiple filters 
     /// can be ANDed together by connecting with a space. See the link below for more information.
-    /// If no Project is specified, the default project will be used. This cmdlet returns a DatasetList if 
-    /// no DatasetId was specified, and a Dataset otherwise.
+    /// If no Project is specified, the default project will be used. This cmdlet returns any number of 
+    /// DatasetList.DatasetData objects if no DatasetId was specified, and a Dataset otherwise.
```
ILMTitan: Does a DatasetData contain a Dataset? Does DatasetData contain any additional useful information beyond what Dataset does? If the answers to those 2 questions are yes followed by no, we should just return the contained Dataset.

ahandley: So this is an issue that I found when I was looking into this.  Apparently a DatasetList.DatasetData is not a Dataset, nor does it contain a Dataset.  It's a subset of the information returned by the Get- command, which is an actual Dataset object.  The thing is, when you're interacting with them in powershell, they have the same fields that a Dataset does, so it should be functionally the same for everything except a typeof() check.  

One possible solution that I thought of was to use the fields from the DatasetData and create a somewhat incomplete Dataset and then return that, but that didn't seem like very elegant code.  Plus that'd end up taking a while if there were a large amount of projects.  

That's why I elected to just let the user know what the api returned, and thus what the cmdlet will return.  The main use case here anyway would be to get the DatasetReference.DatasetId and then pipe it back into a get request, or to just observe the string and send it manually in a non-scripted environment.  

What are your thoughts on this?

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: Missed that.  Meant to cut it after I put in the for-each and then ran into the type issues above =/

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: Not to be lazy, but it's actually probably best to just do:

return gRequest.Execute();  (this way the type is stated in the method sig)

Google.PowerShell/BigQuery/GcbqDataset.cs
ILMTitan: In that case, I would leave it as you have it, but you may want to consider adding an additional `PropertyByTypeTransformationAttribute` to the `Dataset` parameter of the `Get-GcbqTable` cmdlet to handle the additional type.

ahandley: That's a pretty good compromise.  I'll go ahead and make sure that the two cmdlets work seamlessly.

ahandley: Alright, take a look now.  Had a minor issue with pester last night so I didn't get the changes up until now.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -27,6 +27,17 @@ Describe "Get-GcbqDataset" {
         }
     }
 
+    It "should list datasets and then get complete dataset objects from the DatasetData" {
+        try {
+            New-GcbqDataset "test_id_2"
+            $a = Get-GcbqDataset | Get-GcbqDataset
```
ILMTitan: This feels silly. If we think the user may want to make this conversion, I would argue for making the extra requests to get the full Dataset object in the first place, and never letting our users see a DatasetList.DatasetsData. @quoctruong What is your opinion about this?

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -86,10 +86,14 @@ private class ParameterSetNames
...
+        /// The ID of the dataset that you want to get a descriptor object for. This field also accepts 
+        /// DatasetData objects so they can be mapped to full Dataset objects.
         /// </para>
         /// </summary>
-        [Parameter(Mandatory = true, Position = 0, ParameterSetName = ParameterSetNames.Get)]
+        [Parameter(Mandatory = true, Position = 0, ValueFromPipeline = true, ParameterSetName = ParameterSetNames.Get)]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(DatasetList.DatasetsData),
+            Property = nameof(DatasetList.DatasetsData.DatasetReference))]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(DatasetReference), Property = nameof(DatasetReference.DatasetId))]
         public string Dataset { get; set; }
```
ILMTitan: I just noticed that DatasetReference has both a DatasetId and a ProjectId. It may not be a good idea to use a single parameter to take both after all. You may give it a Dataset object for one project, and get back a Dataset object with the same name for a different project.

ahandley: In that case though, the cmdlet will always use the default GCP project.  The user would have to specify that they are using another project with the -Project flag when they got the dataset object in the first place, so they should know to use the -Project flag again, or just to switch their default dataset while they're working in another location.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -86,10 +86,14 @@ private class ParameterSetNames
...
         /// <para type="description">
-        /// The ID of the dataset that you want to get a descriptor object for.
+        /// The ID of the dataset that you want to get a descriptor object for. This field also accepts 
+        /// DatasetData objects so they can be mapped to full Dataset objects.
         /// </para>
         /// </summary>
-        [Parameter(Mandatory = true, Position = 0, ParameterSetName = ParameterSetNames.Get)]
+        [Parameter(Mandatory = true, Position = 0, ValueFromPipeline = true, ParameterSetName = ParameterSetNames.Get)]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(DatasetList.DatasetsData),
+            Property = nameof(DatasetList.DatasetsData.DatasetReference))]
```
ahandley: True.  Easy enough to add in.

Google.PowerShell/BigQuery/GcbqDataset.cs
ILMTitan: Consider:

    gcloud config set core/project your-project
    $whatTables = Get-BqDataset -Project "my-project" | Get-BqTable

`$whatTables` will now contain Tables from "your-project" even though the command was piped in Datasets from "my-project". Users will find that very surprising. Instead, we need to have two parameter sets, one that takes a -Project [string] and a -DatasetId [string], and one that takes a -Dataset [DatasetReference], because DatasetReference has both project and datasetId

ahandley: Alright, so that's looking like it's going to take a lot more thought and code fixing, because the parameter sets are going to get kinda messy (with all of the data transforms and stuff).  I'm going to just commit this and then open up a new PR for changes to address this issue once I'm done with some of the design doc changes and demo prep that I want to get done for monday.  plus this PR is huge, and it'll help to be able to focus in on these few dataset issues in their own PR instead of in a Table PR.

ILMTitan: That sounds reasonable.

PR #476, opened 2017-02-27
Get-GcbqDataset fixes from past PR

ahandley:
@ILMTitan - Here's the promised other PR to focus in on that Database issue.  

Context from last PR:

Consider:
```
gcloud config set core/project your-project
$whatTables = Get-BqDataset -Project "my-project" | Get-BqTable
```
$whatTables will now contain Tables from "your-project" even though the command was piped in Datasets from "my-project". Users will find that very surprising. Instead, we need to have two parameter sets, one that takes a -Project [string] and a -DatasetId [string], and one that takes a -Dataset [DatasetReference], because DatasetReference has both project and datasetId

### DIFF
--- a/Google.PowerShell/BigQuery/GcbqDataset.cs
+++ b/Google.PowerShell/BigQuery/GcbqDataset.cs
@@ -58,7 +58,8 @@ namespace Google.PowerShell.BigQuery
         private class ParameterSetNames
         {
             public const string List = "List";
-            public const string Get = "Get";
+            public const string GetWithString = "GetWithString";
+            public const string GetWithRef = "GetWithRef";
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


ahandley:
In additon, there was one more pending comment about returning DatasetList.DatasetsData objects from the list form of Get-

From Jim:  
This feels silly. If we think the user may want to make this conversion, I would argue for making the extra requests to get the full Dataset object in the first place, and never letting our users see a DatasetList.DatasetsData. @quoctruong What is your opinion about this?

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -95,12 +96,22 @@ private class ParameterSetNames
...
         /// </para>
         /// </summary>
-        [Parameter(Mandatory = true, Position = 0, ValueFromPipeline = true, ParameterSetName = ParameterSetNames.Get)]
+        [Parameter(Mandatory = true, Position = 0, ValueFromPipeline = true, 
+            ParameterSetName = ParameterSetNames.GetWithString)]
+        public string Dataset { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// INSERT TEXT HERE
```
ILMTitan: Text needs to be inserted.

ahandley: Hah.  I'll get on that.

ILMTitan:
LGTM.

PR #479, opened 2017-02-28
New-GcbqTable and formatting fixes

ahandley:
New cmdlet and tests, as well as a bunch of formatting fixes to dataset code and changing a parameter name to be consistent with documentation.

One area that I would like comments on - testing the -Expiration flag.  
Is there a better way to do that other than "set a table to expire 10 seconds from instantiation and wait 20 seconds after the call returns to check"?

### DIFF
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
@@ -3,7 +3,7 @@ $project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
 
 Describe "Get-GcbqDataset" {
 
-    It "should list datasets." {
+    It "should list datasets" {
         try {
             New-GcbqDataset "test_id_2"
             New-GcbqDataset "test_id_3"
@@ -16,7 +16,7 @@ Describe "Get-GcbqDataset" {
         }
     }
 
-    It "should get a dataset that exists and has permissions." {
+    It "should get a dataset that exists and has permissions" {
         try {
             New-GcbqDataset "test_id_4"
             $a = Get-GcbqDataset "test_id_4"
@@ -27,7 +27,7 @@ Describe "Get-GcbqDataset" {
         }
     }
 
-    It "should list datasets and then get complete dataset objects from the DatasetData." {
+    It "should list datasets and then get complete dataset objects from the DatasetData" {
         try {
             New-GcbqDataset "test_id_2"
             $a = Get-GcbqDataset | Get-GcbqDataset
@@ -38,7 +38,7 @@ Describe "Get-GcbqDataset" {
         }
     }
 
-    It "should get using a pre-existing Dataset (to refresh data from the cloud)." {
+    It "should get using a pre-existing Dataset (to refresh data from the cloud)" {
         try {
             New-GcbqDataset "test_id_2"
             $a = Get-GcbqDataset "test_id_2"
@@ -50,7 +50,7 @@ Describe "Get-GcbqDataset" {
         }
     }
 
-    It "should get using a DatasetReference." {
+    It "should get using a DatasetReference" {
         try {
             New-GcbqDataset "test_id_2"
             $a = Get-GcbqDataset "test_id_2"
@@ -62,15 +62,15 @@ Describe "Get-GcbqDataset" {
         }
     }
 
-    It "should handle when a dataset does not exist." {
+    It "should handle when a dataset does not exist" {
         { Get-GcbqDataset "test_id_5" -ErrorAction Stop } | Should Throw 404
     }
 
-    It "should handle projects that do not exist." {
+    It "should handle projects that do not exist" {
         { Get-GcbqDataset -Project $nonExistProject "test_id_5" -ErrorAction Stop } | Should Throw 404
     }
 
-    It "should handle projects that the user does not have permissions for." {
+    It "should handle projects that the user does not have permissions for" {
         { Get-GcbqDataset -Project $accessErrProject "test_id_5" } | Should Throw 400
     }
 }
@@ -115,7 +115,7 @@ Describe "Set-GcbqDataset" {
         }
     }
 
-    It "should not overwrite data if a set request is malformed." {
+    It "should not overwrite data if a set request is malformed" {
         try {
             New-GcbqDataset "test_dataset_id3" -Name "Testdata" -Description "Some interesting data!" -Expiration $oneDaySec
             $data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
@@ -127,7 +127,7 @@ Describe "Set-GcbqDataset" {
         }
     } 
 
-    It "should not update a set that does not exist." {
+    It "should not update a set that does not exist" {
         $data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
         $data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
         $data.DatasetReference.DatasetId = "test_dataset_id4"
@@ -142,7 +142,7 @@ Describe "Set-GcbqDataset" {
 
 Describe "New-GcbqDataset" {
 
-    It "should take a name, description, and time to make a dataset." {
+    It "should take a name, description, and time to make a dataset" {
         try {
             $data = New-GcbqDataset "test_data_id1" -Name "Testdata" -Description "Some interesting data!" -Expiration $oneDaySec
             $data | Should Not BeNullOrEmpty
@@ -156,7 +156,7 @@ Describe "New-GcbqDataset" {
         }
     }
 
-    It "should accept a dataset object from pipeline." {
+    It "should accept a dataset object from pipeline" {
         try {
             $data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
             $data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
@@ -174,7 +174,7 @@ Describe "New-GcbqDataset" {
         }
     }
 
-    It "should accept a more complex dataset object from pipeline." {
+    It "should accept a more complex dataset object from pipeline" {
         try {
             $data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
             $data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
@@ -193,7 +193,7 @@ Describe "New-GcbqDataset" {
         }
     }
 
-    It "should accept a dataset object from a command line argument." {
+    It "should accept a dataset object from a command line argument" {
         try {
             $data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
             $data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
@@ -215,15 +215,15 @@ Describe "New-GcbqDataset" {
         { New-GcbqDataset } | Should Throw 400
     }
 
-    It "should reject datasets with malformed ids." {
+    It "should reject datasets with malformed ids" {
         { New-GcbqDataset "test-?ata-4" } | Should Throw 400
     }
 
-    It "should reject datasets with empty ids." {
+    It "should reject datasets with empty ids" {
         { New-GcbqDataset "" } | Should Throw "Cannot validate argument on parameter 'DatasetId'"
     }
 
-    It "should reject datasets that do not have unique ids." {
+    It "should reject datasets that do not have unique ids" {
         try {
             $data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
             $data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
@@ -239,7 +239,7 @@ Describe "New-GcbqDataset" {
         }
     }
 
-    It "should use the -Project tag correctly." {
+    It "should use the -Project tag correctly" {
         try {
             $data = New-GcbqDataset -Project $project "test_data_id7" -Name "Testdata" -Description "Some interesting data!"
             $data | Should Not BeNullOrEmpty
@@ -250,12 +250,12 @@ Describe "New-GcbqDataset" {
         }
     }
 
-    It "should throw when you try to add to a project that doesnt exist." {
+    It "should throw when you try to add to a project that doesnt exist" {
         { New-GcbqDataset -Project $nonExistProject "test_data_id8" `
             -Name "Testdata" -Description "Some interesting data!" } | Should Throw 404
     }
 
-    It "should throw when you try to add to a project that is not yours." {
+    It "should throw when you try to add to a project that is not yours" {
         { New-GcbqDataset -Project $accessErrProject "test_data_id9" `
             -Name "Testdata" -Description "Some interesting data!" } | Should Throw 400
     }
@@ -263,7 +263,7 @@ Describe "New-GcbqDataset" {
 
 Describe "Remove-GcbqDataset" {
 
-    It "should delete an empty dataset from the pipeline with no -Force." {
+    It "should delete an empty dataset from the pipeline with no -Force" {
         try{
             New-GcbqDataset "test_set_1"
         }
@@ -272,7 +272,7 @@ Describe "Remove-GcbqDataset" {
         }
     }
 
-    It "should delete an empty dataset from an argument with no -Force." {
+    It "should delete an empty dataset from an argument with no -Force" {
         try {
             New-GcbqDataset "test_set_2"
             $a = Get-GcbqDataset "test_set_2" 
@@ -284,7 +284,7 @@ Describe "Remove-GcbqDataset" {
 
     #TODO(ahandley): Add in tests that check the -Force switch when New-GcbqTable is written.
 
-    It "should handle when a dataset does not exist." {
+    It "should handle when a dataset does not exist" {
         try {
             $data = New-Object -TypeName Google.Apis.Bigquery.v2.Data.Dataset
             $data.DatasetReference = New-Object -TypeName Google.Apis.Bigquery.v2.Data.DatasetReference
@@ -296,7 +296,7 @@ Describe "Remove-GcbqDataset" {
         }
     }
 
-    It "should handle projects that do not exist." {
+    It "should handle projects that do not exist" {
         try {
             $data = New-GcbqDataset "test_set_4"
             { Get-GcbqDataset "test_set_4" | Remove-GcbqDataset `
@@ -307,7 +307,7 @@ Describe "Remove-GcbqDataset" {
         }
     }
 
-    It "should handle projects that the user does not have permissions for." {
+    It "should handle projects that the user does not have permissions for" {
         try {
             $data = New-GcbqDataset "test_set_5"
             { Get-GcbqDataset "test_set_5" | Remove-GcbqDataset `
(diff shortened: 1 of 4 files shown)


Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
```diff
@@ -1,56 +1,117 @@
 ﻿. $PSScriptRoot\..\BigQuery\GcbqCmdlets.ps1
 $project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
 
+$testDataset = "pshell_testing"
+
 Describe "Get-GcbqTable" {
 
-    #TODO(ahandley): Include setup and teardown code in these tests as New- and Remove- are developed.
+    $d = New-GcbqDataset $testDataset
```
quoctruong: Instead of pshell_testing, you should add a random variable to prevent collision. For example:
```
$r = Get-Random;
$testDataset = "pshell_testing_$r";
$dataset = New-GcbqDataset $testDataset";
```

quoctruong: Also, we should use something meaningful for variable name. For example $dataset instead of $d.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
```diff
@@ -1,56 +1,117 @@
...
-        { Get-GcbqTable -project $nonExistProject -Dataset "pshell_testing"} | Should Throw "404"
+        { Get-GcbqTable -project $nonExistProject -Dataset $testDataset} | Should Throw "404"
+    }
+
+    $d | Remove-GcbqDataset -Force
+}
+
+Describe "New-GcbqTable" {
+
+    $d = New-GcbqDataset $testDataset
```
quoctruong: Here, you should initialize:
```
$r = Get-Random;
$testDataset = "pshell_testing_$r";
$dataset = New-GcbqDataset $testDataset";
```

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
```diff
@@ -1,56 +1,117 @@
...
+
+    It "should take a name, description, and time to make a table."{
+        $a = New-GcbqTable -Dataset $testDataset "my_table_1" -Name "CSV" -Description `
```
quoctruong: Please change $a to something meaningful like $table.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
```diff
@@ -3,7 +3,7 @@ $project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
 
 Describe "Get-GcbqDataset" {
 
-    It "should list datasets" {
+    It "should list datasets." {
```
quoctruong: I don't think you really need to add "." at the end of this. If you look at the Pester [repo](https://github.com/PowerShell/PowerShell/blob/master/docs/testing-guidelines/WritingPesterTests.md), they don't have "." at the end of It statement.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
```diff
@@ -1,56 +1,117 @@
...
     It "should throw when the dataset is not found."{
         { Get-GcbqTable -Dataset $nonExistDataset } | Should Throw "404"
     }
 
     It "should throw when the project is not found."{
-        { Get-GcbqTable -project $nonExistProject -Dataset "pshell_testing"} | Should Throw "404"
+        { Get-GcbqTable -project $nonExistProject -Dataset $testDataset} | Should Throw "404"
+    }
+
+    $d | Remove-GcbqDataset -Force
```
quoctruong: This should be in an "AfterAll" Pester block.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
```diff
@@ -1,56 +1,117 @@
...
+
+    It "should throw when the dataset is not found."{
+        { New-GcbqTable "my_table_1" -Dataset $nonExistDataset } | Should Throw "404"
+    }
+
+    It "should throw when the project is not found."{
+        { New-GcbqTable -Dataset $testDataset "my_table_1" -project $nonExistProject} | Should Throw "404"
     }
+
+    $d | Remove-GcbqDataset -Force
```
quoctruong: This should be in "AfterAll" Pester block.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -375,7 +375,7 @@ protected override void ProcessRecord()
             catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.Conflict)
             {
                 WriteError(new ErrorRecord(ex,  
-                    $"A dataset with the name {DatasetId} already exists.",
+                    $"A dataset with the name {DatasetId} already exists in this project.",
```
quoctruong: Since you actually have the Project's name, I think a better error message would be:
`$"A dataset with the name '{DatasetId}' already exists in project '{Project}'."`

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -112,4 +112,145 @@ protected override void ProcessRecord()
...
+            // Add the new dataset to the project supplied.
+            try
+            {
+                Table response = request.Execute();
+                WriteObject(response);
+            }
+            catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.Conflict)
+            {
+                WriteError(new ErrorRecord(ex,
+                    $"A table with the name {TableId} already exists in this dataset.",
```
quoctruong: It would be great if you can add the name of the dataset here:
`$"A table with the name '{TableId}' already exists in the dataset '{Dataset}'."`

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -112,4 +112,145 @@ protected override void ProcessRecord()
...
+        public string Description { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The lifetime of this table from the time of creation (in seconds).
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.ByValues)]
+        [ValidateRange(3600, (long.MaxValue / 1000))]
+        public long Expiration { get; set; }
```
quoctruong: It seems you don't have a test for this parameter. You should add it if that's the case.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -112,4 +112,145 @@ protected override void ProcessRecord()
...
+                    newTab = new Table();
+                    newTab.TableReference = new TableReference();
+                    newTab.TableReference.DatasetId = Dataset;
+                    newTab.TableReference.ProjectId = Project;
+                    newTab.TableReference.TableId = TableId;
+                    newTab.FriendlyName = Name;
+                    newTab.Description = Description;
+                    if (Expiration != 0)
+                    {
+                        double currentMillis = DateTime.Now.ToUniversalTime().Subtract(
```
quoctruong: You can use this built-in [method](https://msdn.microsoft.com/en-us/library/system.datetimeoffset.tounixtimemilliseconds(v=vs.110).aspx) to get milliseconds since epoch.

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: That's what I was asking about above.  I was planning to make one, but I don't know if there's a more elegant way than creating a table with a somewhat short life and then just calling thread.sleep() and checking later.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
ahandley: Got it!

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqDataset.Tests.ps1
ahandley: Alright, fixed that.

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: So I tried using that method:
```
 DateTimeOffset d = new DateTimeOffset(DateTime.Now);
 newTab.ExpirationTime = d.ToUnixTimeMilliseconds();
```
and the compiler is coming up with "DateTimeOffset doesnt have a method 'ToUnixTimeMilliseconds()"... 
I have System imported, and it's recognizing the data type, but just not finding the method.  Any idea what could be causing this?

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: That's what I was asking about in the comment at the top.  I didn't want to write bad code, but that's the best way I can think of at the moment.

quoctruong: You can create the table and check that the table has correct expiration time. The expiration time should be present in the table object right?

ahandley: Yes, it is.  I wasnt sure if that was a sufficient test or not.  I'll code that up then.

Google.PowerShell/BigQuery/GcbqTable.cs
quoctruong: I think it's because that method is only in .NET 4.6.  In that case, we can just go with your current code.

Google.PowerShell/BigQuery/GcbqTable.cs
ILMTitan: It is sufficient because we are not testing that BigQuery works correctly. That is someone else's job, We are testing that our API works correctly.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -303,7 +303,7 @@ private class ParameterSetNames
         /// </para>
         /// </summary>
         [Parameter(Mandatory = false, ValueFromPipeline = true, ParameterSetName = ParameterSetNames.ByObject)]
-        public Dataset Dataset { get; set; }
+        public Dataset InputObject { get; set; }
```
ILMTitan: I'm not sure why this needed to be renamed. "InputObject" is an acceptable generic name, but "Dataset" is a better more specific name. Perhaps an alias is useful?

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -112,4 +112,145 @@ protected override void ProcessRecord()
...
+        [Parameter(Mandatory = true, ValueFromPipeline = true, ParameterSetName = ParameterSetNames.ByObject)]
+        public Table InputObject { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The project to put the table in. If not set via PowerShell parameter processing, will
+        /// default to the Cloud SDK's DefaultProject property.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false)]
```
ILMTitan: This is only relevant to the parameter set ByValues.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -112,4 +112,145 @@ protected override void ProcessRecord()
...
+        [ConfigPropertyName(CloudSdkSettings.CommonProperties.Project)]
+        override public string Project { get; set; }
```
ILMTitan: This should be `public override string Project { get; set; }`

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -112,4 +112,145 @@ protected override void ProcessRecord()
...
+
+        /// <summary>
+        /// <para type="description">
+        /// The Dataset that you would like to add to. This field takes strings, Datasets, and DatasetRefrences.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, ValueFromPipeline = true, ParameterSetName = ParameterSetNames.ByValues)]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(Dataset), Property = nameof(Apis.Bigquery.v2.Data.Dataset.DatasetReference))]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(DatasetReference), Property = nameof(DatasetReference.DatasetId))]
```
ILMTitan: You probably need a third ParameterSet to take a Dataset object rather than a string, because the `DatasetReference` will specify the project for you.

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: For consistency.  I did this because whenever i have an object that should be passed in via pipeline instead of by parameter, that's what I've been calling the variable.  I can go in and add aliases, but this is also the convention with other GCP cmdlets and it should be fine as is.

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: Had that in there earlier, but it got lost in debugging.  Thanks.

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: Added this in.  Going to submit another CL tomorrow and I'm going to update all previous cmdlets to support this.  I added a section to my design doc and I'm just going to make this the standard for referencing down the project->dataset->table chain.

ahandley:
Ptal.  Everything above should have been addressed.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -112,4 +112,179 @@ protected override void ProcessRecord()
...
+                    new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc)
+                    ).TotalMilliseconds);
+                newTab.ExpirationTime = (Expiration * 1000) + currentMillis;
+                // Note: The code below is a more elegant solution, but is not supported by the current version
+                //newTab.ExpirationTime = new DateTimeOffset(DateTime.Now).ToUnixTimeMilliseconds();
+            }
+            return Service.Tables.Insert(newTab, Project, DatasetId);
+        }
+
+        public TableReference getTableRef(string p, string d, string t)
```
ILMTitan: These names should be full words: projectId, datasetId, tableId. The local variable should also probably be tableReference, but the function is short enough that it is not as important as the parameter names.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -112,4 +112,179 @@ protected override void ProcessRecord()
...
+        {
+            TableReference tr = new TableReference();
```
ILMTitan: You can make this easier with C#'s object initialization syntax

    return new TableReference { DatasetId = datasetId, ProjectId = projectId, ...};

You may not even need this function doing that.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -112,4 +112,179 @@ protected override void ProcessRecord()
...
+                WriteError(new ErrorRecord(ex,
+                    $"A table with the name {TableId} already exists in {Project}:{DatasetId}.",
+                    ErrorCategory.InvalidArgument,
+                    TableId));
+            }
+        }
+
+        public TablesResource.InsertRequest makeInsertReq()
+        {
+            Table newTab;
```
ILMTitan: There is no need to split declaration and assignment here. The variable name should be the full word newTable.

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: Yep.  Just killed the function and inlined it.  I originally made the function when I had to repeat that logic twice, but then I made a function out of the calling logic.

Google.PowerShell/BigQuery/GcbqTable.cs
ahandley: Got it.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
```diff
@@ -3,53 +3,159 @@ $project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
...
         { Get-GcbqTable -Dataset $nonExistDataset } | Should Throw "404"
     }
 
-    It "should throw when the project is not found."{
-        { Get-GcbqTable -project $nonExistProject -Dataset "pshell_testing"} | Should Throw "404"
+    It "should throw when the project is not found"{
+        { Get-GcbqTable -project $nonExistProject -Dataset $datasetName} | Should Throw "404"
+    }
+
+    AfterAll{
```
ILMTitan: Nit: Space between `AfterAll` and `{`.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.GcbqTable.Tests.ps1
```diff
@@ -3,53 +3,159 @@ $project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
...
+
+    It "should throw when the dataset is not found"{
+        { New-GcbqTable "my_table_d404" -DatasetId $nonExistDataset } | Should Throw "404"
+    }
+
+    It "should throw when the project is not found"{
+        { New-GcbqTable -DatasetId $datasetName "my_table_p404" -project $nonExistProject} | Should Throw "404"
+    }
+
+    Afterall {
```
ILMTitan: Nit: `AfterAll` pascal case.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -112,4 +112,174 @@ protected override void ProcessRecord()
...
+    /// <para type="description">
+    /// Creates a new empty table in the specified dataset. A Table can be supplied by object 
+    /// via the pipeline or the -InputObject parameter, or it can be instantiated by value 
+    /// with the flags below. The Dataset ID can be specified by passing in a string to 
+    /// -DatasetId, or you can pass a Dataset or DatasetReference to the -Dataset parameter. 
+    /// If no Project is specified, the default project will be used. This cmdlet returns 
+    /// a Table object.
+    /// </para>
+    /// <example>
+    ///   <code>PS C:\> New-GcbqTable “new_tab” -Dataset “my_data” -Description “Some nice data!”
```
ILMTitan: Nit: Code blocks preserve whitespace.  Multi-line code examples should start on a new line so the intended alignment matches the actual alignment. 

This line should also have a back-tick ( \` ) character at the end to indicate it continues on the next line. Otherwise, someone who copies and pastes this example will get errors about `-Expiration` not being a valid operator.

Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -112,4 +112,174 @@ protected override void ProcessRecord()
...
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.ByValues)]
+        [ConfigPropertyName(CloudSdkSettings.CommonProperties.Project)]
+        public override string Project { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The DatasetId that you would like to add to. This field takes strings.
+        /// To pass in a Dataset or DatasetId object for this field, use the ByValuesWithRef parameter set.
```
ILMTitan: I don't know what a DatasetId object is. Just say Dataset object. You don't need to talk about `DatasetReference` objects here.

Also, don't talk about the name of the parameter set. The name is not visible to the users. Instead, just mention the Dataset parameter.

ahandley:
Thanks!

PR #483, opened 2017-03-01
Fix a null pointer in Get-ManagedInstanceGroup

ILMTitan:


ILMTitan:
I just realized I should add a test for this. Doing that now.

ILMTitan:
Fixing the tests is hard because we have problems with regional managed instance groups.

PR #485, opened 2017-03-01
Convert remaining cmdlets

ahandley:
Made changes to Remove-Dataset and Get-Table to convert them to the new pattern of input (where datasetreferences override dataset and project id's).  Most of this was just retooling the way parameters were taken and adding parameter sets.

Involved changing parameters, so I updated docs and tests as well.

### DIFF
--- a/Google.PowerShell/BigQuery/GcbqDataset.cs
+++ b/Google.PowerShell/BigQuery/GcbqDataset.cs
@@ -176,7 +176,7 @@ namespace Google.PowerShell.BigQuery
             catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.NotFound)
             {
                 WriteError(new ErrorRecord(ex,
-                    $"Error 404: Dataset {dataset} not found in {project}.",
+                    $"Error 404: Dataset '{dataset}' not found in '{project}'.",
                     ErrorCategory.ObjectNotFound,
                     dataset));
                 return null;
@@ -242,14 +242,14 @@ namespace Google.PowerShell.BigQuery
             catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.Conflict)
             {
                 WriteError(new ErrorRecord(ex,
-                    $"Conflict while updating {InputObject.DatasetReference.DatasetId}.",
+                    $"Conflict while updating '{InputObject.DatasetReference.DatasetId}'.",
                     ErrorCategory.WriteError,
                     InputObject));
             }
             catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.Forbidden)
             {
                 WriteError(new ErrorRecord(ex,
-                    $"You do not have permission to modify {InputObject.DatasetReference.DatasetId}.",
+                    $"You do not have permission to modify '{InputObject.DatasetReference.DatasetId}'.",
                     ErrorCategory.PermissionDenied,
                     InputObject));
             }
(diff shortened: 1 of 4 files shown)


Google.PowerShell/BigQuery/GcbqTable.cs
```diff
@@ -41,74 +41,122 @@ namespace Google.PowerShell.BigQuery
     [Cmdlet(VerbsCommon.Get, "GcbqTable")]
     public class GetGcbqTable : GcbqCmdlet
     {
+        private class ParameterSetNames
+        {
+            public const string ByValue = "ByValue";
+            public const string ByHybrid = "ByHybrid";
```
ILMTitan: I think `ByDatasetObject` would be a better name.

ahandley: Yeah, that sounds more descriptive.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -455,14 +487,14 @@ protected override void ProcessRecord()
...
-                            new TablesResource.ListRequest(Service, Project, InputObject.DatasetReference.DatasetId);
+                            new TablesResource.ListRequest(Service, Project, Dataset);
                         var tableResponse = checkTables.Execute();
                         if (tableResponse.TotalItems == 0)
                         {
                             response = request.Execute();
                         }
                         else if (ShouldContinue(
-                            GetConfirmationMessage(InputObject, tableResponse.TotalItems),
+                            GetConfirmationMessage(Dataset, tableResponse.TotalItems),
```
quoctruong: Normally if you use ShouldContinue then you should provider a -Force switch. Otherwise, the user has no way of bypassing it (if they are in an environment where they cannot use prompt).

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -474,16 +506,16 @@ protected override void ProcessRecord()
...
+                        Dataset));
                 } 
             }
         }
 
-        private string GetConfirmationMessage(Dataset d, int? tables)
+        private string GetConfirmationMessage(string datasetId, int? tables)
         {
-            return $@"{d.DatasetReference.DatasetId} has {tables} tables and the -Force parameter was not specified. "
+            return $@"{datasetId} has {tables} tables and the -Force parameter was not specified. "
```
quoctruong: You should add single quote (') around {datasetId} and {tables} to make it more readable.

Google.PowerShell/BigQuery/GcbqDataset.cs
```diff
@@ -440,10 +460,22 @@ public class RemoveGcbqDataset : GcbqCmdlet
...
+                    break;
+                case ParameterSetNames.ByValues:
+                    request = Service.Datasets.Delete(Project, Dataset);
+                    break;
+                default:
+                    throw UnknownParameterSetException;
+            }
             String response = "Dataset Removal Stopped.";
-            if (ShouldProcess(InputObject.DatasetReference.DatasetId))
+            if (ShouldProcess(Dataset))
```
quoctruong: Can you add a test to check that if you run the cmdlet with -Whatif then the data is not deleted?

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: I do.  It's checked above the surrounding else (line 482).  That way if -Force is supplied, we don't waste time doing a check to see if the table is empty first.  It's just deleted anyway.

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: Got it.  Filled that in for all of the other error messages where I had the same formatting as well.

Google.PowerShell/BigQuery/GcbqDataset.cs
ahandley: Added this.  It works!

Google.PowerShell/BigQuery/GcbqDataset.cs
quoctruong: My bad. I forgot it didn't show up in the CR.

ahandley: Oh true.  It only grabs a few lines above.  No worries!

PR #489, opened 2017-03-02
remove-bqtable cmdlet & tests

ahandley:
Remove-BqTable cmdlet + integration tests.  

(also changed over a few Gcbq*'s to Bq*'s that I missed in the previous commit)

### DIFF
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqTable.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqTable.Tests.ps1
@@ -183,8 +183,8 @@ Describe "Remove-BqTable" {
 
     It "should delete an empty table from an argument with no -Force" {
         New-BqTable -Dataset $test_set "table_empty_arg"
-        $a = Get-BqTable -Dataset $test_set "table_empty_arg" 
-        Remove-BqTable -InputObject $a
+        $table = Get-BqTable -Dataset $test_set "table_empty_arg" 
+        Remove-BqTable -InputObject $table
         { Get-BqTable -Dataset $test_set "table_empty_arg" -ErrorAction Stop } | Should Throw 404
     }
 
@@ -202,19 +202,22 @@ Describe "Remove-BqTable" {
         $table.TableReference.DatasetId = $test_set
         $table.TableReference.ProjectId = $project
         $table.TableReference.TableId = "table_not_actually_there_for_some_reason"
-        { Remove-BqTable -InputObject $table } | Should Throw 404
+        { Remove-BqTable -InputObject $table -ErrorAction Stop } | Should Throw 404
     }
 
     It "should handle projects that do not exist" {
-        { Remove-BqTable -Project $nonExistProject -DatasetId $nonExistDataset -Table "table" } | Should Throw 404
+        { Remove-BqTable -Project $nonExistProject -DatasetId $nonExistDataset `
+            -Table "table" -ErrorAction Stop } | Should Throw 404
     }
 
     It "should handle project:dataset combinations that do not exist" {
-        { Remove-BqTable -Project $project -DatasetId $nonExistDataset -Table "table" } | Should Throw 404
+        { Remove-BqTable -Project $project -DatasetId $nonExistDataset `
+            -Table "table" -ErrorAction Stop } | Should Throw 404
     }
 
     It "should handle projects that the user does not have permissions for" {
-        { Remove-BqTable -Project $accessErrProject -DatasetId $nonExistDataset -Table "table" } | Should Throw 400
+        { Remove-BqTable -Project $accessErrProject -DatasetId $nonExistDataset `
+            -Table "table" -ErrorAction Stop } | Should Throw 400
     }
 
     AfterAll {
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqTable.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqTable.Tests.ps1
@@ -1,61 +1,61 @@
-﻿. $PSScriptRoot\..\BigQuery\GcbqCmdlets.ps1
+﻿. $PSScriptRoot\..\BigQuery\BqCmdlets.ps1
 $project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
 
-Describe "Get-GcbqTable" {
+Describe "Get-BqTable" {
 
     BeforeAll {
         $r = Get-Random
         $datasetName = "pshell_testing_$r"
         $test_set = New-BqDataset $datasetName
-        $test_set | New-GcbqTable "my_table" -Name "my_table" -Description "Test data table"
-        $test_set | New-GcbqTable "my_other_table" -Name "my_other_table" -Description "Another test table"
+        $test_set | New-BqTable "my_table" -Name "my_table" -Description "Test data table"
+        $test_set | New-BqTable "my_other_table" -Name "my_other_table" -Description "Another test table"
     }
 
     It "should list tables from a dataset object from pipeline"{
-        $tables = Get-BqDataset $datasetName | Get-GcbqTable
+        $tables = Get-BqDataset $datasetName | Get-BqTable
         $tables.Count | Should BeGreaterThan 0
     }
 
     It "should list tables from a dataset object via parameter"{
         $dataset = Get-BqDataset $datasetName 
-        $tables = Get-GcbqTable -Dataset $dataset
+        $tables = Get-BqTable -Dataset $dataset
         $tables.Count | Should BeGreaterThan 0
     }
 
     It "should list tables from a dataset ID by parameter"{
-        $tables = Get-GcbqTable -DatasetId $datasetName
+        $tables = Get-BqTable -DatasetId $datasetName
         $tables.Count | Should BeGreaterThan 0
     }
 
     It "should get a singular table with a dataset object from pipeline"{
-        $table = Get-BqDataset $datasetName | Get-GcbqTable "my_table"
+        $table = Get-BqDataset $datasetName | Get-BqTable "my_table"
         $table.TableReference.TableId | Should be "my_table"
         $table.Description | Should Be "Test data table"
     }
 
     It "should get a singular table with a dataset object via parameter"{
         $dataset = Get-BqDataset $datasetName
-        $table = Get-GcbqTable -Dataset $dataset "my_other_table"
+        $table = Get-BqTable -Dataset $dataset "my_other_table"
         $table.TableReference.TableId | Should be "my_other_table"
         $table.Description | Should Be "Another test table"
     }
 
     It "should get a singular table with a dataset ID by parameter"{
-        $table = Get-GcbqTable -DatasetId $datasetName "my_table"
+        $table = Get-BqTable -DatasetId $datasetName "my_table"
         $table.TableReference.TableId | Should be "my_table"
         $table.Description | Should Be "Test data table"
     }
 
     It "should throw when the table is not found"{
-        { Get-GcbqTable -DatasetId $datasetName $nonExistTable -ErrorAction Stop} | Should Throw "404"
+        { Get-BqTable -DatasetId $datasetName $nonExistTable -ErrorAction Stop} | Should Throw "404"
     }
 
     It "should throw when the dataset is not found"{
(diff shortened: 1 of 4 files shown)


Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -330,4 +330,163 @@ public TablesResource.InsertRequest makeInsertReq()
...
+        {
+            public const string ByValue = "ByValue";
+            public const string ByDatasetObject = "ByDatasetObject";
+            public const string ByObject = "ByObject";
+        }
+
+        /// <summary>
+        /// <para type="description">
+        /// The project to look for tables in. If not set via PowerShell parameter processing, it will
+        /// default to the Cloud SDK's DefaultProject property.
```
ILMTitan: DefaultProject is not the name of the property (technically it is core/project). You could just say "the Cloud SDK's default project." We can assume our users will know what that means.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -330,4 +330,163 @@ public TablesResource.InsertRequest makeInsertReq()
...
+        /// The ID of the table that you want to get a descriptor object for.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, Position = 0, ParameterSetName = ParameterSetNames.ByValue)]
+        [Parameter(Mandatory = true, Position = 0, ParameterSetName = ParameterSetNames.ByDatasetObject)]
+        public string Table { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The ID of the dataset to search. Can be a string, a Dataset, a DatasetReference, or a DatasetsData object.
```
ILMTitan: It can only be a string.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -330,4 +330,163 @@ public TablesResource.InsertRequest makeInsertReq()
...
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, ValueFromPipeline = true, ParameterSetName = ParameterSetNames.ByDatasetObject)]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(Dataset),
+            Property = nameof(Apis.Bigquery.v2.Data.Dataset.DatasetReference))]
+        public DatasetReference Dataset { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The Table object that will be sent to the server to be inserted. Also takes TableReferences.
```
ILMTitan: Removed, not inserted. Mention Table object in the comments, not TableReference. TableReference is the actual type of the parameter, and is visible to the user. Table is not visible so it needs the comment.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -330,4 +330,163 @@ public TablesResource.InsertRequest makeInsertReq()
...
+                    throw UnknownParameterSetException;
+            }
+
+            TablesResource.DeleteRequest request = Service.Tables.Delete(Project, DatasetId, Table);
+            String response = "Table Removal Stopped.";
+            if (ShouldProcess(Table))
+            {
+                try
+                {
+                    if (Force == true)
```
ILMTitan: This is a personal preference, but I don't like `BoolVar == boolLiteral` when `BoolVar` or `!BoolVar` would do.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -330,4 +330,163 @@ public TablesResource.InsertRequest makeInsertReq()
...
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        public SwitchParameter Force { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            switch (ParameterSetName)
+            {
+                case ParameterSetNames.ByValue:
+                    // No processing needed for ByValue parameter sets.
```
ILMTitan: Good comment.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -330,4 +330,163 @@ public TablesResource.InsertRequest makeInsertReq()
...
+                    else
+                    {
+                        TablesResource.GetRequest checkTables =
+                            new TablesResource.GetRequest(Service, Project, DatasetId, Table);
+                        var tableResponse = checkTables.Execute();
+                        if (tableResponse.NumRows == 0)
+                        {
+                            response = request.Execute();
+                        }
+                        else if (ShouldContinue(
```
ILMTitan: You may want to use the 4 parameter version of ShouldContinue. It may look like it is only called once, but if you pipeline in a bunch of non-empty table objects, a Yes to All/No to All option would be nice.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqTable.Tests.ps1
```diff
@@ -136,22 +136,85 @@ Describe "New-GcbqTable" {
...
+    
+    It "should delete an empty table from the pipeline with no -Force" {
+        New-BqTable -Dataset $test_set -Table "table_empty_pipe"
+        Get-BqTable -Dataset $test_set "table_empty_pipe" | Remove-BqTable
+        { Get-BqTable -Dataset $test_set "table_empty_pipe" -ErrorAction Stop } | Should Throw 404
+    }
+
+    It "should delete an empty table from an argument with no -Force" {
+        New-BqTable -Dataset $test_set "table_empty_arg"
+        $a = Get-BqTable -Dataset $test_set "table_empty_arg"
```
quoctruong: nit: Change variable name to $table instead of $a.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -330,4 +330,163 @@ public TablesResource.InsertRequest makeInsertReq()
...
+        /// default to the Cloud SDK's DefaultProject property.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.ByValue)]
+        [ConfigPropertyName(CloudSdkSettings.CommonProperties.Project)]
+        override public string Project { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The ID of the table that you want to get a descriptor object for.
```
quoctruong: The description is wrong. Should be the table that you want to remove.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -330,4 +330,163 @@ public TablesResource.InsertRequest makeInsertReq()
...
+        /// The ID of the dataset to search. Can be a string, a Dataset, a DatasetReference, or a DatasetsData object.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, ParameterSetName = ParameterSetNames.ByValue)]
+        [ValidatePattern("[a-zA-Z0-9_]")]
+        public string DatasetId { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The Dataset that you would like to search. This field takes Dataset or DatasetRefrence objects.
```
quoctruong: I think you should mention something about the table that is being removed (and its connection to the dataset) in this comment.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -330,4 +330,163 @@ public TablesResource.InsertRequest makeInsertReq()
...
+                        else if (ShouldContinue(
+                            GetConfirmationMessage(Table, tableResponse.NumRows),
+                            "Confirm Deletion"))
+                        {
+                            response = request.Execute();
+                        }
+                    }
+                    WriteObject(response);
+                }
+                catch (GoogleApiException ex) when (ex.HttpStatusCode == HttpStatusCode.Forbidden)
```
quoctruong: Is there a different HttpStatusCode for table not found vs forbidden? It seems to me that forbidden means you don't have access to the dataset.

ahandley: Yep.  Forgot to change the text.

Google.PowerShell/BigQuery/BqTable.cs
ahandley: "This dataset should contain the table you wish to remove."  added to both this variable and DatasetId.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqTable.Tests.ps1
```diff
@@ -136,22 +136,85 @@ Describe "New-GcbqTable" {
...
+Describe "Remove-BqTable" {
+
+    BeforeAll {
+        $r = Get-Random
+        $datasetName = "pshell_testing_$r"
+        $test_set = New-BqDataset $datasetName
+    }
+
+    It "should not delete the table if -WhatIf is specified" {
+        $table = New-BqTable -Dataset $test_set -Table "table_if"
```
ahandley: I randomize the datasetId and each table is named for the test case that it's in.  When each set of tests is finished, the dataset is deleted, so there should never be collisions.  I'm not opposed to adding it, but I think it'll keep the test cases cleaner and more concise without it.

Google.PowerShell/BigQuery/BqTable.cs
ahandley: Looking into this.  I agree.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqTable.Tests.ps1
quoctruong: I'm okay with not adding it in that case.

ahandley:
PTAL.  All changes should have been addressed.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -451,10 +458,14 @@ protected override void ProcessRecord()
...
                 try
                 {
-                    if (Force == true)
+                    if (Force)
                     {
                         response = request.Execute();
                     }
+                    else if (noToAll)
+                    {
+                        response = "Removal Stopped: NoToAll";
```
ILMTitan: We don't want to pollute the the output pipeline with an information string. Just check `noToAll` in the same `if` as `ShouldProcess`, and skip everything once it has been set.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -464,19 +475,19 @@ protected override void ProcessRecord()
                         {
                             response = request.Execute();
                         }
-                        else if (ShouldContinue(
+                        else if (yesToAll || ShouldContinue(
```
ILMTitan: I am pretty sure you can skip the empty check once yesToAll is set. Treat yesToAll like `Force`, and move it to that `if`.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -330,4 +330,174 @@ public TablesResource.InsertRequest makeInsertReq()
...
+                case ParameterSetNames.ByDatasetObject:
+                    Project = Dataset.ProjectId;
+                    DatasetId = Dataset.DatasetId;
+                    break;
+                default:
+                    throw UnknownParameterSetException;
+            }
+
+            TablesResource.DeleteRequest request = Service.Tables.Delete(Project, DatasetId, Table);
+            String response = "Table Removal Stopped.";
```
ILMTitan: We don't want to pollute the output pipeline with an information string. I would refactor this code with a new function contains all of the should delete logic and returns a boolean. Then, if it returns true, execute the delete request. I'm not sure this cmdlet should write anything to the pipeline. The response says it is a string, but I suspect it is always an empty one.

ahandley:
Refactored that method and addressed the other changes.  It looks much cleaner now.

Google.PowerShell/BigQuery/BqTable.cs
```diff
@@ -452,51 +451,46 @@ protected override void ProcessRecord()
...
+            else if (Force || yesToAll)
+            {
+                return true;
+            }
+            else
+            {
+                var tableResponse = new TablesResource.GetRequest(
+                    Service, Project, DatasetId, Table).Execute();
+
+                return (tableResponse.NumRows == 0) ? true :
```
ILMTitan: `(tableResponse.NumRows == 0) || ShouldContinue` might be a little clearer, but that is debatable.

ahandley: That's probably true.  Less obfuscating characters.

PR #496, opened 2017-03-07
fixed git issue.  Also New-Schema

ahandley:
Lots of junk XML with this release because I had a minor snafu with the NuGet and our VCS.  Should be all resolved now, and then the code for New-BqSchema is here.


Google.PowerShell.IntegrationTests/packages.config
```diff
@@ -1,10 +1,10 @@
...
-  <package id="Google.Apis.Core" version="1.20.0" targetFramework="net452" />
+  <package id="Google.Apis" version="1.21.0" targetFramework="net452" />
+  <package id="Google.Apis.Auth" version="1.21.0" targetFramework="net452" />
+  <package id="Google.Apis.Bigquery.v2" version="1.21.0.785" targetFramework="net452" />
+  <package id="Google.Apis.Core" version="1.21.0" targetFramework="net452" />
   <package id="log4net" version="2.0.3" targetFramework="net452" />
   <package id="Newtonsoft.Json" version="7.0.1" targetFramework="net452" />
   <package id="Zlib.Portable.Signed" version="1.11.0" targetFramework="net452" />
-</packages>
+</packages>
```
quoctruong: Missing a line here.

Google.PowerShell.IntegrationTests/app.config
```diff
@@ -0,0 +1,15 @@
...
+        <assemblyIdentity name="Google.Apis" publicKeyToken="4b01fa6e34db77ab" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-1.21.0.0" newVersion="1.21.0.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="Google.Apis.Core" publicKeyToken="4b01fa6e34db77ab" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-1.21.0.0" newVersion="1.21.0.0" />
+      </dependentAssembly>
+    </assemblyBinding>
+  </runtime>
+</configuration>
```
quoctruong: Missing a line here.

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -0,0 +1,135 @@
...
+                InputObject = new TableSchema();
+                InputObject.Fields = new List<TableFieldSchema>();
+            }
+
+            if (Mode == null)
+            {
+                Mode = "NULLABLE";
+            }
+
+            foreach (TableFieldSchema t in InputObject.Fields)
```
quoctruong: Instead of foreach loop, you can do a LINQ expression like:
```if (InputObject.Fields.Any(field => Name.Equals(field.Name)))```

Comment (quoctruong) [inline Google.PowerShell/BigQuery/BqTabledata.cs]: Also, in this case, should we write a terminating error to the user instead of just a normal error? Since the object returned is not going to be valid anyway.

File: Google.PowerShell/BigQuery/BqTabledata.cs [hunk 4]
```diff
@@ -0,0 +1,135 @@
...
+    /// PS C:\> $table | Set-BqTable
+    ///   </code>
+    ///   <para>This will create a new schema, assign it to a table, and then send the 
+    ///   revised table to the server to be saved.</para>
+    /// </example> 
+    /// <para type="link" uri="(https://cloud.google.com/bigquery/docs/reference/rest/v2/tables)">
+    /// [BigQuery Tables]
+    /// </para>
+    /// </summary>
+    [Cmdlet(VerbsCommon.New, "BqSchema")]
```
quoctruong: I feel like this cmdlet should probably be named "Add-BqFieldSchema" instead of "New-BqSchema" (the -InputObject parameter can be named "BqSchema"). @chrsmith and @ILMTitan what do you guys think?

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -0,0 +1,135 @@
...
+        public string Name { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The type of the column to be added. Possible values include STRING, BYTES, INTEGER (INT64), 
+        /// FLOAT (FLOAT64), BOOLEAN (BOOL), TIMESTAMP, DATE, TIME, DATETIME, and RECORD (STRUCT).
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true)]
+        [ValidateSet("STRING", "BYTES", "INTEGER", "INT64", "FLOAT", "FLOAT64", "BOOLEAN", "BOOL",
```
quoctruong: Should we set IgnoreCase = true here too?

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -0,0 +1,135 @@
...
+        public string Description { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The mode of the column to be added. Possible values include NULLABLE, REQUIRED, and REPEATED.
+        /// The default value is NULLABLE.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        [ValidateSet("NULLABLE", "REQUIRED", "REPEATED", IgnoreCase = true)]
```
quoctruong: If you are going to use these values elsewhere (like in a related cmdlet), I recommend using an enum. Same for Type parameter.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqTabledata.Tests.ps1
```diff
@@ -0,0 +1,66 @@
...
+        $outer = New-BqSchema -Name "Nest" -Type "RECORD" -Fields $inner
+        $outer.Fields[0].Fields.Count | Should Be 2
+    }
+
+    It "should deny invalid types"{
+        { New-BqSchema -Name "Title" -Type "NotAType" -ErrorAction Stop} | Should Throw
+    }
+
+    It "should deny invalid modes"{
+        { New-BqSchema -Name "Title" -Type "STRING" -Mode "NotAMode" -ErrorAction Stop} | Should Throw
```
quoctruong: What's the error message here? Same for "invalid types" and "duplicated column names" tests.

Google.PowerShell/BigQuery/BqTabledata.cs
ahandley: Yep.  Was having issues with the parameter tags and forgot to put that back in after testing.

Google.PowerShell/BigQuery/BqTabledata.cs
ahandley: Was reading into see if i could just supply an array of strings, but that would fall under the same logic.  This thread seems to think not.

**http://stackoverflow.com/questions/36656171/can-i-use-a-constant-in-the-validateset-attribute-of-a-powershell-function-param**

Similarly, there was another article that I found saying that dynamic parameters was the only way to not have to hard code values.  I can look into that, but I don't think there are going to be any other cmdlets using these params.  Once they're sent the first time, they're only accessible through viewing the table objects and hopefully shouldn't ever be touched again.

quoctruong: What I meant was something like [this](https://github.com/GoogleCloudPlatform/google-cloud-powershell/blob/master/Google.PowerShell/Logging/GcLogCmdlets.cs#L26) which is used [here](https://github.com/GoogleCloudPlatform/google-cloud-powershell/blob/master/Google.PowerShell/Logging/GcLogCmdlets.cs#L260)

ahandley: Oh, that makes a lot more sense.  I'll see what I can do with that.

Google.PowerShell/BigQuery/BqTabledata.cs
ahandley: Yep.  As I was writing this I was thinking about a lambda expression kinda thing, but I wasn't confident in my C# foo.  

And about the errors, yes.  There's no point in continuing on after that, so it'll be better not to harm the original array.  I'll add something to the docs about that.

Google.PowerShell.IntegrationTests/app.config
ahandley: No idea why it isnt showing that this one is fixed.  If you view the file, the newline is there.

Google.PowerShell/BigQuery/BqTabledata.cs
ahandley: I think "Add-BqSchemaField" would sound better, but I agree on the switch from New- to Add-.

ahandley:
Made changes to the above, with the one exception of the name change.  Waiting for further comment there.

ahandley:
Added Set-BqSchema and refactored the integration test from New- that had a TODO on it.  

PTAL specifically that new cmdlet & tests while the discussion is still happening for the naming conventions.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqTabledata.Tests.ps1
```diff
@@ -0,0 +1,87 @@
...
+        $table.TableReference.TableId = "not_gonna_happen_today"
+        $schema = New-BqSchema -Name "Title" -Type "STRING"
+        { $schema | Set-BqSchema $table } | Should Throw 404
+    } 
+
+    AfterAll {
+        $test_set | Remove-BqDataset -Force
+    }
+}
+
```
ILMTitan: Remove extra line.

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -0,0 +1,223 @@
...
+            Table.ETag = null;
+            Table response;
+            var request = Service.Tables.Update(Table,
+                Table.TableReference.ProjectId,
+                Table.TableReference.DatasetId,
+                Table.TableReference.TableId);
+            try
+            {
+                response = request.Execute();
+                WriteObject(response);
```
ILMTitan: The way this is written, I think we *do* want return the table to the pipeline, but only once, the last time.

Google.PowerShell/BigQuery/BqTabledata.cs
ILMTitan: I would propose a larger change, where the cmdlet remains New-BqSchema, but it returns 'TableFieldSchema' objects. 'TableSchema' is kind of useless, being just a wrapper around a list of 'TableFieldSchema' objects. This cmdlet's `ProcessRecord()` would just pass along any InputObjects, and then `EndProcessing()` would create a new one and append it to the pipeline. Set-BqSchema would then gather all of the pipeline objects, and only make the update call in `EndProcessing()`.
Look at New-GceFirewallProtocol and Add-GceFirewall for an example of what I am thinking of.

Google.PowerShell/BigQuery/BqTabledata.cs
ahandley: So I can't respond to the previous comment, but I agree with that new layout and it won't be hard to implement.  Should the cmdlet be called "New-BqSchemaField"?  Or is it better to keep the name the same for both even if it's slightly less accurate?  

What does everyone else think?

ILMTitan: I think New-BqSchema still works. Neither of the objects in question are just `Schema` and both have Schema in the name.

ahandley:
Aight.  Finally got that changeset working.  It's much cleaner this way too.

PTAL.

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -190,18 +185,34 @@ public class SetBqSchema : BqCmdlet
...
+        List<TableFieldSchema> Columns = new List<TableFieldSchema>();
+
         protected override void ProcessRecord()
         {
-            Table.Schema = InputObject;
-            Table.ETag = null;
-            Table response;
+            if (Columns.Any(field => InputObject.Name.Equals(field.Name)))
+            {
+                // ATTENTION:  should this throw? (to notify users that data may have been lost)
```
ILMTitan: I agree with throwing here.

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -65,12 +66,11 @@ public class NewBqSchema : PSCmdlet
...
         /// <summary>
         /// <para type="description">
-        /// The existing TableSchema that you wish to add a column to. 
-        /// If this value is not present, a new schema will be created.
+        /// Holder parameter to allow cmdlet to forward TableFieldSchemas down the pipeline.
         /// </para>
         /// </summary>
         [Parameter(Mandatory = false, ValueFromPipeline = true)]
-        public TableSchema InputObject { get; set; }
+        public TableFieldSchema InputObject { get; set; }
```
ILMTitan: I would consider changing the name of this to `PassThruObject`.

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -37,22 +37,21 @@ public enum ColumnMode
...
     /// </para>
     /// <example>
     ///   <code>
-    /// PS C:\> $table = Get-BqDataset “book_data” | Get-BqTable "21st_century"
-    /// PS C:\> $schema = New-BqSchema -Name "Title" -Type "STRING" -Description "Book Title"
-    /// PS C:\> $schema = $schema | New-BqSchema -Name "Author" -Type "STRING" -Description "Book Author"
-    /// PS C:\> $table.Schema = $schema
-    /// PS C:\> $table | Set-BqTable
+    /// PS C:\> $table = $dataset_books | New-BqTable "book_info"
+    /// PS C:\> $result = New-BqSchema -Name "Author" -Type "STRING" | `
```
ILMTitan: Fyi, ending a line with a pipe is one of the few times powershell does not need a backtick to know the statement continues on the next line.

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -0,0 +1,233 @@
...
+    [Cmdlet(VerbsCommon.New, "BqSchema")]
+    public class NewBqSchema : PSCmdlet
+    {
+        /// <summary>
+        /// <para type="description">
+        /// Holder parameter to allow cmdlet to forward TableFieldSchemas down the pipeline.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ValueFromPipeline = true)]
+        public TableFieldSchema InputObject { get; set; }
```
ILMTitan: I would consider renaming this to `PassThruObject`

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -0,0 +1,233 @@
...
+        [ValidateNotNull]
+        public Table Table { get; set; }
+
+        List<TableFieldSchema> Columns = new List<TableFieldSchema>();
+
+        protected override void ProcessRecord()
+        {
+            if (Columns.Any(field => InputObject.Name.Equals(field.Name)))
+            {
+                // ATTENTION:  should this throw? (to notify users that data may have been lost)
```
ILMTitan: I agree with throwing here.

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -0,0 +1,233 @@
...
+    /// column are Name and Type. Possible values for Type include STRING, BYTES, INTEGER, 
+    /// FLOAT, BOOLEAN, TIMESTAMP, DATE, TIME, DATETIME, and RECORD (where RECORD indicates 
+    /// that the field contains a nested schema). Case is ignored for both Type and Mode. 
+    /// This command forwards all TableFieldSchemas that it is passed, and will add a new 
+    /// TableFieldSchema object to the pipeline.
+    /// </para>
+    /// <example>
+    ///   <code>
+    /// PS C:\> $table = $dataset_books | New-BqTable "book_info"
+    /// PS C:\> $result = New-BqSchema -Name "Author" -Type "STRING" | `
```
ILMTitan: FYI, ending a line with a pipe is one of the few way to indicate the statement continues on the next line without needing a backtick.

ahandley: Oh cool.  That makes things a lot cleaner.

Google.PowerShell/BigQuery/BqTabledata.cs
ahandley: Fair enough.  I just needed a name since the users should never be using the flag explicitly.

PR #497, opened 2017-03-08
Add New-GkeNodeConfig cmdlet

quoctruong:
This cmdlet creates a Google Container NodeConfig object. This object will be used in the cmdlet New-GkeContainerCluster (to be created after this) to configure nodes in a container cluster.
Also enable New-GceServiceAccountConfig to use the project's [default service account](https://cloud.google.com/compute/docs/access/service-accounts#compute_engine_default_service_account) when service account email is not provided.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/497)

### DIFF
--- a/Google.PowerShell/Container/GkeClusterCmdlets.cs
+++ b/Google.PowerShell/Container/GkeClusterCmdlets.cs
@@ -170,12 +170,333 @@ namespace Google.PowerShell.Container
     }
 
     /// <summary>
+    /// Abstract class for cmdlets that deal with node configuration such as
+    /// New-GkeNodeConfig and Add-GkeCluster.
+    /// </summary>
+    public abstract class GkeNodeConfigCmdlet : GkeCmdlet, IDynamicParameters
+    {
+        // IAM Service used for getting roles that can be granted to a project.
+        private Lazy<ComputeService> _computeService =
+            new Lazy<ComputeService>(() => new ComputeService(GetBaseClientServiceInitializer()));
+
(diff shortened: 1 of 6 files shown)


Google.PowerShell/Container/GkeClusterCmdlets.cs
```diff
@@ -161,4 +168,390 @@ private IEnumerable<Cluster> GetClustersByZone(string project, string zone)
...
         }
     }
+
+    /// <summary>
+    /// <para type="synopsis">
+    /// Creates a Google Container Engine Node Config.
+    /// </para>
+    /// <para type="description">
+    /// Creates a Google Container Engine Node Config. The node config is used to configure various properties
+    /// of a node in a container cluster so you can use the object returned by the cmdlet in New-GkeCluster
```
ILMTitan: New-GkeCluster, or Add-GkeCluster? We do mix these two, but I do like the convention we have in places where `New` creates a new object, while `Add` actually puts things on GCP.

quoctruong:
I refactored the code so they can be reused in Add-GkeCluster.
The ```List<Attribute>``` is actually not an instance of ```Collection<Attribute>``` (Collection here is [System.ObjectModel.Collection](https://msdn.microsoft.com/en-us/library/ms132397(v=vs.110).aspx)).
@ILMTitan, @ahandley PTAL

quoctruong:
Thanks!

PR #506, opened 2017-03-24
Get-BqTabledata + tests

ahandley:
Last Tabledata cmd + tests + Format.ps1xml

bqr

Powershell output / example of format.ps1xml


Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -431,4 +431,76 @@ protected override void ProcessRecord()
...
+        protected override void ProcessRecord()
+        {
+            var Client = BigQueryClient.Create(InputObject.ProjectId);
+            
+            try
+            {
+                var response = Client.ListRows(InputObject, null,
+                new ListRowsOptions() { PageSize = PageSize });
+
+                foreach (var row in response)
```
quoctruong: You can just do WriteObject(response, true). Also, should we check whether response is null or not?

Google.PowerShell/ReleaseFiles/GoogleCloudPlatform.Format.ps1xml
```diff
@@ -162,5 +162,52 @@
         </TableRowEntries>
       </TableControl>
     </View>
+    <View>
+      <Name>BigQuery Data Row</Name>
```
quoctruong: Can you post a screenshot of what the output looks like?

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -431,4 +431,76 @@ protected override void ProcessRecord()
...
+                {
+                    WriteObject(row);
```
quoctruong: One thing to consider here (especially since it's likely that the data size is big), is to use WriteObject whenever you get a batch of result (like when you use the Google.Apis.BigQuery call  with PageToken). That way, we will output the rows as soon as some arrive instead of waiting for all of them.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqTabledata.Tests.ps1
```diff
@@ -216,4 +216,54 @@ Describe "Add-BqTabledata" {
...
+    }
+
+    It "should return an entire table of 10 rows" {
+        $list = $table | Get-BqTabledata 
+        $list.Count | Should Be 10
+    }
+
+    It "should paging correctly" {
+        $list = $table | Get-BqTabledata -PageSize 3
+        $list.Count | Should Be 10
```
quoctruong: Should we check one or two row in the list?

Google.PowerShell/ReleaseFiles/GoogleCloudPlatform.Format.ps1xml
ahandley: Added it to the original comment.

Google.PowerShell/BigQuery/BqTabledata.cs
ahandley: Added the check, and refactored that code.  I had it as a foreach because that was the formatting recommended by the veneer page (outside of PS).

Google.PowerShell/BigQuery/BqTabledata.cs
ahandley: The way that ListRows() works is it returns a lazy-loading set that handles all of the paging in the backend.  To Powershell, it looks like there's just a magical big set and the iterator always has a .next().  But it's actually just paging when the iterator hits the end, or gets close, and feeding the results into whatever buffer the iterator is looking through.

Thus, we should already be sending the results as soon as they arrive.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqTabledata.Tests.ps1
ahandley: Probably.  I put that in.

ahandley:
Aight.  Changes applied - PTAL.

Google.PowerShell/ReleaseFiles/GoogleCloudPlatform.Format.ps1xml
quoctruong: I think a better way would be for the output to look like a table like:
Book Name            Author          Published
-----------------------------------------------
Omelette              Shakespeare   1603

Do you think it would be possible?

quoctruong: I'm referring to this [TableView](https://msdn.microsoft.com/en-us/library/dd901841(v=vs.85).aspx)

ahandley: That's totally doable, but then what it ends up looking like is 

Title, Author, Year
Omelette, Shakespeare, 1603
Title, Author, Year
Omelette, Shakespeare, 1603
Title, Author, Year
Omelette, Shakespeare, 1603

...etc.  I tried to control the formatting of the List<BigQueryRow>, but that didn't work well at all.  Jim sent me a few links, and I'm looking into what options we have.

quoctruong: That's weird. I'm fine if we fix it in another check-in though. Can you add a TODO for it?

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -431,4 +431,78 @@ protected override void ProcessRecord()
...
+    /// You can specify how many results to return per page.  This will affect how fast the 
+    /// data is returned.  Rows are returned as Google.Cloud.BigQuery.V2.BigQueryRow objects.
+    /// Data can be extracted by indexing by column name (ex: (string) row["title"]; ).
+    /// </para>
+    /// <example>
+    ///   <code>
+    /// PS C:\> $table = get-bqtable -DatasetID "book_data" "classics"
+    /// PS C:\> $list = $table | get-bqtabledata
+    ///   </code>
+    ///   <para>Fetches all of the rows in book_data:classics and exports them to $list</para>
```
quoctruong: nit: This sentence should end with ".".

Google.PowerShell/BigQuery/BqTabledata.cs
```diff
@@ -431,4 +431,76 @@ protected override void ProcessRecord()
...
+        public TableReference InputObject { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The number of results returned per call to the server. If there are more results
+        /// than the initial page, they will be fetched lazily with subsequent pages. 
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        public int? PageSize { get; set; }
```
quoctruong: I'm not sure if this parameter is useful to the end user though. @ILMTitan, what do you think?

Google.PowerShell/ReleaseFiles/GoogleCloudPlatform.Format.ps1xml
ahandley: Yeah, it's because it treats the rows as individual items since they're written to the pipeline as individual items.  Makes things odd for printing.

I'll add a TODO and give it the rest of the day to see if I can improve it.  It's going to matter for exporting data from jobs too, so I'd like to find a better solution if one exists.

Google.PowerShell/BigQuery/BqTabledata.cs
ahandley: got it.

Google.PowerShell/BigQuery/BqTabledata.cs
ILMTitan: I agree. It is an implementation detail we should hide from the user. We should just "do the right thing".

The only scenario I can come up with where the user would want it is if they know exactly how many rows they want to inspect, and don't want to waste data cap getting extra. I don't think that is a scenario we need to cater to.

ahandley: That wouldn't really help here either, because the veneer library doesn't support a data cap.  This just controls the size of each page as it is lazily loaded through the iteration.

ahandley: I guess it gets the axe then.  I'll send another revision in a few.

ahandley:
Alright, PTAL one more time.

ILMTitan:
LGTM

PR #507, opened 2017-03-28
Get-Bqjob + tests

ahandley:
First cmdlet for job functionality.  Does Lists and Gets.

Same general template as the other 3 sections.

### DIFF
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -14,11 +14,12 @@ Describe "Get-BqJob" {
         New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
             New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $table | 
             Add-BqTabledata $filename CSV -SkipLeadingRows 1
+        $table | Add-BqTabledata $filename CSV -SkipLeadingRows 1
     }
 
     It "should list jobs from the past 6 months" {
         $jobs = Get-BqJob
-        $jobs.Count | Should BeGreaterThan 0
+        $jobs.Count | Should BeGreaterThan 1
     }
 
     It "should get specific job via pipeline" {
@@ -44,7 +45,7 @@ Describe "Get-BqJob" {
     }
 
     It "should handle projects that the user does not have permissions for" {
-        { Get-BqJob -Project $accessErrProject -ErrorAction Stop } | Should Throw 400
+        { Get-BqJob -Project $accessErrProject -ErrorAction Stop } | Should Throw "400"
     }
 
     AfterAll {
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -23,6 +23,7 @@ Describe "Get-BqJob" {
     }
 
     #TODO(ahandley): When Start- and Stop-BqJob are written, add in tests for AllUsers and State
+    #TODO(ahandley): When Start- is ready, add test with alternate project via jobReference
 
     It "should get specific job via pipeline" {
         $jobs = Get-BqJob
@@ -31,10 +32,17 @@ Describe "Get-BqJob" {
         $return.JobReference.JobId | Should Be $job.JobReference.JobId
     }
 
-    It "should get specific job via parameter" {
+    It "should get specific job via string parameter" {
         $jobs = Get-BqJob
         $job = $jobs[0] 
-        $return = Get-BqJob -JobId $job.JobReference.JobId
+        $return = Get-BqJob $job.JobReference.JobId
+        $return.JobReference.JobId | Should Be $job.JobReference.JobId
+    }
+
+    It "should get specific job via object parameter" {
+        $jobs = Get-BqJob
+        $job = $jobs[0] 
+        $return = Get-BqJob $job
         $return.JobReference.JobId | Should Be $job.JobReference.JobId
     }
 
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -22,6 +22,8 @@ Describe "Get-BqJob" {
         $jobs.Count | Should BeGreaterThan 1
     }
 
+    #TODO(ahandley): When Start- and Stop-BqJob are written, add in tests for AllUsers and State
+
     It "should get specific job via pipeline" {
         $jobs = Get-BqJob
         $job = $jobs[0] 
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -1,0 +1,55 @@
+﻿. $PSScriptRoot\..\BigQuery\BqCmdlets.ps1
+$project, $zone, $oldActiveConfig, $configName = Set-GCloudConfig
+
+Describe "Get-BqJob" {
+
+    BeforeAll {
+        $r = Get-Random
+        $datasetName = "pshell_testing_$r"
+        $test_set = New-BqDataset $datasetName
+        $folder = Get-Location
+        $folder = $folder.ToString()
+        $filename = "$folder\classics.csv"
+        $table = New-BqTable -Dataset $test_Set "table_$r"
+        New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
+            New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $table | 
+            Add-BqTabledata $filename CSV -SkipLeadingRows 1
+    }
+
+    It "should list jobs from the past 6 months" {
+        $jobs = Get-BqJob
+        $jobs.Count | Should BeGreaterThan 0
+    }
+
+    It "should get specific job via pipeline" {
+        $jobs = Get-BqJob
+        $job = $jobs[0] 
+        $return = $job | Get-BqJob
+        $return.JobReference.JobId | Should Be $job.JobReference.JobId
+    }
+
+    It "should get specific job via parameter" {
+        $jobs = Get-BqJob
+        $job = $jobs[0] 
+        $return = Get-BqJob -JobId $job.JobReference.JobId
+        $return.JobReference.JobId | Should Be $job.JobReference.JobId
+    }
+
+    It "should throw when the job is not found"{
+        { Get-BqJob $nonExistJob -ErrorAction Stop } | Should Throw "404"
+    }
+
+    It "should throw when the project is not found"{
+        { Get-BqJob -project $nonExistProject -ErrorAction Stop } | Should Throw "404"
+    }
+
+    It "should handle projects that the user does not have permissions for" {
+        { Get-BqJob -Project $accessErrProject -ErrorAction Stop } | Should Throw 400
+    }
+
+    AfterAll {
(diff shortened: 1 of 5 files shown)


Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -0,0 +1,132 @@
...
+namespace Google.PowerShell.BigQuery
+{
+    /// <summary>
+    /// <para type="synopsis">
+    /// Lists all jobs that you started in the specified project or returns information about a specific job.
+    /// </para>
+    /// <para type="description">
+    /// If no Job is specified though ID parameter or object via pipeline, this cmdlet lists all jobs that 
+    /// you started in the specified project. If a Job is specified, it will return a descriptor object 
+    /// for that job. Listing requires Can View or Is Owner project roles. Viewing information about a
```
iantalarico: Should these be "View" and "Owner"? Can View and Is Owner seem odd

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// <summary>
+    /// <para type="synopsis">
+    /// Lists all jobs that you started in the specified project or returns information about a specific job.
+    /// </para>
+    /// <para type="description">
+    /// If no Job is specified though ID parameter or object via pipeline, this cmdlet lists all jobs that 
+    /// you started in the specified project. If a Job is specified, it will return a descriptor object 
+    /// for that job. Listing requires Can View or Is Owner project roles. Viewing information about a 
+    /// specific job requires that you’re the one who ran the job, or that you have the Is Owner project 
+    /// role. Job information is stored for a six month period after its creation.
```
iantalarico: Job information is stored for a six month period after its creation.

Maybe:

Job information is stored for six months after its creation.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -0,0 +1,132 @@
...
+    /// </para>
+    /// <example>
+    ///   <code>PS C:\> Get-BqJob</code>
+    ///   <para>This will list all past or present jobs from the default project.</para>
```
iantalarico: I think you can drop "This will" and say "Lists" in all of these

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -0,0 +1,132 @@
...
+            {
+                if (JobId == null)
+                {
+                    JobId = InputObject.JobId;
+                }
+                DoGetRequest(Service.Jobs.Get(Project, JobId));
+            }
+        }
+
+        public void DoListRequest(JobsResource.ListRequest request)
```
iantalarico: comments on functions?

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            do
+            {
+                JobList response = request.Execute();
```
iantalarico: Catch GoogleApiExceptions?

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -0,0 +1,55 @@
...
+        $folder = Get-Location
+        $folder = $folder.ToString()
+        $filename = "$folder\classics.csv"
+        $table = New-BqTable -Dataset $test_Set "table_$r"
+        New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
+            New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $table | 
+            Add-BqTabledata $filename CSV -SkipLeadingRows 1
+    }
+
+    It "should list jobs from the past 6 months" {
```
iantalarico: Is this really testing anything we have control over?

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -0,0 +1,55 @@
...
+    It "should throw when the job is not found"{
+        { Get-BqJob $nonExistJob -ErrorAction Stop } | Should Throw "404"
+    }
+
+    It "should throw when the project is not found"{
+        { Get-BqJob -project $nonExistProject -ErrorAction Stop } | Should Throw "404"
+    }
+
+    It "should handle projects that the user does not have permissions for" {
+        { Get-BqJob -Project $accessErrProject -ErrorAction Stop } | Should Throw 400
```
iantalarico: Should 400 be in quotes?  If nothing else to be consistent?

Google.PowerShell/BigQuery/BqJob.cs
ahandley: I took those permissions directly off of the BQ Jobs API page.  I do see that they're Legacy permissions and that the IAM roles are the way to go.  Should I keep these roles to be consistent with the docs or should I put the new IAM roles?  

(I sent feedback on the page mentioning this as well.  Is that read at all, or should I file a bug on the page for being out of date?)

Google.PowerShell/BigQuery/BqJob.cs
ahandley: I figured the names would be self-documenting, but I threw on some additional details.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -0,0 +1,132 @@
...
+using System.Management.Automation;
+
+namespace Google.PowerShell.BigQuery
+{
+    /// <summary>
+    /// <para type="synopsis">
+    /// Lists all jobs that you started in the specified project or returns information about a specific job.
+    /// </para>
+    /// <para type="description">
+    /// If no Job is specified though ID parameter or object via pipeline, this cmdlet lists all jobs that
```
ahandley: Fixed that paragraph entirely.  A lot of it was language directly from the API page, so it was stitched together somewhat awkwardly.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: got it.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
ahandley: PS auto-boxes/types, but yes.  For consistency, it should.

ahandley:
Fixed a bunch of stuff.  PTAL.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -14,11 +14,12 @@ Describe "Get-BqJob" {
         New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
             New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $table | 
             Add-BqTabledata $filename CSV -SkipLeadingRows 1
+        $table | Add-BqTabledata $filename CSV -SkipLeadingRows 1
     }
 
     It "should list jobs from the past 6 months" {
         $jobs = Get-BqJob
-        $jobs.Count | Should BeGreaterThan 0
+        $jobs.Count | Should BeGreaterThan 1
```
iantalarico: This just tests that a job exists.  Is this something we need a test for?  We test the get calls below

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -15,19 +16,18 @@ namespace Google.PowerShell.BigQuery
     /// Lists all jobs that you started in the specified project or returns information about a specific job.
     /// </para>
     /// <para type="description">
-    /// If no Job is specified though ID parameter or object via pipeline, this cmdlet lists all jobs that 
-    /// you started in the specified project. If a Job is specified, it will return a descriptor object 
-    /// for that job. Listing requires Can View or Is Owner project roles. Viewing information about a 
-    /// specific job requires that you’re the one who ran the job, or that you have the Is Owner project 
-    /// role. Job information is stored for a six month period after its creation.
+    /// If no Job is specified through the JobId parameter or object via pipeline, a list of all jobs in the
```
iantalarico: Very nice :)

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
ahandley: Yes, because that tests the "list" functionality of the cmdlet.  It checks that we can access job history (and presently running jobs if there are any), whereas the other two test getting a specific job.

iantalarico: SGTM

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -0,0 +1,148 @@
...
+            Property = nameof(JobList.JobsData.JobReference))]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(Apis.Bigquery.v2.Data.Job),
+            Property = nameof(Apis.Bigquery.v2.Data.Job.JobReference))]
+        public JobReference InputObject { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            if (JobId == null && InputObject == null)
+            {
+                DoListRequest(Service.Jobs.List(Project));
```
ILMTitan: It is a little odd that your are passing a request here. The standard structure for our Cmdlets has been to build the request inside the specific function, and then have that function return either the single resource or an IEnumerable of resources, and do the WriteObject in the ProcessRecord. This separates the Google Api stuff (the request, response etc) and the PowerShell Cmdlet stuff (ProcessRecord override, WriteObject).

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -0,0 +1,148 @@
...
+                }
+                DoGetRequest(Service.Jobs.Get(Project, JobId));
+            }
+        }
+
+        /// <summary>
+        /// Executes a List Jobs request and writes returned objects or errors.
+        /// </summary>
+        /// <param name="request">Pre-built request ready to be executed</param>
+        public void DoListRequest(JobsResource.ListRequest request)
```
ILMTitan: There appears to be a stateFilter in the Jobs.List API. As it is optional, it is probably a property rather than a parameter. There should be a Cmdlet parameter that sets it.
There is also an allUsers boolean. There should be a switch parameter for that as well.

ahandley: Added both of these.  Tests will come later when I have Start- and Stop- usable.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: I refactored one, and sort of refactored the other.  I couldn't find a more elegant way of doing it other than bringing both of the catch blocks into ProcessRecord, which severely clutters it.  So I'm not passing in a request anymore, but the helper function also doesn't return anything.  I think it's a decent compromise, while keeping the code readable.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -0,0 +1,58 @@
...
+        $return.JobReference.JobId | Should Be $job.JobReference.JobId
+    }
+
+    It "should get specific job via parameter" {
+        $jobs = Get-BqJob
+        $job = $jobs[0] 
+        $return = Get-BqJob -JobId $job.JobReference.JobId
+        $return.JobReference.JobId | Should Be $job.JobReference.JobId
+    }
+
```
ILMTitan: Add a test like:

$jobs = Get-BqJob -Project $someOtherProject
$job = $jobs[0]
$return = $job | Get-BqJob
$return.JobReference.JobId | Should Be $job.JobReference.JobId
$return.JobReference.ProjectId | Should Be $job.JobReference.ProjectId

ahandley:
Fixed the project issue.  Can't write a test for it today, but I'll put it in tomorrow with the Start- tests.

Also implemented the parameter sets.  Thanks for the hint on that.

PR #509, opened 2017-03-30
The Big Query

ahandley:
Query functionality + tests

### DIFF
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -87,7 +87,7 @@ Describe "BqJob-Query" {
     It "should query out of a pre-loaded table" {
         $alt_tab = New-BqTable -Dataset $test_Set "table_res_$r"
         $job = Start-BqJob -Query "select * from $datasetName.table_$r where Year > 1900" `
-            -DefaultDataset $test_set -DestinationTable $alt_tab -PollUntilComplete
+                           -DefaultDataset $test_set -DestinationTable $alt_tab -PollUntilComplete
         $job = $job | Get-BqJob
         $alt_tab = $alt_tab | Get-BqTable
         $job.Status.State | Should Be "DONE"
@@ -99,6 +99,11 @@ Describe "BqJob-Query" {
     It "should use legacy SQL when asked" {
         $job = Start-BqJob -Query "select * from $datasetName.table_$r where Year > 1900" -UseLegacySql
         $job.Configuration.Query.UseLegacySql | Should Be True
+        #TODO(ahandley): Add in checking the data once receive is done.
+    }
+
+    It "should properly halt when -WhatIf is passed" {
+        Start-BqJob -Query "select * from $datasetName.table_$r where Year > 1900" -WhatIf | Should Be $null
     }
 
     # Note - Priority cannot be tested, as the server does not output the Job.Configuration.Query.Priority property.
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -63,4 +63,53 @@ Describe "Get-BqJob" {
     }
 }
 
+Describe "BqJob-Query" {
+
+    BeforeAll {
+        $r = Get-Random
+        $datasetName = "pshell_testing_$r"
+        $test_set = New-BqDataset $datasetName
+        $folder = Get-Location
+        $folder = $folder.ToString()
+        $filename = "$folder\classics.csv"
+        $table = New-BqTable -Dataset $test_Set "table_$r"
+        New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
+            New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $table | 
+            Add-BqTabledata $filename CSV -SkipLeadingRows 1
+    }
+
+    It "should query out of a pre-loaded table" {
+        $job = Start-BqJob -Query "select * from $datasetName.table_$r where Year > 1900"
+        $job | Should Not Be $null
+        #TODO(ahandley): Add in checking the data once receive is done.
+    }
+
+    It "should query out of a pre-loaded table" {
+        $alt_tab = New-BqTable -Dataset $test_Set "table_res_$r"
+        $job = Start-BqJob -Query "select * from $datasetName.table_$r where Year > 1900" `
+            -DefaultDataset $test_set -DestinationTable $alt_tab -PollUntilComplete
+        $job = $job | Get-BqJob
+        $alt_tab = $alt_tab | Get-BqTable
+        $job.Status.State | Should Be "DONE"
+        $job.Configuration.Query.DefaultDataset.DatasetId | Should Be $test_set.DatasetReference.DatasetId
+        $job.Configuration.Query.DestinationTable.TableId | Should Be $alt_tab.TableReference.TableId
+        $alt_tab.NumRows | Should Be 2
+    }
+
+    It "should use legacy SQL when asked" {
+        $job = Start-BqJob -Query "select * from $datasetName.table_$r where Year > 1900" -UseLegacySql
+        $job.Configuration.Query.UseLegacySql | Should Be True
+    }
+
+    # Note - Priority cannot be tested, as the server does not output the Job.Configuration.Query.Priority property.
+
+    It "should handle projects that the user does not have permissions for" {
+        { Start-BqJob -Query "select * from $datasetName.table_$r" -Project $accessErrProject } | Should Throw "400"
+    }
+
+    AfterAll {
+        $test_set | Remove-BqDataset -Force
+    }
+}
+
 Reset-GCloudConfig $oldActiveConfig $configName
--- a/Google.PowerShell/BigQuery/BqCmdlet.cs
+++ b/Google.PowerShell/BigQuery/BqCmdlet.cs
@@ -21,7 +21,6 @@ namespace Google.PowerShell.BigQuery
 
         public BqCmdlet()
         {
-            //Service = new BigqueryService(GetBaseClientServiceInitializer());
             _service = new Lazy<BigqueryService>(() => new BigqueryService(GetBaseClientServiceInitializer()));
             _client = new Lazy<BigQueryClient>(() => BigQueryClient.Create(Project));
         }
--- a/Google.PowerShell/BigQuery/BqCmdlet.cs
+++ b/Google.PowerShell/BigQuery/BqCmdlet.cs
@@ -2,7 +2,9 @@
 // Licensed under the Apache License Version 2.0.
 
 using Google.Apis.Bigquery.v2;
+using Google.Cloud.BigQuery.V2;
 using Google.PowerShell.Common;
+using System;
 
 namespace Google.PowerShell.BigQuery
 {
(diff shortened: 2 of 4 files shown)


Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -63,4 +63,53 @@ Describe "Get-BqJob" {
...
+        $alt_tab = $alt_tab | Get-BqTable
+        $job.Status.State | Should Be "DONE"
+        $job.Configuration.Query.DefaultDataset.DatasetId | Should Be $test_set.DatasetReference.DatasetId
+        $job.Configuration.Query.DestinationTable.TableId | Should Be $alt_tab.TableReference.TableId
+        $alt_tab.NumRows | Should Be 2
+    }
+
+    It "should use legacy SQL when asked" {
+        $job = Start-BqJob -Query "select * from $datasetName.table_$r where Year > 1900" -UseLegacySql
+        $job.Configuration.Query.UseLegacySql | Should Be True
```
quoctruong: We should check the result of the job too.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -178,5 +178,257 @@ public void DoGetRequest()
...
+    /// <example>
+    ///   <code>
+    /// PS C:\> $job = Start-BqJob -Query "select * from book_data.classics where Year > 1900"
+    ///   </code>
+    ///   <para>Queries the classics table and returns a Job object so that results can be viewed.</para>
+    /// </example>
+    /// <example>
+    ///   <code>
+    /// PS C:\> $job = Start-BqJob -Query "select * from classics where Year > 1900" `
+    /// -DefaultDataset $dataset -DestinationTable $table
```
quoctruong: nit: -DefaultDataset should be at the same indentation level as -Query.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -178,5 +178,257 @@ public void DoGetRequest()
...
+            public const string DoCopy = "DoCopy";
+            public const string DoLoad = "DoLoad";
+            public const string DoExtract = "DoExtract";
+        }
+
+        // All-Type Parameters.
+
+        /// <summary>
+        /// <para type="description">
+        /// The project to look for jobs in. If not set via PowerShell parameter processing, it will
```
quoctruong: The comment seems to be for Get-BqJob.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -178,5 +178,257 @@ public void DoGetRequest()
...
+        public override string Project { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// Turns the async call into a synchronous call by polling until the job is complete before 
+        /// returning. Can also be accessed by '-Synch'.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        [Alias("Synch")]
```
quoctruong: I think the alias should be "Synchronous" instead of just "Synch". That way, user can still use -Synch if they want.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -178,5 +178,257 @@ public void DoGetRequest()
...
+
+        /// <summary>
+        /// <para type="description">
+        /// Causes BigQuery to not run the job. Instead, if the query is valid, BigQuery returns 
+        /// statistics about the job such as how many bytes would be processed. If the query is 
+        /// invalid, an error returns. //This feature is not currently available.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.DoQuery)]
+        public SwitchParameter DryRun { get; set; }
```
quoctruong: We should probably make use of the built-in -WhatIf parameter of PowerShell (when you do SupportsShouldProcess) and remove this parameter.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -178,5 +178,257 @@ public void DoGetRequest()
...
+        /// <summary>
+        /// <para type="description">
+        /// A dataset to use for any unqualified table names in the query.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.DoQuery)]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(Dataset),
+            Property = nameof(Dataset.DatasetReference))]
+        [Alias("Dataset")]
+        public DatasetReference DefaultDataset { get; set; }
```
quoctruong: I think this parameter should just be Dataset. From the user's perspective, the Default prefix does not add any value and may be confusing.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
ahandley: Should have had a TODO in for that one too.  I need Receive- for that, and I'll add it in asap.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: I thought it worked, but I re-worded it to be more exact.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: What does -Synch do?  I can't find any docs on it.  
I went ahead and updated it anyway, though.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: That's probably fair.  The counter argument would be that people might think all referenced tables would be looked for in that dataset, so leaving the "default" in there would help.  
Is it sufficient to explain how the default works in the parameter docs and leave it at that?

Google.PowerShell/BigQuery/BqJob.cs
ahandley: That'll work for now.  Lemme patch that in.

Google.PowerShell/BigQuery/BqJob.cs
quoctruong: I mean in PowerShell if your parameter is -FooName and on the command line, you just type -Foo then PowerShell can still resolve it to -FooName.

ahandley: Oh, gotcha.  That works.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: @quoctruong - bump?

ahandley:
Addressed all the above issues (with the exception of the Dataset thing.  Discussion open on that).  Added Shouldprocess to the query subset, and I'll continue to add that for each other section as they get coded.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -178,5 +178,257 @@ public void DoGetRequest()
...
+        /// <para type="description">
+        /// Selects job type Extract.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, ParameterSetName = ParameterSetNames.DoExtract)]
+        public SwitchParameter Extract { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            var Client = BigQueryClient.Create(Project);
```
ILMTitan: In our Apiary based cmdlets put the Client in an Inherited property. I suggest you do the same here (possibly making a separate abstract subclass).

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -178,5 +178,257 @@ public void DoGetRequest()
...
+                }
+
+                BigQueryJob bqr = client.CreateQueryJob(QueryString, options);
+
+                if (PollUntilComplete)
+                {
+                    bqr.PollUntilCompleted();
+                }
+
+                WriteObject(bqr.Resource);
```
ILMTitan: These functions should return the Resource to ProcessRecord. The WriteObject call should be done there.

ILMTitan: These functions should return the Resource to `ProcessRecord`. The `WriteObject` call should be done there.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -178,5 +178,257 @@ public void DoGetRequest()
...
+                    break;
+                case ParameterSetNames.DoExtract:
+                    DoExtract(Client);
+                    break;
+                default:
+                    ThrowTerminatingError(new ErrorRecord(
+                        new Exception("You must select a valid BQ Job type."),
+                        "Type Not Found", ErrorCategory.ObjectNotFound, this));
+                    break;
+            }
```
ILMTitan: You can set a single variable from all of the functions above, and do a single `WriteObject` call here (if they return the same type of resource).

ahandley: Yep.  It's all a Apis.Bigquery.v2.Data.Job.  Simple enough.

Google.PowerShell/BigQuery/BqJob.cs

Google.PowerShell/BigQuery/BqJob.cs
ILMTitan: I am very split on this, but I think I am siding with @ahandley. I think this parameter is different enough from other Dataset parameters that giving it a special name is warranted.

quoctruong: I think DefaultDataset is okay as long as we document it well.

ahandley: I added in a few clarifying words to the parameter description, and added a section in the cmdlet docs.  Will push that in the next commit.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: So I looked into putting it in the pre-built BqCmdlet class with the Apis.BQ Service:

```
        public BigqueryService Service { get; private set; }
        public BigQueryClient Client { get; private set; }

        public BqCmdlet()
        {
            Service = new BigqueryService(GetBaseClientServiceInitializer());
            Client = BigQueryClient.Create();  //takes a ProjectId as an argument though.
        }
```

The issue with this is that you need to initialize the BigQueryClient with a project ID.  This is not known, in most cases, until parameter sets are parsed, which is after the super calls happen.  

If I'm reading your comment the wrong way, let me know, but I think the call is currently best placed where it is.

ahandley: So I tried playing around with the Lazy<>, but it still had the chicken-or-the-egg problem.  I needed to put the lazy declaration on the property, but "Project" wasn't available until the constructor.  

It's definitely messier this way, but I couldn't find a clean solution (or any at all) after messing around with the Lazy's

ILMTitan: You don't have to set the readonly Lazy field in an initialize. You can set the readonly field in the constructor. From there you can reference (but not resolve) Project. Then, when Client is first resolved, Project will have a value, and everything should work.

ahandley:
Alright.  Try that.  Much better.

Google.PowerShell/BigQuery/BqCmdlet.cs
```diff
@@ -11,11 +13,17 @@ namespace Google.PowerShell.BigQuery
...
+        private readonly Lazy<BigqueryService> _service;
+        private readonly Lazy<BigQueryClient> _client;
+
+        public BigqueryService Service => _service.Value;
+        public BigQueryClient Client => _client.Value;
 
         public BqCmdlet()
         {
-            Service = new BigqueryService(GetBaseClientServiceInitializer());
+            //Service = new BigqueryService(GetBaseClientServiceInitializer());
```
ILMTitan: Remove commented code.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -178,5 +178,251 @@ public void DoGetRequest()
...
+        public SwitchParameter Extract { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            Apis.Bigquery.v2.Data.Job result;
+
+            switch (ParameterSetName)
+            {
+                case ParameterSetNames.DoQuery:
+                    result = DoQuery(Client);
```
ILMTitan: Part of the point of doing the Client property was you didn't have to pass this parameter anymore.

Google.PowerShell/BigQuery/BqCmdlet.cs
```diff
@@ -11,11 +13,17 @@ namespace Google.PowerShell.BigQuery
...
+            _service = new Lazy<BigqueryService>(() => new BigqueryService(GetBaseClientServiceInitializer()));
```
quoctruong: Not sure if you need a Lazy for the service since it's almost always used by the cmdlet.

ahandley: I figured that I would use it because of consistency and because a few cmdlets don't use it, so might as well save time.

ILMTitan: The ones that use Client don't use Service. I think it is fine.

ahandley:
Thanks!

PR #510, opened 2017-04-03
Receive-BqJob + tests

ahandley:
Pretty short commit.  Added to the Query tests instead of making a new section.

### DIFF
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -81,7 +81,10 @@ Describe "BqJob-Query" {
     It "should query out of a pre-loaded table" {
         $job = Start-BqJob -Query "select * from $datasetName.table_$r where Year > 1900"
         $job | Should Not Be $null
-        #TODO(ahandley): Add in checking the data once receive is done.
+        $results = $job | Receive-BqJob
+        $results.Count | Should Be 2
+        $results[0]["Author"] | Should Be "Orson Scott Card"
+        $results[1]["Year"] | Should Be 1967
     }
 
     It "should query out of a pre-loaded table" {
@@ -99,7 +102,19 @@ Describe "BqJob-Query" {
     It "should use legacy SQL when asked" {
         $job = Start-BqJob -Query "select * from $datasetName.table_$r where Year > 1900" -UseLegacySql
         $job.Configuration.Query.UseLegacySql | Should Be True
-        #TODO(ahandley): Add in checking the data once receive is done.
+        $results = Receive-BqJob $job
+        $results.Count | Should Be 2
+        $results[0]["Year"] | Should Be 1967
+        $results[1]["Author"] | Should Be "Orson Scott Card"
+    }
+
+    It "should handle Timeouts" {
+        $job = Start-BqJob -Query "select * from $datasetName.table_$r where Year > 1900" -UseLegacySql
+        $job.Configuration.Query.UseLegacySql | Should Be True
+        $results = Receive-BqJob $job -Timeout 1000
+        $results.Count | Should Be 2
+        $results[0]["Year"] | Should Be 1967
+        $results[1]["Author"] | Should Be "Orson Scott Card"
     }
 
     It "should properly halt when -WhatIf is passed" {
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -87,7 +87,7 @@ Describe "BqJob-Query" {
         $results[1]["Year"] | Should Be 1967
     }
 
-    It "should query out of a pre-loaded table" {
+    It "should query out of a pre-loaded table with more options than ever before!" {
         $alt_tab = New-BqTable -Dataset $test_Set "table_res_$r"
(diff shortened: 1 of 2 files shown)


Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -421,4 +423,71 @@ public Apis.Bigquery.v2.Data.Job DoExtract()
...
+        /// <para type="description">
+        /// Max time, in seconds, to wait for the job to complete before failing (Default: 10).
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        public int Timeout { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            Project = InputObject.ProjectId;
```
ILMTitan: Add a comment as to why you are doing this. It is a bit arcane for those not in the know.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -421,4 +423,71 @@ public Apis.Bigquery.v2.Data.Job DoExtract()
...
+            do
+            {
+                options.StartIndex = rowsRead;
+                result = Client.GetQueryResults(InputObject, options);
+                foreach (BigQueryRow row in result.GetRows())
+                {
+                    WriteObject(row);
+                    rowsRead++;
+                }
+            } while (rowsRead < result.TotalRows);
```
ILMTitan: Do you need this outer loop? It looks to me like `result.GetRows()` will get them all.

> Returns a lazy enumeration of all the rows in the query (from any originally-specified page token onwards), making server calls as required. If the job has not completed already, it is polled until it has completed using pollSettings.

If you don't need the outerloop, you can just do `WriteObject(result.GetRows(), true)`.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: Got it!

Google.PowerShell/BigQuery/BqJob.cs
ahandley: This is correct...  I must have been looking at the wrong docs, because I couldn't find that earlier.  (hence, rolling my own).  Thanks!

ahandley:
PTAL now.  Both issues should be fixed.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -421,4 +423,73 @@ public Apis.Bigquery.v2.Data.Job DoExtract()
...
+        public int Timeout { get; set; }
+
+        protected override void ProcessRecord()
+        {
+            //Set Project for the lazy instantiation of a BQ Client object
+            Project = InputObject.ProjectId;
+
+            var options = new GetQueryResultsOptions();
+            options.Timeout = new TimeSpan(0, 0, (Timeout < 10) ? 10 : Timeout);
+            BigQueryResults result;
```
ILMTitan: You don't need to separate the declaration and the assignment anymore. Put them together.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -421,4 +423,73 @@ public Apis.Bigquery.v2.Data.Job DoExtract()
...
+        [ValidateNotNull]
+        public JobReference InputObject { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// Max time, in seconds, to wait for the job to complete before failing (Default: 10).
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false)]
+        public int Timeout { get; set; }
```
ILMTitan: Consider making this a TimeSpan object. PowerShell can convert a string like "0:0:20" to a timespan of 20 seconds.

ahandley: I did, but I figured it was better to stick with the convention of using raw seconds for time limits.  It's what I did in a few of the other cmdlets as well.

Might be less powerful, but it'll be easier for users this way.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -81,7 +81,10 @@ Describe "BqJob-Query" {
     It "should query out of a pre-loaded table" {
```
quoctruong: This is not related to this CR but it seems we don't have a test for -Synchronous switch of Start-BqJob? Maybe you should add one.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -421,4 +423,73 @@ public Apis.Bigquery.v2.Data.Job DoExtract()
...
+            Project = InputObject.ProjectId;
+
+            var options = new GetQueryResultsOptions();
+            options.Timeout = new TimeSpan(0, 0, (Timeout < 10) ? 10 : Timeout);
+            BigQueryResults result;
+
+            try
+            {
+                result = Client.GetQueryResults(InputObject, options);
+                WriteObject(result.GetRows(), true);
```
quoctruong: Should we do null check for result?

ahandley: Yep.  I'll have that throw down to the catch that I just put in.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
ahandley: That's not a bad idea.  I'll use the line of code that I showed in the demo today.

ahandley:
Got em!  Thanks.  PTAL.

ahandley:
Thanks yall!

PR #511, opened 2017-04-03
Stop- + tests (last new cmdlet!)

ahandley:
Last new cmdlet.  

Tests below.  It's kind of weak at the moment, but I will add more cases to make sure it works as I can do more types of job requests.

### DIFF
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -135,4 +135,33 @@ Describe "BqJob-Query" {
     }
 }
 
+Describe "Stop-BqJob" {
+
+    BeforeAll {
+        $r = Get-Random
+        $datasetName = "pshell_testing_$r"
+        $test_set = New-BqDataset $datasetName
+        $folder = Get-Location
+        $folder = $folder.ToString()
+        $filename = "$folder\classics_large.csv"
+        $table = New-BqTable -Dataset $test_Set "table_$r"
+        New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
+            New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $table | 
+            Add-BqTabledata $filename CSV -SkipLeadingRows 1
+    }
+
+    It "should query out of a pre-loaded table" {
+        $job = Start-BqJob -Query "select * from book_data.classics where Year > 1900"
+        $res = $job | Stop-Bqjob
+        $can = $res | Get-BqJob
+        $can.Status.State | Should Be "DONE"
+    }
+
+    #TODO(ahandley): Add a few more test cases with different types of jobs that may run longer than a single query
+
+    AfterAll {
+        $test_set | Remove-BqDataset -Force
+    }
+}
+
 Reset-GCloudConfig $oldActiveConfig $configName
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -152,9 +152,16 @@ Describe "Stop-BqJob" {
 
     It "should stop a query job" {
         $job = Start-BqJob -Query "select * from book_data.classics where Year > 1900"
-        $res = $job | Stop-Bqjob
-        $can = $res | Get-BqJob
-        $can.Status.State | Should Be "DONE"
+        $res = $job | Stop-Bqjob| Get-BqJob
+        $res.Status.State | Should Be "DONE"
+    }
+
+    It "should handle jobs that are already done" {
+        $job = Start-BqJob -Query "select * from book_data.classics" -Synchronous
+        $job = $job | Get-BqJob
+        $job.Status.State | Should Be "DONE"
+        $res = $job | Stop-Bqjob | Get-BqJob
+        $res.Status.State | Should Be "DONE"
     }
 
     #TODO(ahandley): Add a few more test cases with different types of jobs that may run longer than a single query
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -78,7 +78,7 @@ Describe "BqJob-Query" {
             Add-BqTabledata $filename CSV -SkipLeadingRows 1
     }
 
-    It "should query out of a pre-loaded table" {
+    It "should query a pre-loaded table" {
         $job = Start-BqJob -Query "select * from $datasetName.table_$r where Year > 1900"
         $job | Should Not Be $null
         $results = $job | Receive-BqJob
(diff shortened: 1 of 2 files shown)


Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -496,5 +496,67 @@ protected override void ProcessRecord()
         }
     }
 
-    
+    /// <summary>
+    /// <para type="synopsis">
+    /// Requests that a running BQ Job be canceled.
```
iantalarico: Optional: I would spell out BQ

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -496,5 +496,67 @@ protected override void ProcessRecord()
...
+    /// </para>
+    /// <para type="description">
+    /// Requests that a job be canceled. This call will return immediately, and the client
```
iantalarico: I would split this up a bit it's a little long winded.

Maybe:
Requests that a job be canceled. This call will return immediately and the client 
is responsible for polling for job status. Canceled jobs may still incur costs. This cmdlet returns a Job object if successful.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -496,5 +496,67 @@ protected override void ProcessRecord()
...
 
-    
+    /// <summary>
+    /// <para type="synopsis">
+    /// Requests that a running BQ Job be canceled.
+    /// </para>
+    /// <para type="description">
+    /// Requests that a job be canceled. This call will return immediately, and the client 
+    /// will need to poll for job status to see if the cancel completed successfully. 
+    /// Canceled jobs may still incur costs. This cmdlet returns a Job object if successful.
```
iantalarico: What happens if it is not successful does it throw?

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -496,5 +496,67 @@ protected override void ProcessRecord()
...
+    /// </para>
+    /// <example>
+    ///   <code>
+    /// PS C:\> $job = Start-BqJob -Query "select * from book_data.classics"
```
iantalarico: Nit: SELECT * FROM book_data.classics;

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -135,4 +135,33 @@ Describe "BqJob-Query" {
...
+        $folder = Get-Location
+        $folder = $folder.ToString()
+        $filename = "$folder\classics_large.csv"
+        $table = New-BqTable -Dataset $test_Set "table_$r"
+        New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
+            New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $table | 
+            Add-BqTabledata $filename CSV -SkipLeadingRows 1
+    }
+
+    It "should query out of a pre-loaded table" {
```
iantalarico: "should query out of a pre-loaded table"? maybe drop "out of"?

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -135,4 +135,33 @@ Describe "BqJob-Query" {
...
+        $job = Start-BqJob -Query "select * from book_data.classics where Year > 1900"
+        $res = $job | Stop-Bqjob
+        $can = $res | Get-BqJob
+        $can.Status.State | Should Be "DONE"
```
iantalarico: given the description below should you poll here?

ahandley: Probably.  I'm generally unsatisfied with this test anyway because the query is probably going to finish long before the stop request even gets there, so I'm probably not even able to test if stop works.  I think I'm just going to have to leave this test alone until I have load/export jobs, and then maybe I can find something that takes 20-30 seconds to complete there.  Then i'll be able to kill it and poll for a response.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: Yep.  If the server responds with null or if any other exception is thrown by the api, the error will bubble up to the terminal.

Is this something I should explicitly mention?  I figured it was something that would be obvious to the user, but if not, I'll just add that to my list of fixes for all cmdlets.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
ahandley: Just realized that name was copy-pasted.  Fixed the instances that I copied from, and renamed this one.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -135,4 +135,33 @@ Describe "BqJob-Query" {
...
+        $folder = $folder.ToString()
+        $filename = "$folder\classics_large.csv"
+        $table = New-BqTable -Dataset $test_Set "table_$r"
+        New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
+            New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $table | 
+            Add-BqTabledata $filename CSV -SkipLeadingRows 1
+    }
+
+    It "should stop a query job" {
+        $job = Start-BqJob -Query "select * from book_data.classics where Year > 1900"
```
quoctruong: What happened if you call Stop-BqJob on a job that is already stopped. Can we add a test case for that?

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -496,5 +496,67 @@ protected override void ProcessRecord()
...
+    /// <para type="description">
+    /// Requests that a job be canceled. This call will return immediately, and the client 
+    /// is responible for polling for job status. Canceled jobs may still incur costs. 
+    /// This cmdlet returns a Job object if successful.
+    /// </para>
+    /// <example>
+    ///   <code>
+    /// PS C:\> $job = Start-BqJob -Query "SELECT * FROM book_data.classics"
+    /// PS C:\> $job = $job | Stop-BqJob | Get-BqJob
+    /// PS C:\> $job.Status.State
```
quoctruong: I don't think you need to add this line. Just add in the comment that the $job.Status.State should be "Done".

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
ahandley: Got it!

Google.PowerShell/BigQuery/BqJob.cs
ahandley: Fair enough.  Clarified it.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -152,9 +152,16 @@ Describe "Stop-BqJob" {
 
     It "should stop a query job" {
         $job = Start-BqJob -Query "select * from book_data.classics where Year > 1900"
-        $res = $job | Stop-Bqjob
-        $can = $res | Get-BqJob
-        $can.Status.State | Should Be "DONE"
+        $res = $job | Stop-Bqjob| Get-BqJob
```
quoctruong: nit: Space between Stop-BqJob and |.

ahandley: ill have it fixed in the next changeset.  thanks.

PR #512, opened 2017-04-05
Copy jobs + tests

ahandley:
Added the -Copy flag to Start-BqJob, and fleshed out some of the infrastructure for the other two flags.

Tests are included, as well as a new test .csv file.  

Minor change was made to Get-BqTable to correct for past oversight and to allow me to test things in an elegant fashion.


Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -313,6 +321,41 @@ private class ParameterSetNames
...
+        /// The destination table to write to.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, Position = 0, ParameterSetName = ParameterSetNames.DoCopy)]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(Table),
+            Property = nameof(Table.TableReference))]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(TableList.TablesData),
+            Property = nameof(TableList.TablesData.TableReference))]
+        [ValidateNotNull]
+        public TableReference Destination { get; set; }
```
ILMTitan: This parameter and `DestinationTable` can be one parameter. Just add a second `[Parameter()]` attribute for the new parameter set.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -396,16 +441,59 @@ public Apis.Bigquery.v2.Data.Job DoQuery()
...
+                    var copyjob = new Apis.Bigquery.v2.Data.Job()
+                    {
+                        Configuration = new JobConfiguration()
+                        {
+                            Copy = new JobConfigurationTableCopy()
+                            {
+                                DestinationTable = Destination,
+                                SourceTable = Source,
+                                WriteDisposition = (WriteMode != null) ? WriteMode.ToString() 
+                                    : WriteDisposition.WriteIfEmpty.ToString()
```
ILMTitan: You can simplify this part a bit by initializing `WriteDisposition WriteMode {get; set;} = WriteDisposition.WriteIfEmpty`.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -396,16 +441,59 @@ public Apis.Bigquery.v2.Data.Job DoQuery()
...
+                            }
+                        }
+                    };
+                    var request = new JobsResource.InsertRequest(Service, copyjob, Project);
```
ILMTitan: You can make this call with Service.Jobs.Insert(copyjob, Project).

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -415,12 +503,39 @@ public Apis.Bigquery.v2.Data.Job DoLoad()
...
         {
             throw new NotImplementedException(
                 "Extract jobs are not implemented yet.  Use Get-BqTabledata instead.");
         }
+
+        /// <summary>
+        /// This function waits for a job to reach the "DONE" status and then returns.
+        /// </summary>
+        /// <param name="job">Job to poll for completion.</param>
+        public Apis.Bigquery.v2.Data.Job PollForCompletion(Apis.Bigquery.v2.Data.Job job)
```
ILMTitan: Instead of reinventing the wheel, maybe you want to just do Client.GetJob(job.JobReference).PollUntilCompleted().

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -396,16 +441,59 @@ public Apis.Bigquery.v2.Data.Job DoQuery()
...
+                            }
+                        }
+                    };
+                    var request = new JobsResource.InsertRequest(Service, copyjob, Project);
+                    var response = request.Execute();
+
+                    if (PollUntilComplete)
+                    {
+                        response = PollForCompletion(response);
+                    }
```
ILMTitan: You are going to do this for every operation. You could instead do it once in `ProcessRecord()` (assuming the response result stays in `ProcessRecord()`.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -396,16 +441,59 @@ public Apis.Bigquery.v2.Data.Job DoQuery()
...
+                try
+                {
+                    var copyjob = new Apis.Bigquery.v2.Data.Job()
+                    {
+                        Configuration = new JobConfiguration()
+                        {
+                            Copy = new JobConfigurationTableCopy()
+                            {
+                                DestinationTable = Destination,
+                                SourceTable = Source,
```
ILMTitan: There is a `SourceTables` parameter. This may be a case where you want to collect all of the objects in the pipeline, and send them out as one request.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -313,6 +321,41 @@ private class ParameterSetNames
...
+        /// The source table to copy from.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, ValueFromPipeline = true, ParameterSetName = ParameterSetNames.DoCopy)]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(Table),
+            Property = nameof(Table.TableReference))]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(TableList.TablesData),
+            Property = nameof(TableList.TablesData.TableReference))]
+        [ValidateNotNull]
+        public TableReference Source { get; set; }
```
ILMTitan: You may want to make this take an array of TableReference. If you do, look at the ArrayPropertyTransformAttribute as well.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: Had thought about that, but didnt pull the trigger.  Just added that.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: Much more elegant.  That works.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: So for some reason, this code is giving me an issue:

**var request = new BqCmdlet.Service.Jobs.Insert(copyjob, Project);**

"The type 'Service' does not exist in the type BqCmdlet"  (which it does)

Taking the "BqCmdlet." off doesn't help either.  Still not found.  Yet the code thats already there works.  I have no idea why, but I'm going to refrain from refactoring here and come back to it when i'm done with the other issues.

ILMTitan: Isn't that just supposed to be `var request = Service.Jobs.Insert(copyjob, Project);`?

ahandley: Aaaaahhh...   missed the "new".

And that error was particularly unhelpful.  Or maybe I just need a nap.  Thanks.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: That works.  Plus more Cloud.Bigquery code is more consistent & cleaner code.

ahandley: So in the process of transplanting all of the polling code and upgrading it, I got a weird error.  My code is currently:

**
// Check if the user is requesting a synchronous operation.
            if (PollUntilComplete)
            {
                //new code
                BigQueryJob bqj = Client.GetJob(result.JobReference);
                bqj.PollQueryUntilCompleted();
                result = bqj.Resource;
                //old code
                //result = PollForCompletion(result);
                //result = Client.GetJob(result.JobReference).Resource;
            }
**

When the old code is active, the tests all run fine.  however, when I swap to the new code (which is functionally the same), I get the error:
"Message[Job ~:job_~ is not a query job]"

I think this may have something to do with the fact that the veneer can't handle Copy, Load, and Extract jobs at the moment, so I'm going to leave the old code as what gets run, and the new code commented out.  This way, I can re-test when I have Load and Extract written to see if I can learn any more about this.  Also the code that I wrote works, so if I can't get the veneer polling to work, it's still functional.

Google.PowerShell/BigQuery/BqJob.cs

Google.PowerShell/BigQuery/BqJob.cs
ahandley: I looked into doing that earlier with the ArrayPropertyTransformAttribute, and ran into a whole host of issues, both for code and for usability.  

I don't want to do code using process to aggregate and end to run the operation because then I would have to revamp get to bypass table objects on the pipeline.  This is directly conflicting with one of the frequent uses cases, which is passing a table via pipeline to get an updated table object:
```
$table = $table | Get-BqTable
```

So people would have to manually make an array of tables to pass in, or they would have to do a list request with the label-filter options that I'm going to write after this.  The array option is doable, but messy, and the label option is going to be messy for a similar reason.  they're both going to require a lot of user input and finagling with data types.  

The proposed solution, which I can make note of in the docs, is going to be to use the WriteDisposition -WriteAppend, and just do multiple copy jobs sequentially.  

So instead of saying "copy a,b,c,d to z", it'll be just "copy a to z, copy append b, copy append c, copy append d".  This also works better with the pipeline, because each command takes one argument off the pipeline and copies it at a time.  That's the best compromise that I came up with earlier.  Let me know if you have other ideas for this.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: See below, line 461 comment.

Google.PowerShell/BigQuery/BqJob.cs
ILMTitan: Darn, that's too bad.

ahandley:
Thanks!

PR #513, opened 2017-04-08
Finished Start-BqJob + tests

ahandley:
Super long PR, but most of it is just *all* of the Load parameters and the tests.  

That's the rest of Start-BqJob, so it includes the -Load and -Extract functionality.

### DIFF
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
@@ -213,6 +213,125 @@ Describe "BqJob-Copy" {
     }
 }
 
+Describe "BqJob-EL" {
+
+    BeforeAll {
+        $r = Get-Random
+        $datasetName = "pshell_testing_$r"
+        $test_set = New-BqDataset $datasetName
+        $folder = Get-Location
+        $folder = $folder.ToString()
+        $filename = "$folder\classics.csv"
+        $table = New-BqTable -Dataset $test_Set "table_$r"
+        New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
+            New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $table | 
+            Add-BqTabledata $filename CSV -SkipLeadingRows 1
+        $bucket = New-GcsBucket "ps_test_$r"
+        $gcspath = "gs://ps_test_$r"
+    }
+
+    It "should set Load parameters correctly" {
+        $alt_tab = $test_set | New-BqTable "param_test_$r"
+        $table | Start-BqJob -Extract CSV "$gcspath/param.csv" -Synchronous
+        New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
+            New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $alt_tab
+        $job = $alt_tab | Start-BqJob -Load CSV "$gcspath/param.csv" -WriteMode WriteAppend `
+            -Encoding "ISO-8859-1" -FieldDelimiter "|" -Quote "'" -MaxBadRecords 3 `
+            -SkipLeadingRows 2 -AllowUnknownfields -AllowJaggedRows -AllowQuotedNewlines
+        $job.Configuration.Load.AllowJaggedRows | Should Be $true
+        $job.Configuration.Load.AllowQuotedNewlines | Should Be $true
+        $job.Configuration.Load.Encoding | Should Be "ISO-8859-1"
+        $job.Configuration.Load.FieldDelimiter | Should Be "|"
+        $job.Configuration.Load.IgnoreUnknownValues | Should Be $true
+        $job.Configuration.Load.MaxBadRecords | Should Be 3
+        $job.Configuration.Load.Quote | Should Be "'"
+        $job.Configuration.Load.SkipLeadingRows | Should Be 2
+        $job.Configuration.Load.SourceFormat.ToString() | Should Be "CSV"
+        $job.Configuration.Load.WriteDisposition.ToString() | Should Be "WRITE_APPEND"
+    }
+
+    It "should set Extract parameters correctly" {
+        $job = $table | Start-BqJob -Extract CSV "$gcspath/otherparam.csv" -Compress -FieldDelimiter "|" -NoHeader
+        $job.Configuration.Extract.Compression | Should Be $true
+        $job.Configuration.Extract.DestinationFormat.ToString() | Should Be "CSV"
+        $job.Configuration.Extract.DestinationUri | Should Be "$gcspath/otherparam.csv"
+        $job.Configuration.Extract.FieldDelimiter | Should Be "|"
+        $job.Configuration.Extract.PrintHeader | Should Be $false
+    }
+
+    It "should Extract/Load a basic CSV" {
+        $job = $table | Start-BqJob -Extract CSV "$gcspath/basic.csv" -Synchronous
+        $job.Status.State | Should Be "DONE"
+        $job.Status.ErrorResult | Should Be $null
+
+        $file = Get-GcsObject "ps_test_$r" "basic.csv"
+        $file | Should Not Be $null
+
+        $alt_tab = $test_set | New-BqTable "basic_test_$r"
+        New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
+            New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $alt_tab
+        $boj = $alt_tab | Start-BqJob -Load CSV "$gcspath/basic.csv" -SkipLeadingRows 1 -Synchronous
+        $boj.Status.State | Should Be "DONE"
+
+        $alt_tab = $alt_tab | Get-BqTable
+        $alt_tab.NumRows | Should Be 10
+    }
+
+    It "should Extract/Load a more complex CSV request" {
+        $job = $table | Start-BqJob -Extract CSV "$gcspath/complex.csv" -FieldDelimiter "|" -NoHeader -Synchronous
+        $job.Status.State | Should Be "DONE"
+        $job.Status.ErrorResult | Should Be $null
+
+        $file = Get-GcsObject "ps_test_$r" "basic.csv"
+        $file | Should Not Be $null
+
+        $alt_tab = $test_set | New-BqTable "complex_test_$r"
+        New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
+            New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $alt_tab
+        $boj = $alt_tab | Start-BqJob -Load CSV "$gcspath/complex.csv" -FieldDelimiter "|" -Synchronous
+        $boj.Status.State | Should Be "DONE"
+
+        $alt_tab = $alt_tab | Get-BqTable
+        $alt_tab.NumRows | Should Be 10
+    }
+
+    It "should handle WriteMode correctly" {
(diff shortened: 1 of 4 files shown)


Google.PowerShell/BigQuery/BqCmdlet.cs
```diff
@@ -25,4 +25,12 @@ public BqCmdlet()
             _client = new Lazy<BigQueryClient>(() => BigQueryClient.Create(Project));
         }
     }
+
+    /// <summary>
+    /// Data formats for input and output
+    /// </summary>
+    public enum DataFormats
+    {
+        AVRO, CSV, JSON, DATASTORE_BACKUP
```
iantalarico: A comment for each of these would be nice, even if they are small.

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -214,6 +214,18 @@ public void DoGetRequest()
     ///   </code>
     ///   <para>Copies the contents of the source as long as the source and destination schemas match.</para>
     /// </example>
+    /// <example>
+    ///   <code>
+    /// PS C:\> $job = $table | Start-BqJob -Load CSV "$gcspath/basic.csv" -SkipLeadingRows 1 -Synchronous
```
iantalarico: I think it's a little confusing having variables that aren't defined in the example. Can they be defined?

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -353,6 +369,90 @@ private class ParameterSetNames
         [Parameter(Mandatory = true, ParameterSetName = ParameterSetNames.DoLoad)]
         public SwitchParameter Load { get; set; }
 
+        /// <summary>
+        /// <para type="description">
+        /// The format to input/output (CSV, JSON, AVRO).
```
iantalarico: Should this list also have DATASTORE_BACKUP?

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -353,6 +369,90 @@ private class ParameterSetNames
...
+        /// The value that is used to quote data sections in a CSV file. Default value is double-quote (").
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.DoLoad)]
+        public string Quote { get; set; } = "\"";
+
+        /// <summary>
+        /// <para type="description">
+        /// The number of malformed rows that the request will ignore before throwing an error. 
+        /// This value is zero if not specified.
```
iantalarico: "The default is zero...."?

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -484,8 +607,41 @@ public Apis.Bigquery.v2.Data.Job DoCopy()
...
+                                MaxBadRecords = MaxBadRecords,
+                                SkipLeadingRows = SkipLeadingRows,
+                                IgnoreUnknownValues = AllowUnknownFields,
+                                AllowJaggedRows = AllowJaggedRows,
+                                AllowQuotedNewlines = AllowQuotedNewlines,
+                                WriteDisposition = WriteMode.ToString()
+                            }
+                        }
+                    };
+                    //deal with source format
```
iantalarico: What's this comment for?

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -484,8 +607,41 @@ public Apis.Bigquery.v2.Data.Job DoCopy()
...
+                    return Service.Jobs.Insert(loadjob, Project).Execute();
+                }
+                catch (Exception ex)
```
iantalarico: I don't think you should catch everything here.

iantalarico: Should you catch 403s? 404s?

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -495,8 +651,34 @@ public Apis.Bigquery.v2.Data.Job DoLoad()
...
+                                DestinationFormat = (Type == DataFormats.JSON) ? "NEWLINE_DELIMITED_JSON" : Type.ToString(),
+                                Compression = (Compress) ? "GZIP" : "NONE",
+                                PrintHeader = !NoHeader,
+                                FieldDelimiter = FieldDelimiter
+                            }
+                        }
+                    };
+                    return Service.Jobs.Insert(extractjob, Project).Execute();
+                }
+                catch (Exception ex)
```
iantalarico: Same here and below

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -484,8 +607,41 @@ public Apis.Bigquery.v2.Data.Job DoCopy()
...
+                    //deal with source format
+                    return Service.Jobs.Insert(loadjob, Project).Execute();
+                }
+                catch (Exception ex)
+                {
+                    ThrowTerminatingError(new ErrorRecord(ex, "Load Failed",
+                        ErrorCategory.InvalidOperation, this));
+                }
+            }
+            return null;
```
iantalarico: Why do you need null here?  Doesn't this throw or return above?

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -495,8 +651,34 @@ public Apis.Bigquery.v2.Data.Job DoLoad()
...
+                {
+                    var extractjob = new Apis.Bigquery.v2.Data.Job()
+                    {
+                        Configuration = new JobConfiguration()
+                        {
+                            Extract = new JobConfigurationExtract()
+                            {
+                                SourceTable = Source,
+                                DestinationUris = DestinationUris,
+                                DestinationFormat = (Type == DataFormats.JSON) ? "NEWLINE_DELIMITED_JSON" : Type.ToString(),
```
iantalarico: There are a fair amount of string constants here.  I would put them in variables and comment them.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -213,6 +213,125 @@ Describe "BqJob-Copy" {
...
+        $filename = "$folder\classics.csv"
+        $table = New-BqTable -Dataset $test_Set "table_$r"
+        New-BqSchema -Name "Title" -Type "STRING" | New-BqSchema -Name "Author" -Type "STRING" |
+            New-BqSchema -Name "Year" -Type "INTEGER" | Set-BqSchema $table | 
+            Add-BqTabledata $filename CSV -SkipLeadingRows 1
+        $bucket = New-GcsBucket "ps_test_$r"
+        $gcspath = "gs://ps_test_$r"
+    }
+
+    It "should set Load parameters correctly" {
```
iantalarico: I think it would be nice to have a test that also checks defaults are properly set too. (and below)

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
```diff
@@ -213,6 +213,125 @@ Describe "BqJob-Copy" {
...
+        $alt_tab.NumRows | Should Be 20
+
+        $alt_tab | Start-BqJob -Load CSV "$gcspath/write.csv" -SkipLeadingRows 1 `
+            -WriteMode WriteTruncate -Synchronous
+        $alt_tab = $alt_tab | Get-BqTable
+        $alt_tab.NumRows | Should Be 10
+    }
+
+    It "Load should properly halt when -WhatIf is passed" {
+        $table | Start-BqJob -Load AVRO "$gcspath/whatif.json" -WhatIf | Should Be $null
```
iantalarico: I didn't see "WhatIf" below.  Is this a PS thing? Does this halt it or stop it before it starts?

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -462,7 +585,7 @@ public Apis.Bigquery.v2.Data.Job DoCopy()
                             {
                                 DestinationTable = Destination,
                                 SourceTable = Source,
-                                WriteDisposition = WriteMode.ToString()
+                                WriteDisposition = (WriteMode == null) ? WriteDisposition.WriteIfEmpty.ToString() : WriteMode.ToString()
```
ILMTitan: This could be written as WriteMode?.ToString() ?? WriteDisposition.WriteIfEmpty.ToString();

Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -484,8 +607,41 @@ public Apis.Bigquery.v2.Data.Job DoCopy()
         /// </summary>
         public Apis.Bigquery.v2.Data.Job DoLoad()
         {
-            throw new NotImplementedException(
-                "Load jobs are not implemented yet.  Use Set-BqTabledata instead.");
+            if (ShouldProcess($"{Destination.TableId} (LOAD)"))
+            {
+                try
+                {
+                    var loadjob = new Apis.Bigquery.v2.Data.Job()
```
ILMTitan: When doing object initialization with an empty constructor, leave out the `()`.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: That was for the compiler.  It's for the case that we Should (not) Process, and then neither the return or the throw will be hit.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqJob.Tests.ps1
ahandley: -WhatIf is a PS [builtin](https://msdn.microsoft.com/en-us/library/system.management.automation.cmdletcommonmetadataattribute.supportsshouldprocess(v=vs.85).aspx) that does effectively the same thing as the dry run flag of BqJobs - it outputs to the user some information about what would happen if this cmd would run, so they can judge whether it has the intended consequences.  

I added it to the cmdlet by specifying:
`[Cmdlet("Start", "BqJob", SupportsShouldProcess = true)]`

Google.PowerShell/BigQuery/BqJob.cs
ahandley: Got it!  That's a pretty cool convention.  Didn't realize that could be used in between an object and a method call.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: Got it.  (fixed for the whole file, and I'll add that to my list to fix elsewhere)

Google.PowerShell/BigQuery/BqCmdlet.cs
ahandley: Got it!

Google.PowerShell/BigQuery/BqJob.cs
ahandley: Yep.  Just added in the setup cmdlets.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: Added.  Wrote the list and then found out there was an additional possibility!

Google.PowerShell/BigQuery/BqJob.cs
ahandley: "The default value is zero."  yep.

Google.PowerShell/BigQuery/BqJob.cs
ahandley: It should have been deleted with a block of old code that I was using as a workaround.

Google.PowerShell/BigQuery/BqJob.cs

Google.PowerShell/BigQuery/BqJob.cs
ahandley: 404's, 403's, and 409's are the main ones to look out for with BQ.  I'll catch the ones that are appropriate per function.

ahandley:
Alright, I addressed the comments above.  PTAL.

ahandley:
Thanks, yall.

PR #515, opened 2017-04-15
Code-heavy P1 bugs

ahandley:
Set- should add on 404  (for databases and tables both)
Switch to Add-BqTableRows and Get-BqTableRows  (rename)
Schema builder from JSON string or file
Add schema to table creation with flag
(p2) Fix Add-TableRows: -Format [CSV|JSON] (reorder params to match job syntax)


ahandley:
Fixed this, but I already have code revisions on top of it.  Going to be pushed in the next update.

PR #516, opened 2017-04-20
p2 fixes + testing junk

ahandley:
Again, a massive commit.  Lots of little fixes and text edits.  

Major feature to review is the Label flags in Get- and Set-BqDataset.  Tests are added too.

### DIFF
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqDataset.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqDataset.Tests.ps1
@@ -19,18 +19,22 @@ Describe "Get-BqDataset" {
         try {
             $data1 = New-BqDataset "test_label_1"
             $data2 = New-BqDataset "test_label_2"
-            $data2 = $data2 | Set-BqDataset -SetLabel "test" "one"
+            $data2 = $data2 | Set-BqDataset -SetLabel @{"test"="one"}
             $data3 = New-BqDataset "test_label_3"
-            $data3 = $data3 | Set-BqDataset -SetLabel "test","other" "three","two"
+            $data3 = $data3 | Set-BqDataset -SetLabel @{"test"="three";"other"="two"}
             
             $ds = Get-BqDataset -Filter "test"
             $ds.Count | Should Be 2
+            $ds[0].DatasetReference.DatasetId | Should Not Be "test_label_1"
+            $ds[1].DatasetReference.DatasetId | Should Not Be "test_label_1"
 
             $ds = Get-BqDataset -Filter "test:one"
             $ds.Count | Should Be 1
+            $ds[0].DatasetReference.DatasetId | Should Be "test_label_2"
 
             $ds = Get-BqDataset -Filter "other:two"
             $ds.Count | Should Be 1
+            $ds[0].DatasetReference.DatasetId | Should Be "test_label_3"
         } finally {
             Get-BqDataset "test_label_1" | Remove-BqDataset
             Get-BqDataset "test_label_2" | Remove-BqDataset
@@ -158,7 +162,7 @@ Describe "Set-BqDataset" {
     It "should label things properly" {
         try {
             $data = New-BqDataset "test_label_1"
-            $data = $data | Set-BqDataset -SetLabel "test" "one"
+            $data = $data | Set-BqDataset -SetLabel @{"test"="one"}
             $data.Labels.Count | Should Be 1
             $data.Labels["test"] | Should Be "one"
         } finally {
@@ -169,7 +173,7 @@ Describe "Set-BqDataset" {
     It "should handle a lot of labels at once" {
         try {
             $data = New-BqDataset "test_label_2"
-            $data = $data | Set-BqDataset -SetLabel "test","other","third" "one","two","three"
+            $data = $data | Set-BqDataset -SetLabel @{"test"="one";"other"="two";"third"="three"}
             $data.Labels.Count | Should Be 3
             $data.Labels["test"] | Should Be "one"
             $data.Labels["other"] | Should Be "two"
@@ -182,7 +186,7 @@ Describe "Set-BqDataset" {
     It "should clear just one label" {
         try {
             $data = New-BqDataset "test_label_3"
-            $data = $data | Set-BqDataset -SetLabel "test","other" "one","two"
+            $data = $data | Set-BqDataset -SetLabel @{"test"="one";"other"="two"}
             $data = $data | Set-BqDataset -ClearLabel "test"
             $data.Labels.Count | Should Be 1
             $data.Labels["other"] | Should Be "two"
@@ -191,11 +195,11 @@ Describe "Set-BqDataset" {
         }
     }
 
-    It "should clear multiple labels" {
+    It "should clear multiple labels + nonexistant labels" {
         try {
             $data = New-BqDataset "test_label_4"
-            $data = $data | Set-BqDataset -SetLabel "test","other" "one","two"
-            $data = $data | Set-BqDataset -ClearLabel "test", "other"
+            $data = $data | Set-BqDataset -SetLabel @{"test"="one";"other"="two"}
+            $data = $data | Set-BqDataset -ClearLabel "test","other","doesnotexist"
             $data.Labels | Should Be $null
         } finally {
             Get-BqDataset "test_label_4" | Remove-BqDataset
--- a/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqDataset.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqDataset.Tests.ps1
@@ -15,6 +15,29 @@ Describe "Get-BqDataset" {
         }
     }
 
+    It "should list by label" {
+        try {
+            $data1 = New-BqDataset "test_label_1"
+            $data2 = New-BqDataset "test_label_2"
+            $data2 = $data2 | Set-BqDataset -SetLabel "test" "one"
+            $data3 = New-BqDataset "test_label_3"
+            $data3 = $data3 | Set-BqDataset -SetLabel "test","other" "three","two"
+            
+            $ds = Get-BqDataset -Filter "test"
+            $ds.Count | Should Be 2
+
+            $ds = Get-BqDataset -Filter "test:one"
+            $ds.Count | Should Be 1
+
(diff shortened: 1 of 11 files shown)


Google.PowerShell/BigQuery/BqDataset.cs
```diff
@@ -242,16 +269,83 @@ public class SetBqDataset : BqCmdlet
...
+        [Parameter(Mandatory = true, ValueFromPipeline = true, ParameterSetName = ParameterSetNames.ClearLabel)]
         public Dataset Dataset { get; set; }
 
+        /// <summary>
+        /// <para type="description">
+        /// Sets the labels in Keys to the values in Values for the target Dataset.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, ParameterSetName = ParameterSetNames.SetLabel)]
+        public SwitchParameter SetLabel { get; set; }
```
quoctruong: Why don't we just use a HashTable instead of having a SwitchParameter with 2 arrays?

ILMTitan: I agree. Two lists related by index is a bad format to input data. An array to remove and a hash table to add is a much nicer format, and has the added benefit of letting you do both operations  in one command.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqDataset.Tests.ps1
```diff
@@ -15,6 +15,29 @@ Describe "Get-BqDataset" {
...
+            $data2 = New-BqDataset "test_label_2"
+            $data2 = $data2 | Set-BqDataset -SetLabel "test" "one"
+            $data3 = New-BqDataset "test_label_3"
+            $data3 = $data3 | Set-BqDataset -SetLabel "test","other" "three","two"
+            
+            $ds = Get-BqDataset -Filter "test"
+            $ds.Count | Should Be 2
+
+            $ds = Get-BqDataset -Filter "test:one"
+            $ds.Count | Should Be 1
```
ILMTitan: It feels like you should add a check that you get the correct one.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqDataset.Tests.ps1
```diff
@@ -120,16 +143,64 @@ Describe "Set-BqDataset" {
...
+        try {
+            $data = New-BqDataset "test_label_4"
+            $data = $data | Set-BqDataset -SetLabel "test","other" "one","two"
+            $data = $data | Set-BqDataset -ClearLabel "test", "other"
+            $data.Labels | Should Be $null
+        } finally {
+            Get-BqDataset "test_label_4" | Remove-BqDataset
+        }
+    }
 }
```
ILMTitan: You don't have a test for miss-numbered `-SetLabel` arrays, but fixing the interface will make that irrelevant.

Google.PowerShell/BigQuery/BqDataset.cs
```diff
@@ -192,6 +191,21 @@ private Dataset DoGetRequest(string project, string dataset)
...
+
+        /// <summary>
+        /// Helper function to take an array of key:value filters and turn it into a filter string.
+        /// </summary>
+        /// <returns></returns>
+        public string MakeFilter()
+        {
+            if (Filter == null) { return null; }
+            string labels = "";
+            foreach (string f in Filter)
```
ahandley: Got it.  Inlined the MakeFilter function as well, since it was only two lines.

Google.PowerShell/BigQuery/BqDataset.cs
```diff
@@ -242,16 +269,83 @@ public class SetBqDataset : BqCmdlet
...
+                        Dataset.Labels.Remove(key);
+                    }
+                    break;
+                default:
+                    throw UnknownParameterSetException;
+            }
+
             Dataset response;
-            bool needToInsert = false;
+            Dataset.ETag = null;
```
ahandley: Yes, precisely.  The previous Set- tests were failing about 25% of the time because of an if-match header exception, and now when I'm doing stuff with labels, they fail 100% of the time.  I thought the ETag conflict happened because I was manually changing values of the Dataset with a Set cmdlet, so I started nulling it out before returning it.

Previously, I thought the ETag was only for list requests, but I was just in the wrong section of documentation.  Now that I found the correct [set of docs](https://cloud.google.com/bigquery/docs/api-performance#patch). It looks like I need to leave the ETag alone.  I commented out the nullification, and it's back to failing 100% of the time.  I don't modify the ETag value at all, so do you have any idea why it's failing with that precondition?

Google.PowerShell/BigQuery/BqDataset.cs
ahandley: So this actually doesn't work.  It looked nice and compiled, but that misses out on the addition of "label." to the beginning of each term.  

Thought about doing `string.Join(" ", Filter)`, but that would end up with 
"test.one labels.other.two" when we need "labels.test.one labels.other.two"

Just going to revert to the original solution.

Google.PowerShell/BigQuery/BqDataset.cs
ahandley: Got it.  Much cleaner now.  Thanks!

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqDataset.Tests.ps1
ahandley: Got it.  Also added checks to the first and third calls.

Google.PowerShell.IntegrationTests/BigQuery/BigQuery.BqDataset.Tests.ps1
ahandley: I tested that in-person.  Should have put that in as well, but yeah.  Went around the issue.

ahandley:
@ILMTitan ok, so look at this crap.  

```
PS C:\Users\alexhandley> $data = New-BqDataset "test_label_1"
PS C:\Users\alexhandley> $data = $data | Set-BqDataset -SetLabel @{"test"="one"}
    Set-BqDataset : Google.Apis.Requests.RequestError #(trimmed)
PS C:\Users\alexhandley> $data = $data | Get-BqDataset
PS C:\Users\alexhandley> $data = $data | Set-BqDataset -SetLabel @{"test"="one"}
PS C:\Users\alexhandley>#it worked.
```

So the issue is `New-BqDataset` doesn't return ETag correctly.  I checked and it does return *an* etag, its just different from the one that is returned by all subsequent Get- calls:

```
PS C:\Users\alexhandley> $data = New-BqDataset "test_label_1"
PS C:\Users\alexhandley> $data
ETag                     : "-AzRU5skO4l0G5c75S4R78a9UuM/LOIlgWANjHx8uscCYuuDicWYCLA"

PS C:\Users\alexhandley> $data | Get-BqDataset
ETag                     : "-AzRU5skO4l0G5c75S4R78a9UuM/MTQ5MjgwNjQ0OTgzOQ"

PS C:\Users\alexhandley> $data | Get-BqDataset
ETag                     : "-AzRU5skO4l0G5c75S4R78a9UuM/MTQ5MjgwNjQ0OTgzOQ"

```

I'm assuming at this point that it's a bug with the library, so I'm going to go pick around that code and see what's happening.

Google.PowerShell/BigQuery/BqDataset.cs
quoctruong: How about `string.Join(" ", Filter.select(filter => $"labels.{filter}"))`?

Google.PowerShell/BigQuery/BqDataset.cs
quoctruong: What's the error that you get when it fails?

ahandley:
```
Set-BqDataset : Google.Apis.Requests.RequestError
Resource gcloud-powershell-testing:test_label_1 did not meet condition IF_MATCH. This may be caused by concurrent
updates to a resource. [412]
Errors [
    Message[Resource gcloud-powershell-testing:test_label_1 did not meet condition IF_MATCH. This may be caused by
concurrent updates to a resource.] Location[If-Match - header] Reason[preconditionFailed] Domain[global]
]
At line:1 char:17
+ $data = $data | Set-BqDataset -SetLabel @{"test"="one"}
+                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Set-BqDataset], GoogleApiException
    + FullyQualifiedErrorId : Google.GoogleApiException,Google.PowerShell.BigQuery.SetBqDataset

```

ahandley: Also if you haven't already, [see this comment below](https://github.com/GoogleCloudPlatform/google-cloud-powershell/pull/516#issuecomment-296298776).

quoctruong: What if you do a get first to get the ETag and then call the update with that ETag?

ahandley: It works perfectly.  That's why my other tests are succeeding.  I do new, get, then set.  But if New- returns a Dataset, it should be able to be updated or labeled like normal.  Or it at least needs to be documented what can and can't be done with it.

quoctruong: I think we should go with calling a Get first then? In the docs you link above, they show that example too. Perhaps the new case is undocumented.

ahandley: That'll definitely work for the unit tests, but I'm also going to have to document that heavily in the Set- documentation.  I'm also going to look into tables to see new returns a bad etag there as well.

quoctruong: Oh what I mean is you do that in the cmdlet. Do a get call and then do an update call. So user won't have to do that.

ahandley: Oh, gotcha.  Yeah, it's probably worth the performace hit.  I was trying to avoid that, but it looks like that's the best option if that's what works.

ahandley:
Ok, updated the code for proper handling of Etags and fixed the filter syntax.  PTAL @ILMTitan , @quoctruong .

ahandley:
Thanks!

PR #518, opened 2017-04-26
BQ Docs

ahandley:


js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
       the Cloud Tools for PowerShell module via <code>Import-Module GoogleCloud</code>.
     </p>
   </div>
+  
+    <!-- Google Cloud BigQuery -->
```
chrsmith: Align with next div

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+      Google BigQuery is a versatile tool that solves the problem of storing and querying massive 
+      datasets without having to worry about data formats, underlying resources, and other things 
+      that distract you from your analysis.
+    </p>
+    
+    <h2>Dataset</h2>
+    <p>
+      To use BigQuery in a Cloud project, first create a <em>Dataset</em> using the 
+      <cmdlet-ref name="New-BqDataset"></cmdlet-ref> cmdlet. This will take in basic information and 
+      create the resource serverside. Locally, a reference Dataset object is returned. To get a reference
```
chrsmith: "server-side" with hyphen.

chrsmith: Use `<code>` for Dataset, since it is a code object?

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+      object for an existing dataset, use <cmdlet-ref name="Get-BqDataset"></cmdlet-ref>.
+    </p>
+    <pre>
+# Makes a new dataset with DatasetId "my_dataset".
+$dataset = New-BqDataset "my_dataset" -Name "Test Dataset" -Description "Some test data for BigQuery"
+
+# The following two lines do the same thing.
```
chrsmith: "do the same thing" isn't useful, since you don't explain what they do.

````
# Two ways to get a dataset: by name or by existing Dataset object.
````

File: js/app/controllers/content-product.ng [hunk 4]
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+$dataset = Get-BqDataset "my_dataset"
+$dataset = $dataset | Get-BqDataset
+    </pre>
+    <p>
+      This object <em>$dataset</em> can be modified and passed into further cmdlets such as
```
chrsmith: `<code>` instead of `<em>`? Just be consistent with other sections of the docs.

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+    </pre>
+    <p>
+      Datasets can be deleted by the <cmdlet-ref name="Remove-BqDataset"></cmdlet-ref> cmdlet. This 
+      cmdlet supports ShouldProcess (the <code>-WhatIf</code> parameter) and will prompt for user 
+      confirmation before deleting a non-empty Dataset. This safeguard can be bypassed with the 
+      <code>-Force</code> parameter when scripting.  
+    </p>
+    <pre>
+# Deletes $dataset.
+$dataset | Remove-BqDataset -Force
```
chrsmith: Replace `-Force` with `-WhatIf`. Docs shouldn't encourage users to `# YOLO` things. :D

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+    </pre>
+    
+    <h2>Table</h2>
+    <p>
+      Each Dataset has a number of <em>Tables</em> to hold data. Tables are created with the 
+      <cmdlet-ref name="New-BqTable"></cmdlet-ref> cmdlet by passing in a TableId and a Dataset 
+      where the table will reside. The Dataset can be passed with a Dataset object or with the
+      <code>-DatasetId</code> parameter. <cmdlet-ref name="Get-BqTable"></cmdlet-ref> and 
+      <cmdlet-ref name="Set-BqTable"></cmdlet-ref> work the same way as the Get- and Set- cmdlets above.
```
chrsmith: `<code>Get-</code>` and `<code>`Set-`</code>`.

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+      Tables need <em>Schemas</em> to describe the format of the data they contain. Schemas are created with
+      the <cmdlet-ref name="New-BqSchema"></cmdlet-ref> and <cmdlet-ref name="Set-BqSchema"></cmdlet-ref>
+      cmdlets. New-BqSchema can take the formats for rows as parameters directly or as a JSON array of 
+      row descriptions. The results of New-BqSchema are always passed into Set-BqSchema which can either output
+      a Schema object or assign the Schema to an existing Table.  
+    </p>
+    <pre>
+# Assigns a Schema to $table
+$table = Get-BqTable "my_table" -DatasetId "my_dataset"
+New-BqSchema "Author" "STRING" | New-BqSchema "Copyright" "STRING" |
```
chrsmith: PowerShell line breaks require a backtick character (`). Please add them so users can copy+paste.

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+      <cmdlet-ref name="Start-BqJob"></cmdlet-ref> starts any of these kinds of jobs as an asynchronous 
+      operation. Use the <code>-PollUntilComplete</code> flag to have the cmdlet block until the job is done.
+      <cmdlet-ref name="Receive-BqJob"></cmdlet-ref> will return the results of a query job once it is 
+      finished. <cmdlet-ref name="Get-BqJob"></cmdlet-ref> will return a reference object detailing the
+      current state and statistics on the job.  <cmdlet-ref name="Stop-BqJob"></cmdlet-ref> will send a 
+      request to the server to stop a certain job, and then returns immediately.
+    </p>
+    <pre>
+# Query Job: starts a query and outputs results into $table.
+Start-BqJob -Query "SELECT * FROM ``my_dataset:my_table``" -Destination $table
```
chrsmith: Is it really "``my_dataset:table``"? That seems strange. Is there another way we can escape/unescape the string in PowerShell? Or is that just how BigQuery rolls?

ahandley: BigQuery needs the first set of `` to denote the datasetid:tableid, and PowerShell treats ` chars as newlines, so you need the second set of `` to escape it.  I used the same concept in my bq4ps demo.  Took me a bit to figure out why it wasn't accepting my table names when I just had `d:t`

js/app/controllers/content-product.ng
ahandley: It was my understanding that you can have a line break on a | character without a backtick.  I just tested copy-pasting this command, and it works.

chrsmith: Probably because it ends with `|` and just "figures it out". The F# dev in me wants to put that on the next line, which would probably cause the PowerShell parser to choke... Just keep it as-is I guess. Thanks for following up.

js/app/controllers/content-product.ng
chrsmith: Very interesting. Why don't you call that out in the docs. Perhaps `<blockquote>` or a separate paragraph? This is definitely something that I wouldn't expect users to pick up on.

ahandley: I was at first just thinking add it to the comment on line 273, but perhaps a blockquote would be better for emphasis and keeping things neat.

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+    </pre>
+    <p>
+      This object <em>$dataset</em> can be modified and passed into further cmdlets such as 
+      <cmdlet-ref name="Set-BqDataset"></cmdlet-ref> to manipulate cloud resources. This cmdlet also
+      handles adding and removing labels with <code>-SetLabel</code> and <code>-ClearLabel</code>.
+    </p>
+    <pre>
+# Updates the Name field of $dataset.
+$dataset.Name = "A Different Name"
+$dataset = Set-BqDataset "my_dataset"
```
quoctruong: I feel like you can just run `Set-BqDataset "my_dataset"` without assigning it to a variable since you are not really using it anywhere in this example?

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+    <h2>Dataset</h2>
+    <p>
+      To use BigQuery in a Cloud project, first create a <em>Dataset</em> using the 
+      <cmdlet-ref name="New-BqDataset"></cmdlet-ref> cmdlet. This will take in basic information and 
+      create the resource serverside. Locally, a reference Dataset object is returned. To get a reference
+      object for an existing dataset, use <cmdlet-ref name="Get-BqDataset"></cmdlet-ref>.
+    </p>
+    <pre>
+# Makes a new dataset with DatasetId "my_dataset".
+$dataset = New-BqDataset "my_dataset" -Name "Test Dataset" -Description "Some test data for BigQuery"
```
quoctruong: No need to assign the result of `New-BqDataset` since we are not using it anywhere else.

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+# Load Job: adds TableRows to $table from the GCS file specified.
+$job = $dest_table | Start-BqJob -Load CSV "gs://my_dataset/data.csv"
+    </pre>
+    <pre>
+# Extract Job: exports $src_table to a GCS file.
+$job = $src_table | Start-BqJob -Extract CSV "gs://my_dataset/data.csv"
+    </pre>
+    <pre>
+# Copy Job: Starts a copy job, cancels it, and polls until the job is complely done.
+$job = $table | Start-BqJob -Copy $target
```
quoctruong: What is `$target` here, is it a table name? If so, then I think it would be better if we use `"destination_table"` or something descriptive.

js/app/controllers/content-product.ng
ILMTitan: I disagree. Showing that `New-BqDataset` has a return value is useful. Also, this is a snippet to be used as part of a larger script where the user might want to have the dataset object.

js/app/controllers/content-product.ng
ILMTitan: Again, I disagree.

js/app/controllers/content-product.ng
ahandley: Yep.  Jim is right in what I was going for.  The majority of the use cases for New-BqDataset result in the user capturing the returned object.

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+$dataset = New-BqDataset "my_dataset" -Name "Test Dataset" -Description "Some test data for BigQuery"
+
+# The following two lines do the same thing.
+$dataset = Get-BqDataset "my_dataset"
+$dataset = $dataset | Get-BqDataset
+    </pre>
+    <p>
+      This object <em>$dataset</em> can be modified and passed into further cmdlets such as 
+      <cmdlet-ref name="Set-BqDataset"></cmdlet-ref> to manipulate cloud resources. This cmdlet also
+      handles adding and removing labels with <code>-SetLabel</code> and <code>-ClearLabel</code>.
```
ahandley: Added in a blockquote and another example to show how to use them.

js/app/controllers/content-product.ng
ahandley: Got all of these small fixes done.  Thanks!

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+    </p>
+    <pre>
+# Updates the Name field of $dataset.
+$dataset.Name = "A Different Name"
+$dataset = Set-BqDataset "my_dataset"
 
+# Adds the labels 'test' and 'other' to $dataset.
+$dataset = Set-BqDataset "my_dataset" -SetLabel @{"test"="one";"other"="two"}
```
ahandley: Got it!

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,158 @@ mkdir my-new-bucket</pre>
...
+    </p>
+    <pre>
+# Deletes $dataset.
+$dataset | Remove-BqDataset -Force
+    </pre>
+    
+    <h2>Table</h2>
+    <p>
+      Each Dataset has a number of <em>Tables</em> to hold data. Tables are created with the 
+      <cmdlet-ref name="New-BqTable"></cmdlet-ref> cmdlet by passing in a TableId and a Dataset
```
ahandley: Got it as per Chris's suggestion above.  I just went through the entire doc and threw <code> flags all over the place.  (Where they belonged, that is)

js/app/controllers/content-product.ng
ahandley: swapped it to $dest_table because that's what I used above.  Good catch.  $target was a unit test name that I had.

ahandley:
@chrsmith and @quoctruong 
Made all of the requested changes.  Also added meaningful table names as per the suggestion up top. 

PTAL.

js/app/controllers/content-product.ng
```diff
@@ -218,72 +227,79 @@ $table | Remove-BqTable -Force
...
+      <em>Extract</em> jobs export BigQuery tables to GCS. <em>Copy</em> jobs copy an existing table to another new or 
+      existing table. <cmdlet-ref name="Start-BqJob"></cmdlet-ref> starts any of these kinds of jobs as an asynchronous 
       operation. Use the <code>-PollUntilComplete</code> flag to have the cmdlet block until the job is done.
       <cmdlet-ref name="Receive-BqJob"></cmdlet-ref> will return the results of a query job once it is 
       finished. <cmdlet-ref name="Get-BqJob"></cmdlet-ref> will return a reference object detailing the
       current state and statistics on the job.  <cmdlet-ref name="Stop-BqJob"></cmdlet-ref> will send a 
       request to the server to stop a certain job, and then returns immediately.
+      <blockquote>
+      Note on formatting table names within query strings: BigQuery format specifies that table names 
+      should be surrounded by backticks (`), but this is Powershell's "continue on the next line"
```
ILMTitan: The problem is not that it is the "continue on the next line" operator, the problem is it is PowerShell's escape character.

ahandley: Re-worded it!

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,173 @@ mkdir my-new-bucket</pre>
...
+    <h2>Dataset</h2>
+    <p>
+      To use BigQuery in a Cloud project, first create a <code>Dataset</code> using the 
+      <cmdlet-ref name="New-BqDataset"></cmdlet-ref> cmdlet. This will take in basic information and 
+      create the resource server-side. Locally, a <code>Dataset</code> <em>reference object</em> is returned. 
+      To get a reference object for an existing dataset, use <cmdlet-ref name="Get-BqDataset"></cmdlet-ref>.
+    </p>
+    <pre>
+# Makes a new dataset with DatasetId "page_views".
+$dataset = New-BqDataset "page_views" -Name "Page Views" -Description "Page views from 2014 onwards"
```
quoctruong: We should make `-Description "Page views from 2014 onwards"` go to the next line (using backtick).

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,173 @@ mkdir my-new-bucket</pre>
...
+      The <code>Get-BqDataset</code> cmdlet has a built in <code>-Filter</code> flag that allows fine grained 
+      control when listing datasets for processing with other cmdlets.
+    </blockquote>
+    <pre>
+# Updates the Name field of $dataset.
+$dataset.Name = "PageView Data"
+$dataset = Set-BqDataset "page_views"
 
+# Adds the labels 'department' and 'purpose' to $dataset.
+$dataset = Set-BqDataset "page_views" -SetLabel @{"department" = "statistics"; "purpose" = "metrics"}
```
quoctruong: We should split this into 2 lines too.

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,173 @@ mkdir my-new-bucket</pre>
...
+
+# Filter Datasets by their labels and return all statistics department Datasets.
```
quoctruong: Split this into 2 lines.

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,173 @@ mkdir my-new-bucket</pre>
...
+      Each <code>Dataset</code> has a number of <code>Tables</code> to hold data. <code>Tables</code> are 
+      created with the <cmdlet-ref name="New-BqTable"></cmdlet-ref> cmdlet by passing in a <em>TableId</em> 
+      and the <code>Dataset</code> where the table will reside. The <code>Dataset</code> can be passed in by object 
+      or with the <code>-DatasetId</code> parameter. <cmdlet-ref name="Get-BqTable"></cmdlet-ref> and 
+      <cmdlet-ref name="Set-BqTable"></cmdlet-ref> work the same way as the <code>Get-</code> and 
+      <code>Set-</code> dataset cmdlets above.
+    </p>
+    <pre>
+# Creates a new table in the dataset from above.
+$table = $dataset | New-BqTable "logs2014" -Description "Log data from Jan 2014 to Dec 2014 inclusive"
```
quoctruong: Split this into 2 lines.

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,173 @@ mkdir my-new-bucket</pre>
...
+    <h2>TableRow</h2>
+    <p>
+      Data is added and removed from <code>Tables</code> in <em>Rows</em>. These rows are accessible using the 
+      <cmdlet-ref name="Add-BqTableRow"></cmdlet-ref> and <cmdlet-ref name="Get-BqTableRow"></cmdlet-ref>
+      cmdlets. <code>Add-BqTableRow</code> takes CSV, JSON, and AVRO files to import into BigQuery.
+    </p>
+    <pre>
+# Ingests a CSV file and appends its rows onto the table 'my_dataset:my_table'.
+$table = New-BqTable "logs2014" -DatasetId "page_views"
+$table | Add-BqTableRow CSV $filename -SkipLeadingRows 1 -WriteMode WriteAppend
```
quoctruong: Split this into 2 lines.

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,173 @@ mkdir my-new-bucket</pre>
...
+      request to the server to stop a certain job, and then returns immediately.
+      <blockquote>
+      Note on formatting table names within query strings: BigQuery format specifies that table names 
+      should be surrounded by backticks (`), but backticks are also Powershell's escape operators. 
+      Because of this, backticks must be escaped by adding a second backtick. See the example below.
+      </blockquote>
+    </p>
+    <pre>
+# Query Job: starts a query and outputs results into $table.
+Start-BqJob -Query "SELECT * FROM ``page_views:logs2014``" -Destination $table
```
quoctruong: Split into 2 lines.

js/app/controllers/content-product.ng
```diff
@@ -138,7 +138,173 @@ mkdir my-new-bucket</pre>
...
+    </pre>
+    <pre>
+# Load Job: adds TableRows to $table from the GCS file specified.
+$job = $dest_table | Start-BqJob -Load CSV "gs://page_views/server_logs_raw_2014.csv"
```
quoctruong: Split into 2 lines.

js/app/controllers/content-product.ng
ahandley: Got them all.  PTAL

PR #528, opened 2017-05-15
Fix bugs where tests reports too much information

quoctruong:
The main problem is we are not setting DISABLE_POWERSHELL_ANALYTICS before we import the module.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/528)


quoctruong:
Thanks!

PR #531, opened 2017-05-18
Fix Pester tests failure

quoctruong:
Also remove -OutputVersionFormat tests since we removed that parameter.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/531)


quoctruong:
Thanks!

PR #533, opened 2017-05-18
Add Remove-GkeNodePool cmdlet

quoctruong:
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/533)

### DIFF
--- a/Google.PowerShell/Container/GkeNodePoolCmdlets.cs
+++ b/Google.PowerShell/Container/GkeNodePoolCmdlets.cs
@@ -24,6 +24,9 @@ namespace Google.PowerShell.Container
     /// Gets Google Container Engine Node Pools from a Cluster. If -Project and/or -Zone parameter is not specified,
     /// the default project and/or the default zone will be used. If -NodePoolName parameter is not used,
     /// the cmdlet will return every node pools in the cluster.
+    /// You can either supply cluster name with -ClusterName or a Cluster object from
+    /// Get-GkeCluster with -ClusterObject. If a Cluster object is used, the cmdlet will use the
+    /// Project and Zone from the object.
     /// </para>
     /// <example>
     ///   <code>PS C:\> Get-GkeNodePool -ClusterName "my-cluster"</code>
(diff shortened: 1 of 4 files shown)


Google.PowerShell/Container/GkeNodePoolCmdlets.cs
```diff
@@ -626,15 +628,124 @@ protected override void ProcessRecord()
...
+        public string NodePoolName { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The name of the container cluster that the node pool is in.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, Position = 1)]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(Google.Apis.Container.v1.Data.Cluster),
+            Property = nameof(Google.Apis.Container.v1.Data.Cluster.Name))]
```
ILMTitan: I had not realized this earlier, but `Google.Apis.Container.v1.Data.Cluster` has a `Zone` property, and there for needs a separate parameter and parameter set (in all cmdlets that might take a `Cluster` object.
Oddly it does not have a project property. Can we pull that from the self-link, or will we have to supply it always as a separate parameter?

Google.PowerShell/Container/GkeNodePoolCmdlets.cs
```diff
@@ -626,15 +628,124 @@ protected override void ProcessRecord()
...
+        public string Zone { get; set; }
+
+        /// <summary>
+        /// <para type="description">
+        /// The name of the node pool to be removed.
+        /// </para>
+        /// </summary>
+        [Parameter(Mandatory = true, Position = 0)]
+        [PropertyByTypeTransformation(TypeToTransform = typeof(NodePool),
+            Property = nameof(Google.Apis.Container.v1.Data.NodePool.Name))]
```
ILMTitan: Can we pull the project, zone and cluster name from the NodePool self-link?

quoctruong:
Thanks!

PR #544, opened 2017-06-09
Add `isBeta` field to documentation JSON dump

chrsmith:
Added an `isBeta` field to the generated JSON dump of our PowerShell products and cmdlets.

In order to differentiate beta and non-beta cmdlets, we need to include that in our JSON dump for the website.
This PR simply adds that by:

- Hard-coding which products (GCE, BigQuery, etc.) contain only beta cmdlets.
- Reading the `BetaCmdlets.ps1` file to mark individual cmdlets.


Tools/GenerateWebsiteData.ps1
```diff
@@ -253,22 +263,31 @@ function Collapse-ParameterDescriptions($parameters) {
...
+    "Gcs"   = GenerateProductInfo      "Google Cloud Storage"    "google-cloud-storage"
+    "Gce"   = GenerateProductInfo      "Google Compute Engine"   "google-compute-engine"
+    "GcSql" = GenerateProductInfo      "Google Cloud SQL"        "google-cloud-sql"
+    "Gcd"   = GenerateProductInfo      "Google Cloud DNS"        "google-cloud-dns"
+    "Gcps"  = GenerateProductInfo      "Google Cloud PubSub"     "google-cloud-pubsub"
+    "GcLog" = GenerateProductInfo      "Google Cloud Logging"    "google-cloud-logging"
+    "GcIam" = GenerateProductInfo      "Google Cloud IAM"        "google-cloud-iam"
+    "GcpProject" = GenerateProductInfo "Google Cloud Project"    "google-cloud-project"
+    "Gke"   = GenerateProductInfo      "Google Container Engine" "google-cloud-container"    $true
+    "Bq"    = GenerateProductInfo      "Google Cloud BigQuery"   "google-cloud-bigquery"
```
quoctruong: Should we add $true to BigQuery, Logging, GcIam, GcpProject and Pubsub too since they are still in Beta?

chrsmith: If they are in beta, yes. Sorry for missing those in my first pass.

Given that 3/5ths of our PowerShell cmdlets are in "Beta", we probably want to be more deliberate in graduating things from Beta to GA. Would "zero bugs and three months of bake time" be sufficient? And if so, when would we mark IAM, Project, and PubSub as GA?

Soonish, right?

quoctruong: Yes, I plan to GA those cmdlets in the next release. I can remove those Beta labels when it is released.

PR #548, opened 2017-06-29
Adds the gs:() function.

ILMTitan:
This lets the user change the directory to gs:\ before loading the module (and therefore the provider).


chrsmith:
I assume this fixes [this SO post](https://stackoverflow.com/questions/44777593/cannot-find-path-because-it-does-not-exist)? Is exposing the "gs:" _function_ the right call here? Or is there a way to expose the PowerShell provider as a concept?

As for the code itself, assuming it is the right fix, it all looks reasonable. (But note the AppVeyor build failed.)

quoctruong:
LGTM. @chrsmith, I think this is the only way to fix it. The AppVeyor build is failing because we don't let the test run before the merge.

ILMTitan:
That SO post was the impetus, but I have been thinking about this for a while. There already exists a "c:" function that does the same thing.

PR #551, opened 2017-07-06
Add -MaxBadRecords parameter

quoctruong:
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/551)

### DIFF
--- a/Google.PowerShell/BigQuery/BqJob.cs
+++ b/Google.PowerShell/BigQuery/BqJob.cs
@@ -469,6 +469,17 @@ namespace Google.PowerShell.BigQuery
         [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.DoLoad)]
         public SwitchParameter AllowQuotedNewlines { get; set; }
 
+        /// <summary>
+        /// <para type="description">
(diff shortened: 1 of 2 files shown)


Google.PowerShell/BigQuery/BqJob.cs
```diff
@@ -656,7 +667,8 @@ public Apis.Bigquery.v2.Data.Job DoLoad()
                                 IgnoreUnknownValues = AllowUnknownFields,
                                 AllowJaggedRows = AllowJaggedRows,
                                 AllowQuotedNewlines = AllowQuotedNewlines,
-                                WriteDisposition = WriteMode.ToString()
+                                WriteDisposition = WriteMode.ToString(),
+                                MaxBadRecords = MaxBadRecords ?? 0
```
iantalarico: Is 0 the API default? maybe only add it if it's set to be safe?

quoctruong: Yes, it is the default API value. https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs#configuration.load

quoctruong:
Thanks!

PR #558, opened 2017-10-05
Fix tests and update log cmdlets

quoctruong:
There are some changes in the Logging API so we have to change some of the logging cmdlets.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/558)

### DIFF
--- a/Google.PowerShell/Logging/GcLogSinkCmdlets.cs
+++ b/Google.PowerShell/Logging/GcLogSinkCmdlets.cs
@@ -144,6 +144,10 @@ namespace Google.PowerShell.Logging
         [ValidateNotNullOrEmpty]
         public string SinkName { get; set; }
 
+        // Don't expose After and Before as parameters since they are deprecated.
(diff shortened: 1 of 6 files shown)


Google.PowerShell/Logging/GcLogSinkCmdlets.cs
```diff
@@ -221,16 +225,6 @@ protected override void ProcessRecord()
                 after: null,
                 otherFilter: Filter);
 
-            if (Before.HasValue)
-            {
-                logSink.EndTime = XmlConvert.ToString(Before.Value, XmlDateTimeSerializationMode.Local);
```
ILMTitan: I feel like we should have at least one release where we deprecate these parameters by writing a warning about the parameters.

quoctruong:
@ILMTitan PTAL, I added warnings for the parameters.

quoctruong:
Thanks!

PR #561, opened 2017-10-28
Fix .Net Core build.

ILMTitan:
The Dotnet core project was failing for me because of NuGet package issues. Add a new NuGet.Config at the solution root. Also, change the way the project imports the files to be less verbose.

### DIFF
--- a/Google.PowerShell.DotnetCore/Google.PowerShell.DotnetCore.csproj
+++ b/Google.PowerShell.DotnetCore/Google.PowerShell.DotnetCore.csproj
@@ -21,7 +21,7 @@
   </PropertyGroup>
 
   <ItemGroup>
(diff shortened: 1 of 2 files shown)


Google.PowerShell.DotnetCore/Google.PowerShell.DotnetCore.csproj
```diff
@@ -21,7 +21,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <Compile Include="..\Google.PowerShell\Common\*.cs;..\Google.PowerShell\Compute\*.cs;..\Google.PowerShell\Dns\*.cs;..\Google.PowerShell\Properties\*.cs;..\Google.PowerShell\Provider\*.cs;..\Google.PowerShell\Sql\*.cs;..\Google.PowerShell\Storage\*.cs;..\Google.PowerShell\Logging\*.cs;..\Google.PowerShell\PubSub\*.cs;..\Google.PowerShell\BigQuery\*.cs" Exclude="bin\**;obj\**;**\*.xproj;packages\**" />
+    <Compile Include="..\Google.PowerShell\**\*.cs" Exclude="..\**\bin\**;..\**\obj\**;**\*.xproj;packages\**;..\Google.PowerShell\CloudResourceManager\IamPolicyCmdlets.cs;..\Google.PowerShell\Container\*" />
```
quoctruong: Does ..\Google.PowerShell\**\*.cs cover the CloudResourceManager and Container too? If so, then I think we can remove those too.

ILMTitan: The exclude excludes Container and the one cmdlet set that is problematic in CloudResourceManager.

PR #563, opened 2017-10-30
Update nuget package

quoctruong:
Also include `IamPolicyCmdlets` and `Container` cmdlets to .NET Core.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/563)

### DIFF
--- a/Google.PowerShell/Container/GkeNodePoolCmdlets.cs
+++ b/Google.PowerShell/Container/GkeNodePoolCmdlets.cs
@@ -998,7 +998,7 @@ namespace Google.PowerShell.Container
                     Uri nodePoolLink;
                     NodePoolName = NodePoolObject.Name;
                     if (Uri.TryCreate(NodePoolObject.SelfLink, UriKind.Absolute, out nodePoolLink)
-                        && nodePoolLink.Scheme == Uri.UriSchemeHttps)
+                        && nodePoolLink.Scheme == "https")
                     {
                         Project = GetProjectNameFromUri(NodePoolObject.SelfLink);
                         Zone = GetUriPart("zones", NodePoolObject.SelfLink);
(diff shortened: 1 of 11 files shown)


Google.PowerShell/Container/GkeNodePoolCmdlets.cs
```diff
@@ -998,7 +998,7 @@ private void ProcessParams()
                     Uri nodePoolLink;
                     NodePoolName = NodePoolObject.Name;
                     if (Uri.TryCreate(NodePoolObject.SelfLink, UriKind.Absolute, out nodePoolLink)
-                        && nodePoolLink.Scheme == Uri.UriSchemeHttps)
+                        && nodePoolLink.Scheme == "https")
```
iantalarico: Is there some way to still use a const here?

Google.PowerShell/Provider/GoogleCloudStorageProvider.cs
```diff
@@ -1086,7 +1086,8 @@ private IEnumerable<Object> ListChildren(GcsPath gcsPath, bool recurse, bool all
 
         private static IEnumerable<Project> ListAllProjects()
         {
-            ProjectsResource.ListRequest request = ResourceService.Projects.List();
+            Google.Apis.CloudResourceManager.v1.ProjectsResource.ListRequest request =
```
iantalarico: Why the full namespace here?

ILMTitan: Because `Google.Apis.Storage.v1` now also has a `ProjectsResource`.

Google.PowerShell/Container/GkeNodePoolCmdlets.cs
ILMTitan: Looks to me like no, unless we define one ourselves.

quoctruong:
Thanks!

PR #566, opened 2017-11-02
Add "-quiet" parameter to Import-Module -ArgumentList

### LINKED ISSUE #562: Add "-quiet" parameter for importing the module.
The `Import-Module` cmdlet can take an `-ArgumentList` that is passed to the module script. We should make use of that to include a second way to install the _Cloud SDK_ without prompts (the first being `$env:GCLOUD_SDK_INSTALLATION_NO_PROMPT=$true`).

ILMTitan:
Fixes #562

### DIFF
--- a/Google.PowerShell/ReleaseFiles/GoogleCloud.psm1
+++ b/Google.PowerShell/ReleaseFiles/GoogleCloud.psm1
@@ -1,4 +1,6 @@
-﻿$script:GCloudModule = $ExecutionContext.SessionState.Module
+﻿param()
+
+$script:GCloudModule = $ExecutionContext.SessionState.Module
 $script:GCloudModulePath = $script:GCloudModule.ModuleBase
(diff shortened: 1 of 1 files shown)


Google.PowerShell/ReleaseFiles/GoogleCloud.psm1
```diff
@@ -1,4 +1,6 @@
-﻿$script:GCloudModule = $ExecutionContext.SessionState.Module
+﻿param()
```
quoctruong: Should we define an argument list in the param block? Something like [this](https://stackoverflow.com/questions/36897511/powershell-module-pass-a-parameter-while-importing-module)

ILMTitan: Do we really want any additional parameters to be ordered? If we do, that is fine, but if we don't, what I have written is orderless.

PR #571, opened 2017-11-08
Change tooling to package .NET Core module

quoctruong:
After this change, the `.\BuildAndPackage.ps1` script will package both the Core CLR and normal CLR bits into one module. The `GoogleCloud.psm1` file will pick which assembly to load based on the PowerShell runtime. This way, our module will work across the different platforms.

I also added logic to install Cloud SDK on Unix systems.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/571)

### DIFF
--- a/Google.PowerShell/Common/CloudSdkSettings.cs
+++ b/Google.PowerShell/Common/CloudSdkSettings.cs
@@ -1,13 +1,12 @@
 ﻿// Copyright 2015-2016 Google Inc. All Rights Reserved.
 // Licensed under the Apache License Version 2.0.
 
-using Newtonsoft.Json.Linq;
 using System;
-using System.Collections;
-using System.Collections.Generic;
-using System.Diagnostics;
 using System.IO;
 using System.Linq;
+#if CORECLR
+using System.Runtime.InteropServices;
+#endif
 
 namespace Google.PowerShell.Common
 {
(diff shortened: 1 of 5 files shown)


Google.PowerShell/Common/CloudSdkSettings.cs
```diff
@@ -141,15 +140,41 @@ public static string GetAnonymousClientID()
         }
 
         /// <summary>
+        /// True if the module is run on Windows. This is a cache
+        /// for IsWindows property.
+        /// </summary>
+        private static bool? s_isWindows;
```
ILMTitan: `Lazy<bool>` would capture your intent here better than `bool?`.

Google.PowerShell/ReleaseFiles/GoogleCloud.psm1
```diff
@@ -50,10 +62,10 @@ function Install-GCloudSdk {
     }
 
     Write-Host "Google Cloud SDK is not found in PATH. The SDK is required to run the module."
-    $noPrompt = $env:GCLOUD_SDK_INSTALLATION_NO_PROMPT -eq $true -or $args -match "-?quiet"
```
ILMTitan: Is this a merge error, or did you not like this feature?

quoctruong: It's a merge error, I'll add it back.

quoctruong:
Thanks!

PR #576, opened 2017-11-21
Add unit tests for custom cpu & memory parameters.

ILMTitan:
Fix the slow CacheItem tests.


ILMTitan:
The tests are written assuming that `CustomMemory` and `CustomCpu`, and `MachineType` are in different parameter sets.

PR #579, opened 2017-12-12
1.0.0.2 release

quoctruong:
Fixes a bug in the release tool where the cmdlets are not exported properly. Also fixes a bug where we are not copying all the necessary assemblies for CoreClr (which causes an issue for `BigQuery` cmdlets).

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/579)


quoctruong:
Thanks!

Issue #582, opened 2017-12-13
WriteMode flag not implemented for Query Jobs in Start-BqJob


adarobin:
It would be useful if Start-BqJob implemented the WriteMode flag for Query Jobs.  I'd like to be able to pass ```-WriteMode WriteTruncate``` to some of my Queries.

quoctruong:
Hi @adarobin, can you clarify on the requirements? I thought the `Start-BqJob` command has `-WriteMode WriteTruncate`?

adarobin:
Hi @quoctruong, The ```Start-BqJob``` has ```-WriteMode WriteTruncate``` but it is only available in the parameter set for Copy Jobs, not Query Jobs.

```PS /Users/adarobin> Start-BqJob -Query -QueryString "SELECT * FROM ``foo``" -Destination $destTable -WriteMode WriteTruncate                                    
Start-BqJob : Parameter set cannot be resolved using the specified named parameters.
At line:1 char:1
+ Start-BqJob -Query -QueryString "SELECT * FROM ``foo``" -Destination  ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+ CategoryInfo          : InvalidArgument: (:) [Start-BqJob], ParameterBindingException
+ FullyQualifiedErrorId : AmbiguousParameterSet,Google.PowerShell.BigQuery.StartBqJob
```

quoctruong:
@adarobin We've implemented the feature. It will be out in the next time the module is released!

quoctruong:
@adarobin Version 1.0.1.3 of the module (which is available on the PowerShellGallery) has this fix.

PR #583, opened 2017-12-13
Add -WriteMode to Start-BqJob -Query

quoctruong:
This is to support https://github.com/GoogleCloudPlatform/google-cloud-powershell/issues/582

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/583)

### DIFF
--- a/Google.PowerShell/BigQuery/BqCmdlet.cs
+++ b/Google.PowerShell/BigQuery/BqCmdlet.cs
@@ -19,10 +19,15 @@ namespace Google.PowerShell.BigQuery
         public BigqueryService Service => _service.Value;
         public BigQueryClient Client => _client.Value;
 
+        internal static BigqueryService DefaultBigQueryService { private get; set; }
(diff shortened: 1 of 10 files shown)


Google.PowerShell/BigQuery/BqCmdlet.cs
```diff
@@ -19,10 +19,15 @@ public class BqCmdlet : GCloudCmdlet
         public BigqueryService Service => _service.Value;
         public BigQueryClient Client => _client.Value;
 
+        internal static BigqueryService DefaultBigQueryService { private get; set; }
```
iantalarico: The naming on these seems a little odd to me.  I think default normally implies if nothing else is provided then use the default. 

These can't ever be overridden by a user using the cmdlts right?  Maybe just have the old ones be the default and override them?

quoctruong:
@iantalarico PTAL.

quoctruong:
Thanks!

Issue #586, opened 2018-01-19
Add-GceInstance : ErrorAction preference not taking effect


rajaprad:
**Module** - 1.0.1.2

**Command**

Add-GceInstance -Project "myproject" -Zone "us-east1-b" -Region "us-east1-b" -Name "p
radebbanvm" -MachineType f1-micro **-ErrorAction SilentlyContinue**

**Error**

Add-GceInstance : Google.Apis.Requests.RequestError
The resource 'projects/myproject/zones/us-east1-b/instances/pradebbanvm' already exists [409]
Errors [
        Message[The resource 'projects/myproject/zones/us-east1-b/instances/pradebbanvm' already exists]
Location[ - ] Reason[alreadyExists] Domain[global]
]
At line:1 char:1
+ Add-GceInstance -Project myproject -Zone "us-east1-b" -R ...
+ ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: (:) [Add-GceInstance], GoogleApiException
    + FullyQualifiedErrorId : Google.GoogleApiException,Google.PowerShell.ComputeEngine.AddGceInstanceCmdlet

**Issue**
-ErrorAction SilentlyContinue does not suppress the error!

rajaprad:
How can this be closer without a fix, and just by copying to another duplicate item. How would there be a ownership of bugs created by a contributor in public GitHub. Wakeup Google. Will take this if the closed issues are not re-opened!

quoctruong:
@rajaprad The issue got automatically closed when I closed the pull request that is going to fix the bug. In any case this is now fixed and will be available in the next release of the module. I will reopen this issue and close it when that happens.

rajaprad:
Am not getting the idea behind leaving it closed when the fix is not deployed to prod! Why should it be closed before it's released?

quoctruong:
@rajaprad the fix is now available in version 1.0.1.14 of the module.

PR #588, opened 2018-01-19
Make add-gceinstance and add-gcemanagedinstancegroup throws non-terminating error

quoctruong:
This is to fix https://github.com/GoogleCloudPlatform/google-cloud-powershell/issues/586.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/588)

### DIFF
--- a/Google.PowerShell.Tests/Compute/AddGceInstanceCmdletTests.cs
+++ b/Google.PowerShell.Tests/Compute/AddGceInstanceCmdletTests.cs
@@ -77,14 +77,7 @@ namespace Google.PowerShell.Tests.Compute
 
             // An error should be thrown (if it is a terminating error,
             // we wouldn't even reach this point).
-            Assert.AreEqual(Pipeline.Error.Count, 1);
-
-            // Verify that the call is made.
-            instances.Verify(
-                resource => resource.Insert(
-                    It.IsAny<Instance>(),
-                    FakeProjectId, FakeZoneName),
-                Times.Once);
+            TestErrorRecord(ErrorCategory.ResourceExists);
         }
 
         /// <summary>
@@ -109,14 +102,7 @@ namespace Google.PowerShell.Tests.Compute
 
             // An error should be thrown (if it is a terminating error,
             // we wouldn't even reach this point).
-            Assert.AreEqual(Pipeline.Error.Count, 1);
-
-            // Verify that the call is made.
-            instances.Verify(
-                resource => resource.Insert(
-                    It.IsAny<InstanceGroupManager>(),
-                    FakeProjectId, FakeZoneName),
-                Times.Once);
+            TestErrorRecord(ErrorCategory.ResourceExists);
         }
 
         [Test]
--- a/Google.PowerShell.Tests/Compute/AddGceInstanceCmdletTests.cs
+++ b/Google.PowerShell.Tests/Compute/AddGceInstanceCmdletTests.cs
@@ -20,6 +20,7 @@ using NUnit.Framework;
 using System.Collections.ObjectModel;
 using System.Linq;
 using System.Management.Automation;
+using System.Net;
 
 namespace Google.PowerShell.Tests.Compute
 {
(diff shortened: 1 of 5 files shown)


iantalarico:
I'm a little confused here (probably due to my lack of PS knowledge).  Why are we only catching conflicts here?  What is the expected result of 'SilentlyContinue'?  Does this mean any failure results in non termination?

quoctruong:
@iantalarico There are 2 types of errors here: terminating and non-terminating error. 'SilentlyContinue' can only suppress non-terminating error. For some cmdlets, I think it makes sense in some common scenarios to write a non-terminating error (like the instance that we are trying to create already exists) instead of throwing the terminating error. This way, the user can suppress it.

Google.PowerShell/Compute/GceInstanceCmdlets.cs
```diff
@@ -630,11 +631,21 @@ protected override void ProcessRecord()
...
             {
-                WriteObject(Service.Instances.Get(Project, Zone, instance.Name).Execute());
-            });
+                Operation operation = request.Execute();
+                AddZoneOperation(Project, Zone, operation, () =>
+                {
+                    WriteObject(Service.Instances.Get(Project, Zone, instance.Name).Execute());
+                });
+            }
+            catch (GoogleApiException apiEx) when (apiEx.HttpStatusCode == HttpStatusCode.Conflict)
```
iantalarico: Should this also be based on 'SilentlyContinue'?

quoctruong: @iantalarico What do you mean by that?

ILMTitan: @iantalarico This is silently continuing now. `WriteResourceExistsError(...)` writes an error record to the error stream (a suppressible non-terminating error). If we let the exception bubble up, the error would be a terminating error.

iantalarico: Do we also need to check 

... when(apiEx.HttpStatusCode == HttpStatusCode.Conflict && SilentlyContinue)?

quoctruong: @iantalarico We don't need to check for the `SilentlyContinue` parameter. It's a built-in PowerShell parameter that will suppress error record (which is written out by the cmdlet in the catch clause).

iantalarico: Ah I see, we just write the error and let PS handle it.  SGTM

Google.PowerShell.Tests/Compute/AddGceInstanceCmdletTests.cs
```diff
@@ -52,6 +53,72 @@ public void TestAcceptingCustomMemoryCustomCpu()
...
+            Pipeline.Commands.AddScript(
+                $"Add-GceInstance -Name instance-name -Disk ${psDiskVar}");
+            Collection<PSObject> results = Pipeline.Invoke();
+
+            // An error should be thrown (if it is a terminating error,
+            // we wouldn't even reach this point).
+            Assert.AreEqual(Pipeline.Error.Count, 1);
+
+            // Verify that the call is made.
+            instances.Verify(
```
ILMTitan: I think this verification is unnecessary. The error in the error pipeline tells us we made the correct call.

Google.PowerShell.Tests/Compute/AddGceInstanceCmdletTests.cs
```diff
@@ -52,6 +53,72 @@ public void TestAcceptingCustomMemoryCustomCpu()
...
+                i => i.Insert(It.IsAny<Instance>(), It.IsAny<string>(), It.IsAny<string>()),
+                apiException);
+
+            Pipeline.Commands.AddScript(
+                $"Add-GceInstance -Name instance-name -Disk ${psDiskVar}");
+            Collection<PSObject> results = Pipeline.Invoke();
+
+            // An error should be thrown (if it is a terminating error,
+            // we wouldn't even reach this point).
+            Assert.AreEqual(Pipeline.Error.Count, 1);
```
ILMTitan: Do we want to test something about the error? Perhaps test `Pipeline.Error[0].ErrorCategory == ErrorCategory.ResourceExists`?

Google.PowerShell.Tests/Compute/AddGceInstanceCmdletTests.cs
```diff
@@ -52,6 +53,72 @@ public void TestAcceptingCustomMemoryCustomCpu()
...
+                apiException);
+
+            Pipeline.Commands.AddScript(
+                $"Add-GceManagedInstanceGroup -Name instance-group -InstanceTemplate template -TargetSize 1");
+            Collection<PSObject> results = Pipeline.Invoke();
+
+            // An error should be thrown (if it is a terminating error,
+            // we wouldn't even reach this point).
+            Assert.AreEqual(Pipeline.Error.Count, 1);
+
```
ILMTitan: Test the error?

Google.PowerShell.Tests/Compute/AddGceInstanceCmdletTests.cs
```diff
@@ -52,6 +53,72 @@ public void TestAcceptingCustomMemoryCustomCpu()
...
+            // Verify that the call is made.
+            instances.Verify(
```
ILMTitan: Again, I don't think this is necessary.

quoctruong:
Thanks! Fixed the tests.

PR #595, opened 2018-02-01
Ensure bucketCollections.CompleteAdding() is always called.

### LINKED ISSUE #594: Browsing storage buckets in gs: causes PowerShell to crash
I am using version 1.0.1.4 (I also saw this with 1.0.0.9).

Repo steps:
1. Open PowerShell and 'cd gs:/'
2. Run a few commands.
3.  Let the terminal sit for a minute or two and the terminal will close unexpectedly.

ILMTitan:
Fixes #594.

### DIFF
--- a/Google.PowerShell.Tests/Common/GoogleApiMockExtensions.cs
+++ b/Google.PowerShell.Tests/Common/GoogleApiMockExtensions.cs
@@ -16,6 +16,7 @@ using Google.Apis.Http;
 using Google.Apis.Requests;
 using Google.Apis.Services;
 using Moq;
+using Moq.Language.Flow;
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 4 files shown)


Google.PowerShell.Tests/Provider/GoogleCloudStorageProviderTests.cs
```diff
@@ -92,5 +102,27 @@ public void TestIssue580()
...
             Assert.AreEqual(bucketName, returnedBucket.Id);
         }
+
+        /// <summary>
+        /// Test the fix for an issue where the GcsProvider would hang.
+        /// </summary>
+        /// <seealso href="https://github.com/GoogleCloudPlatform/google-cloud-powershell/issues/594"/>
+        [Test]
+        [Timeout(200)]
+        public void TestIssue594()
```
quoctruong: nit: I think we should have a descriptive name like `TestListProjectsThrownError` instead of `TestIssue594`.

ILMTitan: Yea, that was me being lazy. Fixed.

quoctruong:
@ILMTitan LGTM.

Google.PowerShell.Tests/Common/GoogleApiMockExtensions.cs
```diff
@@ -140,17 +140,62 @@ public static class GoogleApiMockExtensions
...
+        /// projects.SetupRequestError&lt;ProjectsResource, ProjectsResource.ListRequest, ListProjectsResponse&gt;(
+        ///         p =&gt; o.Get(bucketName, objectName),
+        ///         new Exception("error-message"));
+        /// </code>
+        /// <typeparam name="TResource">The type of resource creating the request.</typeparam>
+        /// <typeparam name="TRequest">The type of the request to make.</typeparam>
+        /// <param name="resourceMock">The mock of the resource that makes the request.</param>
+        /// <param name="requestExpression">The expression of the request. Uses Moq.It functions for wildcards.</param>
+        /// <returns>The mock of the request object. Useful for verification.</returns>
+
```
iantalarico: super nit: extra new line

quoctruong:
@iantalarico I think that sounds good. I think sometimes I just get a little too "greedy" with what I can put in the PR. If we separate them, then it will also be easier if we want to revert a bad PR without reverting the formatting PR.

PR #596, opened 2018-02-01
Fix a bug where BootstrapCloudToolsForPowershell is not found by Cloud Tools For PowerShell app

quoctruong:
The issue is that the "Cloud Tools For PowerShell" application is looking for "BootstrapCloudToolsForPowerShell.ps1" at the wrong place. Also, the bootstrap script itself is importing the module with the wrong directory.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/596)

### DIFF
--- a/Tools/BuildAndPackage.ps1
+++ b/Tools/BuildAndPackage.ps1
@@ -249,11 +249,16 @@ function ConfirmExists($relativePath) {
    }
 }
 
+# Move BootstrapCloudToolsForPowerShell.ps1 to the GoogleCloud folder as this is
+# where "Cloud Tools For PowerShell" program in the Cloud SDK is looking for it.
+Move-Item "$googleCloudDir\BootstrapCloudToolsForPowerShell.ps1" `
+          "$powerShellDir\GoogleCloud\BootstrapCloudToolsForPowerShell.ps1"
+
 ConfirmExists "PowerShell\GoogleCloud\$normalizedVersion\GoogleCloud.psd1"
 ConfirmExists "PowerShell\GoogleCloud\$normalizedVersion\fullclr\Google.PowerShell.dll"
 ConfirmExists "PowerShell\GoogleCloud\$normalizedVersion\coreclr\Google.PowerShell.dll"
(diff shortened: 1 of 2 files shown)


Tools/BuildAndPackage.ps1
```diff
@@ -249,11 +249,16 @@ function ConfirmExists($relativePath) {
...
+# where "Cloud Tools For PowerShell" program in the Cloud SDK is looking for it.
+Move-Item "$googleCloudDir\BootstrapCloudToolsForPowerShell.ps1" `
+          "$powerShellDir\GoogleCloud\BootstrapCloudToolsForPowerShell.ps1"
+
 ConfirmExists "PowerShell\GoogleCloud\$normalizedVersion\GoogleCloud.psd1"
 ConfirmExists "PowerShell\GoogleCloud\$normalizedVersion\fullclr\Google.PowerShell.dll"
 ConfirmExists "PowerShell\GoogleCloud\$normalizedVersion\coreclr\Google.PowerShell.dll"
 ConfirmExists "PowerShell\GoogleCloud\$normalizedVersion\GoogleCloud.psm1"
-ConfirmExists "PowerShell\GoogleCloud\$normalizedVersion\BootstrapCloudToolsForPowerShell.ps1"
+ConfirmExists "PowerShell\GoogleCloud\BootstrapCloudToolsForPowerShell.ps1"
```
iantalarico: Should this be \GoogleCloud\$normalizedVersion\?

Is there a change something with this changes between versions?

quoctruong: Can you clarify the second question?

iantalarico: Can BootstrapCloudToolsForPowerShell.ps1 change between v1 1.0.0.4 and 1.0.0.5?  If so should we version this too?

quoctruong: Oh I see. The BootstrapCloudToolsForPowerShell.ps1 won't change. Its only job is to launch the module.

quoctruong:
Thanks!

PR #600, opened 2018-02-13
Add -Region option to Add-GceManagedInstane cmdlets

quoctruong:
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/600)

### DIFF
--- a/Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
+++ b/Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
@@ -287,7 +287,7 @@ namespace Google.PowerShell.Tests.Compute
                 $"{FakeRegionName} -Zone {FakeZoneName}");
             var error = Assert.Throws<PSInvalidOperationException>(() => Pipeline.Invoke());
 
-            Assert.AreEqual("Parameters -Region and -Zone cannot be used together.",
+            Assert.AreEqual("Parameters -Region and -Zone cannot be used together with -Object.",
                 error.Message);
         }
 
--- a/Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
+++ b/Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
@@ -31,7 +31,8 @@ namespace Google.PowerShell.Tests.Compute
     {
         private InstanceGroupManager FirstTestGroup = new InstanceGroupManager()
         {
-            Name = "One"
+            Name = "One",
+            Region = $@"{ComputeHttpsLink}\region\us-central1"
         };
         private InstanceGroupManager SecondTestGroup = new InstanceGroupManager()
         {
--- a/Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
+++ b/Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
@@ -272,6 +272,26 @@ namespace Google.PowerShell.Tests.Compute
         }
 
         /// <summary>
+        /// Tests that Add-GceManagedInstanceGroup throws error if -Object
+        /// is used with both -Region and -Zone.
+        /// </summary>
+        [Test]
+        public void TestAddGceManagedInstanceGroupByObjectError()
+        {
+            string managedRegionObject = "managedRegionObj";
+            string instanceGroupName = FirstTestGroup.Name;
+            Pipeline.Runspace.SessionStateProxy.SetVariable(managedRegionObject, FirstTestGroup);
+
+            Pipeline.Commands.AddScript(
+                $"${managedRegionObject} | Add-GceManagedInstanceGroup -Region " +
+                $"{FakeRegionName} -Zone {FakeZoneName}");
+            var error = Assert.Throws<PSInvalidOperationException>(() => Pipeline.Invoke());
+
+            Assert.AreEqual("Parameters -Region and -Zone cannot be used together.",
+                error.Message);
+        }
+
+        /// <summary>
         /// Tests that Add-GceManagedInstanceGroup uses -Zone option by default.
         /// </summary>
         [Test]
--- a/Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
+++ b/Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
@@ -90,19 +90,29 @@ namespace Google.PowerShell.Tests.Compute
         }
 
         /// <summary>
+        /// Creates a InstanceGroupManager with a region and self-link.
+        /// </summary>
+        private InstanceGroupManager CreateRegionalInstanceGroup(
+            string groupName, string projectName, string regionName)
+        {
+            string regionLink = $"{ComputeHttpsLink}/projects/{projectName}/regions/{regionName}";
+            return new InstanceGroupManager()
+            {
+                Name = groupName,
+                Region = regionLink,
+                SelfLink = $"{regionLink}/instanceGroupManagers/{groupName}"
+            };
+        }
+
+        /// <summary>
         /// Tests that Remove-GceManagedInstanceGroup works when pipelining regional instance group.
         /// </summary>
         [Test]
         public void TestRemoveGceManagedInstanceGroupPipelineRegional()
         {
             string instanceGroupName = "RegionalInstanceGroup";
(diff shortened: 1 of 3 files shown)


Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
```diff
@@ -148,5 +148,98 @@ public void TestRemoveGceManagedInstanceGroupPipelineZonal()
 
             instances.VerifyAll();
         }
+
+        /// <summary>
+        /// Runs PowerShell script script to create a GceManagedInstanceGroup.
+        /// Checks that the group created has name instanceGroupName.
+        /// </summary>
+        /// <param name="script"></param>
+        private void TestAddGceManagedInstanceGroupHelper(string script, string instanceGroupName)
```
ILMTitan: The name of this method doesn't say what it does, only where it is used. Part of the problem is it does not do one thing; it does two things that happen to be repeated and always next to each other.

The last three lines could conceivably be a good helper function named `AssertSingleGroupReturned`, but I'm not sure the logic is complicated enough to warrant it. I would err on the side of verbosity here and inline this function.

See http://go/unit-test-practices#assertions

Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
```diff
@@ -148,5 +148,98 @@ public void TestRemoveGceManagedInstanceGroupPipelineZonal()
...
+            instances.SetupRequest(
+                  item => item.Insert(
+                      It.Is<InstanceGroupManager>(manager => manager.Name == instanceGroupName),
+                      FakeProjectId, FakeRegionName),
+                  DoneOperation);
+            instances.SetupRequest(
+                i => i.Get(FakeProjectId, FakeRegionName, instanceGroupName),
+                new InstanceGroupManager { Name = instanceGroupName });
+
+            TestAddGceManagedInstanceGroupHelper(
```
ILMTitan: You can see here how the helper function is blurring the act and assert blocks of this test.

Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
```diff
@@ -148,5 +148,98 @@ public void TestRemoveGceManagedInstanceGroupPipelineZonal()
...
+                $"Add-GceManagedInstanceGroup -Name {instanceGroupName} " +
+                $"-InstanceTemplate {templateName} -TargetSize 1 -Region {FakeRegionName}",
+                instanceGroupName);
+            instances.VerifyAll();
```
ILMTitan: You only need to verify the `Insert` here. The `Get` is verified by asserting the returned object is correct.

Google.PowerShell/Compute/GceManagedInstanceGroupCmdlets.cs
```diff
@@ -393,40 +418,82 @@ protected override void ProcessRecord()
...
-                case ParameterSetNames.ByObject:
+                case ParameterSetNames.ByRegionObject:
+                case ParameterSetNames.ByZoneObject:
                     manager = Object;
                     break;
                 default:
                     throw UnknownParameterSetException;
             }
+
+            bool isRegionInstanceGroup = ParameterSetName == ParameterSetNames.ByRegionObject
```
ILMTitan: What happens if the input object already has the Region or Zone (or even Project) fields filled out? Do they get overwritten? It looks like a bug that already existed, but is exacerbated by these changes.

quoctruong: Yes, they will get overwritten. I implemented this way because I think there's a use case where the user has an instance group in region A and he wants to make a copy of that in region B, so he would get the existing instance group and pipe it to the cmdlet and just supply `-Region B`.

quoctruong: Maybe a better solution will be to get the existing Region or Zone in the object and overwrite it if the user supply something via `-Region`, `-Zone` or `-Project`.

quoctruong:
@ILMTitan PTAL. I modify the logic when pipelining the object so that:
1. If -Region is supplied, we will create a Regional Instance Group.
2. Otherwise, create a Zonal Instance Group. If the user supplies either -Project or -Zone parameters, we will use them instead of the default value from the SDK.

I have to add logic for this instead of using ParameterSet because the ParameterSet will be ambiguous otherwise.

Google.PowerShell.Tests/Compute/ManagedGceInstanceCmdletTests.cs
```diff
@@ -148,5 +158,150 @@ public void TestRemoveGceManagedInstanceGroupPipelineZonal()
...
+        }
+
+        /// <summary>
+        /// Tests that Add-GceManagedInstanceGroup works with -Region and -Object.
+        /// </summary>
+        [Test]
+        public void TestAddGceManagedInstanceGroupByObjectWithRegionParam()
+        {
+            string managedRegionObject = "managedRegionObj";
+            string instanceGroupName = FirstTestGroup.Name;
```
ILMTitan: This test will fail if you add this line:

    FirstTestGroup.Region = $@"{ComputeHttpsLink}\region\us-central1";

If we explicitly set a target region or zone, we should not overwrite them.

Google.PowerShell/Compute/GceManagedInstanceGroupCmdlets.cs
```diff
@@ -408,25 +434,90 @@ protected override void ProcessRecord()
...
+            }
+        }
+
+        /// <summary>
+        /// Processes the arguments when the cmdlet receives a piped ManagedInstanceGroup.
+        /// </summary>
+        private void ProcessPipedObjectArguments()
+        {
+            // Only use argument from -Project if the user supplies it.
+            if (MyInvocation.BoundParameters.ContainsKey(nameof(Project))
```
ILMTitan: Should these by `if (!MyInvocation...)`?

Google.PowerShell/Compute/GceManagedInstanceGroupCmdlets.cs
```diff
@@ -408,25 +434,90 @@ protected override void ProcessRecord()
                     break;
                 case ParameterSetNames.ByObject:
                     manager = Object;
+                    // Uses Zone by default unless user explicitly uses -Region.
+                    isZoneInstanceGroup = !MyInvocation.BoundParameters.ContainsKey(nameof(Region));
```
ILMTitan: Should this be `!string.isNullOrEmpty(Zone) && !MyInvocation.BoundParameters.ContainsKey(nameof(Region));`?
Should it also be after the `ProcessPipedObjectArguments()` call?

ILMTitan:
I think the logic should be:
0. If the user specifies both `-Zone` and `-Region`, throw an error.
1. If the user specifies `-Region`, create a Region Instance Group there.
2. If the user specifies `-Zone`, create a Zone Instance Group there.
3. If the user supplies an object that specifies either Zone or Region, create that type of instance group.
4. Else create a Zone Instance Group at the default zone.

quoctruong:
@ILMTitan Thanks for the great suggestion! PTAL.

Google.PowerShell/Compute/GceManagedInstanceGroupCmdlets.cs
```diff
@@ -454,27 +452,52 @@ protected override void ProcessRecord()
...
-            // Only use argument from -Project if the user supplies it.
-            if (MyInvocation.BoundParameters.ContainsKey(nameof(Project))
-                && !string.IsNullOrWhiteSpace(Object.SelfLink))
+            bool regionSpecified = MyInvocation.BoundParameters.ContainsKey(nameof(Region));
+            bool projectSpecified = MyInvocation.BoundParameters.ContainsKey(nameof(Project));
+            bool zoneSpecified = MyInvocation.BoundParameters.ContainsKey(nameof(Zone));
+
+            if (regionSpecified && zoneSpecified)
+            {
+                throw UnknownParameterSetException;
```
ILMTitan: I believe this error will be confusing to the user. It will complain of an unknown parameter set "ByObject", when such a set clearly exists. A more specific exception message is needed.

quoctruong:
@ILMTitan PTAL.

quoctruong:
Thanks!

PR #604, opened 2018-02-21
Add -Region parameter to Wait-GceManagedInstanceGroup

quoctruong:
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/604)

### DIFF
--- a/Google.PowerShell/Compute/GceManagedInstanceGroupCmdlets.cs
+++ b/Google.PowerShell/Compute/GceManagedInstanceGroupCmdlets.cs
@@ -1248,13 +1248,15 @@ namespace Google.PowerShell.Compute
     /// [Managed Instance Group resource definition]
     /// </para>
     /// </summary>
-    [Cmdlet(VerbsLifecycle.Wait, "GceManagedInstanceGroup")]
+    [Cmdlet(VerbsLifecycle.Wait, "GceManagedInstanceGroup",
+        DefaultParameterSetName = ParameterSetNames.ByNameZone)]
     public class WaitGceManagedInstanceGroupCmdlet : GceCmdlet
     {
         private class ParameterSetNames
         {
-            public const string ByName = "ByName";
+            public const string ByNameZone = "ByNameZone";
             public const string ByObject = "ByObject";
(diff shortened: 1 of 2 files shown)


Google.PowerShell/Compute/GceManagedInstanceGroupCmdlets.cs
```diff
@@ -1335,5 +1365,25 @@ protected override void ProcessRecord()
...
+            IList<ManagedInstance> instances = request.Execute().ManagedInstances;
+            while (instances != null && instances.Any(i => i.CurrentAction != "NONE") && !Stopping)
+            {
+                if (Timeout >= 0 && stopwatch.Elapsed.Seconds > Timeout)
+                {
+                    WriteWarning("Wait-GceManagedInstanceGroup timed out for " +
+                                 $"{request.Project}/{request.Region}/{request.InstanceGroupManager}");
+                    break;
+                }
+                Thread.Sleep(150);
```
iantalarico: No need to address here but should we make this configurable?

Google.PowerShell/Compute/GceManagedInstanceGroupCmdlets.cs
```diff
@@ -1335,5 +1365,25 @@ protected override void ProcessRecord()
                 instances = request.Execute().ManagedInstances;
             }
         }
+
+        private void WaitRegionManagedInstance(string project, string region, string name)
```
iantalarico: Suggestion: Create a common function for this and the above that takes a ClientServiceRequest and a string message to print while waiting.  The bulk of the logic is shared.

Google.PowerShell/Compute/GceManagedInstanceGroupCmdlets.cs
quoctruong: I think we can either add another parameter for that or set a config for the whole module. Nobody has asked for this before so I'm not sure how useful it is though.

quoctruong:
@iantalarico PTAL.

quoctruong:
Thanks!

PR #608, opened 2018-03-26
Fix a bug where Get-GceMetadata fails when called the second times

quoctruong:
The bug happens because we keep adding headers to the `DefaultRequestHeaders` of a static `HttpClient`. This means that by the second time this is called, the header will be something like `"Metadata-Flavor" = {"Google", "Google" }` instead of `"Metadata-Flavor" = "Google"`.

### DIFF
--- a/Google.PowerShell.Tests/Compute/GceMetadataTests.cs
+++ b/Google.PowerShell.Tests/Compute/GceMetadataTests.cs
@@ -1,0 +1,75 @@
+﻿// Copyright 2018 Google Inc. All Rights Reserved.
+// 
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+// 
+//     http://www.apache.org/licenses/LICENSE-2.0
+// 
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Google.Apis.Compute.v1.Data;
+using NUnit.Framework;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using System.Management.Automation;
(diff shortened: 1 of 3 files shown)


Google.PowerShell.Tests/Compute/GceMetadataTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            Pipeline.Commands.AddScript(
+                $"Get-GceMetadata");
+            try
+            {
+                Collection<PSObject> results = Pipeline.Invoke();
+            }
+            // Exception thrown because we are not running from a VM.
+            catch { }
+
+            IEnumerable<string> headers = GetGceMetadataCmdlet.Client.DefaultRequestHeaders.GetValues(
```
iantalarico: Should we check:
Assert.Equal("Google", GetGceMetadataCmdlet.Client.DefaultRequestHeaders.Get(GetGceMetadataCmdlet.MetadataFlavorHeader)); ?

Google.PowerShell.Tests/Compute/GceMetadataTests.cs
```diff
@@ -0,0 +1,75 @@
...
+            Pipeline.Commands.AddScript(
+                $"Get-GceMetadata; Get-GceMetadata");
+            try
+            {
+                Collection<PSObject> results = Pipeline.Invoke();
+            }
+            // Exception thrown because we are not running from a VM.
+            catch { }
+
+            IEnumerable<string> headers = GetGceMetadataCmdlet.Client.DefaultRequestHeaders.GetValues(
```
iantalarico: Same here

Google.PowerShell.Tests/Compute/GceMetadataTests.cs
quoctruong: The `DefaultRequestHeaders` only has `GetValues` and no `Get` (see https://msdn.microsoft.com/en-us/library/system.net.http.headers.httprequestheaders(v=vs.118).aspx). Is there another `Get` method that you're thinking of?

iantalarico: Nope, just assumed it had a Get function.

quoctruong:
Thanks!

PR #613, opened 2018-05-06
Properly pipeline the enumeration of GetProjects.

### LINKED ISSUE #612: Cannot filter Get-GcpProject using where-object
Get-GcpProject supports -name, but when trying to run Get-GcpProject | where, filter returns all entries.
Would be great if the command supports filtering, so we're not limited to the filters offered by the command

ILMTitan:
Fixes #612

### DIFF
--- a/Google.PowerShell.Tests/CloudResourceManager/CloudResourceManagerCmdletTestBase.cs
+++ b/Google.PowerShell.Tests/CloudResourceManager/CloudResourceManagerCmdletTestBase.cs
@@ -1,0 +1,38 @@
+﻿// Copyright 2018 Google Inc. All Rights Reserved.
+// Licensed under the Apache License Version 2.0.
+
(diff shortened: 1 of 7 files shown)


Google.PowerShell.Tests/CloudResourceManager/CloudResourceManagerCmdletTestBase.cs
```diff
@@ -0,0 +1,38 @@
...
+using Google.PowerShell.Tests.Common;
+using Moq;
+using NUnit.Framework;
+
+namespace Google.PowerShell.Tests.CloudResourceManager
+{
+    public class CloudResourceManagerCmdletTestBase : PowerShellTestBase
+    {
+        /// <summary>
+        /// The mock of the compute service. Reset before every test.
```
quoctruong: nit: Should be "mock of the cloud resource manager service".

Issue #615, opened 2018-05-08
Extend New-GceAttachedDiskConfig to support local SSD scratch disks


pjmcc2009:
Currently there is no way to use the Google Cloud Powershell tools to create a GCE instance with one or more local SSD scratch disks.  This can be implemented by extending the New-GceAttachedDiskConfig cmdlet to include support for local SSD disks.

Proposed solution:

Extend the parameter set for New-GceAttachedDiskConfig to include a single parameter ScratchDiskZone which takes in a GCE zone as input.  The cmdlet then returns a AttachedConfig object of type SCRATCH and diskType /zones/{zone}/diskTypes/local-ssd

quoctruong:
This fix is available in version 1.0.1.8 of the module.

PR #616, opened 2018-05-08
Pull Request for Issue # 615

pjmcc2009:
Pull request for issue #615


Google.PowerShell/Compute/GceAttachedDiskConfigCmdlets.cs
```diff
@@ -111,19 +115,36 @@ private class ParameterSetNames
...
 
+        /// <summary>
+        /// <para type="description">
+        /// The zone in which the instance resides in.
+        /// </para>
+        /// </summary>
+        [Parameter(ParameterSetName = ParameterSetNames.Scratch, Mandatory = true)]
+        [ConfigPropertyName(CloudSdkSettings.CommonProperties.Zone)]
+        [PropertyByTypeTransformation(Property = "Name", TypeToTransform = typeof(Zone))]
+        public string ScratchDiskZone { get; set; }
```
quoctruong: I think we should just have a `Zone` parameter because `Zone` can also be used with `pd-standard` and `ps-ssd`. The `DiskType` parameter should then be changed to a fixed set of `pd-standard`, `pd-ssd` and `local-ssd`. So the user can do something like `New-GceAttachedDiskConfig -Zone us-west1 -DiskType local-ssd` and the cmdlet would automatically construct `/zones/us-west1/diskTypes/local-ssd` behind the scene.

quoctruong:
@pjmcc2009 Thank you for the contribution! I think we should add a `Zone` parameter and modify the `DiskType` parameter to have a `ValidateSet` attribute with only `pd-standard`, `pd-ssd` and `local-ssd`. This should cover more scenarios.

Google.PowerShell/Compute/GceAttachedDiskConfigCmdlets.cs
pjmcc2009: Putting a fixed set on the Zone parameter may break some existing implementations.  Right now the DiskType parameter is open to taking any string, however I suspect anyone who uses this would be explicitly setting this with the URL.  This is because if you create the AttachedDisk object without the URL explicitly set for the DiskType it throws an error when you try to use it to create or edit an instance.

For example, this code:
```
$foo = New-GceAttachedDiskConfig -SourceImage $image -DiskType "pd-ssd" -Boot
Add-GceInstance -Project $project -Zone "us-central1-b" -Disk $foo -Name "footest"
```
Throws the error:
```
Add-GceInstance : Google.Apis.Requests.RequestError
Invalid value for field 'resource.disks[0].initializeParams.diskType': 'pd-ssd'. The URL is malformed. [400]
```

This would force anyone using this today to explictly set the URL in their code.
```
$foo = New-GceAttachedDiskConfig -SourceImage $image -DiskType "/zones/us-central1-b/diskTypes/pd-ssd" -Boot
Add-GceInstance -Project $project -Zone "us-central1-b" -Disk $foo -Name "footest"
```
I do agree though that it is much cleaner to have separate parameters for Zone and DiskType, with DiskType limited to pd-standard, pd-ssd, and local-ssd.

quoctruong: @pjmcc2009 Thank you for the explanation. I think it would be better to have a slight breaking change here to introduce better cmdlet parameters. Otherwise, I feel the parameter sets will be quite awkward and not as intuitive with `ScratchDiskZone` and no `PersistentDiskZone`.
@ILMTitan, do you have any better suggestions foor the parameter set?

pjmcc2009: What do you think of having just two parameter sets (NewDisk and ExistingDisk) and setting mandatory to false on the SourceImage param?

quoctruong: @pjmcc2009 I do agree that we shouldn't have 3 parameter sets "Persistent", "New" and "Scratch". But I still don't think it would be good to have a Zone parameter that just applies for the ScratchDisk case.
What do you think about simply changing the type of the AttachedDisk to Scratch instead of Persistent if the DiskType URL contains local-ssd (since I don't think it can be Persistent if the DiskType is local-ssd)? This implementation shouldn't break backwards compatibility.

pjmcc2009: Yeah I think that's a good idea.  We would still have to set the SourceImage parameter to be non mandatory, otherwise you wouldn't be able to provision the scratch disk.

I'll send through an update to the pull request with these changes for everyone to take a look.

quoctruong: @pjmcc2009 Sounds great!

Google.PowerShell/Compute/GceAttachedDiskConfigCmdlets.cs
```diff
@@ -80,7 +81,7 @@ private class ParameterSetNames
         /// </para>
         /// </summary>
         [Parameter(ParameterSetName = ParameterSetNames.New)]
-        public long? Size { get; set; }
+        public long? Size { get; set; }
```
quoctruong: nit: Extra space at the end.

Google.PowerShell/Compute/GceAttachedDiskConfigCmdlets.cs
```diff
@@ -134,14 +135,23 @@ protected override void ProcessRecord()
                 Source = Source?.SelfLink
             };
 
+            if(DiskType != null && DiskType.Contains("local-ssd"))
```
quoctruong: We should use the `GetUriPart` function inherited from `GcloudCmdlet` to extract out the type of the disk. A call to `GetUriPart("disktype", DiskType.ToLower())` should return `local-ssd`.

pjmcc2009: Ok, will fix and update.

quoctruong:
@pjmcc2009 Would you like me to merge the changes for you?

pjmcc2009:
Yeah for sure.  Sorry I don't think that I have permissions to do the merge on my side.

quoctruong:
@pjmcc2009 No problem! I just want to make sure that there are no additional changes that you want to check in.

Issue #626, opened 2018-09-07
Content-Encoding gzip error


justlearntutors:
Here is my command for Powershell

New-GcsObject -Bucket ssl-images-justlearn -ContentType application/javascript -ObjectName $JsStr -Metadata @{ "Content-Encoding" = "gzip"} -File C:\Users\Philip\source\repos\Justlearn\Justlearn\wwwroot\script.min.js.gz -Force

I want to add gzip for Content-Type

The result is shown here. https://imgur.com/a/7mtfvtw

How can I solve?

Content-Type gzip is added in the wrong place.

quoctruong:
@justlearntutors I don't think Content-Type can be gzip (gzip is Content-Encoding, not the underlying content type). You can read more about that under the Content-Encoding section [here](https://cloud.google.com/storage/docs/metadata).

justlearntutors:
How can I add Content-Encoding in the right place?

-Metadata @{ "Content-Encoding" = "anything"}

See image.
![Image of Cloud](https://ssl.justlearn.com/0test/google-cloud-content-type-paint.png)

quoctruong:
@justlearntutors I did some investigation and turns out this is a bug in the cmdlet. We will fix this in the next release. Thanks for finding this out!

PR #627, opened 2018-09-10
Fix a bug where fixed type metadata is not handled properly in GCS Object cmdlets.

quoctruong:
This should fix https://github.com/GoogleCloudPlatform/google-cloud-powershell/issues/626. See [link](https://cloud.google.com/storage/docs/metadata#mutable) for more information on fixed-key metadata.

Also fixes minor test errors.

---
This change is [Reviewable](https://reviewable.io/reviews/googlecloudplatform/google-cloud-powershell/627)

### DIFF
--- a/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
+++ b/Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
@@ -295,15 +295,44 @@ Describe "New-GcsObject" {
         Remove-GcsObject $emptyObj
     }
 
+    It "should work with fixed-type metadata" {
+        $obj = New-GcsObject $bucket "metadata-test" `
+            -ContentEncoding "gzip" -ContentType "image/png"
+        $obj.Metadata | Should BeNullOrEmpty
+        $obj.ContentType | Should Be "image/png"
+        $obj.ContentEncoding | Should Be "gzip"
+        Remove-GcsObject $obj
+    }
+
+    It "should work with fixed-type metadata through -Metadata parameter" {
+        $obj = New-GcsObject $bucket "metadata-test" `
+            -Metadata @{ "Content-Encoding" = "gzip"; "Content-Type" = "image/png" }
+        $obj.Metadata | Should BeNullOrEmpty
+        $obj.ContentType | Should Be "image/png"
+        $obj.ContentEncoding | Should Be "gzip"
+        Remove-GcsObject $obj
+    }
+
     It "should write metadata" {
         $obj = New-GcsObject $bucket "metadata-test" `
             -Metadata @{ "alpha" = 1; "beta" = "two"; "Content-Type" = "image/png" }
-        $obj.Metadata.Count = 3
+        $obj.Metadata.Count = 2
         $obj.Metadata["alpha"] | Should Be 1
         $obj.Metadata["beta"] | Should Be "two"
-        $obj.Metadata["Content-Type"] | Should Be "image/png"
         # Content-Type can be set from metadata.
         $obj.ContentType | Should Be "image/png"
+        Remove-GcsObject $obj
+    }
+
+    It "should work with both fixed-type and custom metadata" {
+        $obj = New-GcsObject $bucket "metadata-test" `
+            -Metadata @{ "Content-Encoding" = "gzip"; "Content-Type" = "image/png";
+                         "alpha" = 1; "beta" = "two" }
+        $obj.Metadata.Count = 2
+        $obj.Metadata["alpha"] | Should Be 1
+        $obj.Metadata["beta"] | Should Be "two"
+        $obj.ContentType | Should Be "image/png"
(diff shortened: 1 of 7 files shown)


Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -49,23 +49,53 @@ protected bool TestObjectExists(StorageService service, string bucket, string ob
...
+                ContentDisposition = contentDisposition,
+                CacheControl = cacheControl,
+                ContentLanguage = contentLanguage
             };
 
+            if (metadata != null)
+            {
+                // Handles fixed-key metadata. Removes them so there won't be duplicate. See:
+                // https://cloud.google.com/storage/docs/metadata#mutable
+                if (metadata.ContainsKey("Content-Type"))
```
ILMTitan: This should probably be a list of constants and a loop (or the values of a Dictionary).

Alternatively, it could be moved inside of `GetFixedTypeMetadata`

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
```diff
@@ -295,18 +295,47 @@ Describe "New-GcsObject" {
         Remove-GcsObject $emptyObj
     }
 
+    It "should work with fixed-type metadata" {
```
ILMTitan: Would it be better to write this as a unit test than an integration test?

Google.PowerShell/Storage/GcsCmdlet.cs
```diff
@@ -139,5 +139,30 @@ public static string InferContentType(string file)
...
+        /// Gets fixed type metadata from the cmdlet parameter if user provides that.
+        /// If not, tries retrieving it from the metadata dictionary using keyNameInMetadataDictionary.
+        /// If that also fails, returns the default content type.
+        /// </summary>
+        /// <returns></returns>
+        protected string GetFixedTypeMetadata(
+            string parameterName,
+            Dictionary<string, string> metadataDictionary,
+            string keyNameInMetadataDictionary,
+            string defaultContentType = null)
```
ILMTitan: defaultContentType => defaultValue

Google.PowerShell/Storage/GcsCmdlet.cs
```diff
@@ -139,5 +139,30 @@ public static string InferContentType(string file)
...
+        /// <summary>
+        /// Gets fixed type metadata from the cmdlet parameter if user provides that.
+        /// If not, tries retrieving it from the metadata dictionary using keyNameInMetadataDictionary.
+        /// If that also fails, returns the default content type.
+        /// </summary>
+        /// <returns></returns>
+        protected string GetFixedTypeMetadata(
+            string parameterName,
+            Dictionary<string, string> metadataDictionary,
+            string keyNameInMetadataDictionary,
```
ILMTitan: Perhaps use a ReadOnlyDictionary to map this name?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -49,23 +49,53 @@ protected bool TestObjectExists(StorageService service, string bucket, string ob
...
+                {
+                    metadata.Remove("Content-Type");
```
ILMTitan: Are we concerned about side-effects on the object passed in by the user?

Google.PowerShell/Storage/GcsObject.cs
```diff
@@ -274,8 +304,69 @@ private class ParameterSetNames
         /// </summary>
         [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.ContentsFromFile)]
         [Parameter(Mandatory = false, ParameterSetName = ParameterSetNames.ContentsFromString)]
+        [ValidateNotNullOrEmpty]
         public string ContentType { get; set; }
 
+        /// <summary>
```
ILMTitan: Do we need these extra parameters? The user can currently use the metadata to set these values.

Google.PowerShell/Storage/GcsObject.cs
quoctruong: Yes, these metadata should be set through the properties on the GcsObject and not in the metadata dictionary.

Google.PowerShell/Storage/GcsObject.cs
quoctruong: They are different than the custom metadata parameters and they can directly affect the storage object property (like cache control, etc.) so I think they should get their own parameters.

Google.PowerShell.IntegrationTests/Storage/Storage.GcsObject.Tests.ps1
quoctruong: I needed to test that the object returned had the metadata in the correct place.

quoctruong:
@ILMTitan PTAL.

quoctruong:
Thanks!

